PR #1, opened 2019-02-28
Add initial `eslint` config

orballo:


.eslintrc
```diff
@@ -1,16 +1,25 @@
 {
+  "parser": "@typescript-eslint/parser",
+  "plugins": [
```
luisherranz: I thought that if you use the `plugin:xxx/recommended` you don't need to add them in `plugins`.

orballo: And you thought right. Thing is I tried like you say before, and for some reason it didn't work, but I have it working now.

luisherranz: Super!

luisherranz (approved):
👍

PR #4, opened 2019-03-06
Forces default settings for prettier

orballo:


luisherranz:
I've already done it here @orballo: https://github.com/frontity/frontity/pull/3/commits/668835bc5048e7d37edabc90e97d54a9678d273d

PR #23, opened 2019-04-09
Implement `create` command in `frontity`.

orballo:


packages/core/package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "homepage": "https://frontity.org",
   "license": "Apache-2.0",
-  "main": "src/index.js",
+  "main": "dist/scripts/index.js",
```
luisherranz, line 10: Don't you need to add a `"types"` field here?

packages/core/src/scripts/build.ts
```diff
@@ -60,14 +60,12 @@ const build = async ({
   if (mode === "production") await remove(join(outDir, "bundling"));
 };
 
-(process as NodeJS.EventEmitter).on("unhandledRejection", (error: Error) => {
-  console.error(error);
-  process.exit(1);
-});
+// (process as NodeJS.EventEmitter).on("unhandledRejection", (error: Error) => {
```
luisherranz: You can remove all the comments of the `dev`, `build` and `serve` files to keep them clean.

luisherranz (changes requested):
Apart from those two comments I think it's great!

DAreRodz (approved):
👌

luisherranz (approved):
👏

PR #29, opened 2019-04-22
Add helmet SSR

luisherranz:


luisherranz:
Ready for review 👍

packages/core/src/server/templates/html.ts
```diff
@@ -1,17 +1,21 @@
 import { Template } from "../../types";
 
-const template: Template = ({ html, scriptTags, linkTags }) => `<!doctype html>
+const template: Template = ({ html, frontity, head }) => `<!doctype html>
     <html>
-      <head>
+      <head ${head.htmlAttributes}>
```
orballo: This line seems wrong. Shouldn't those attributes be rendered inside `<html>`?

luisherranz: You're right.

DAreRodz (approved):

packages/core/src/server/templates/html.ts
luisherranz: Fixed

DAreRodz: Uh, I missed that before approving the PR. Nice catch, @orballo.

orballo (approved):

PR #31, opened 2019-04-24
Implement Tiny Router

DAreRodz:


luisherranz (changes requested):
Travis is failing:
```
tsconfig.build.json(7,5): error TS5053: Option 'allowJs' cannot be specified with option 'declaration'.
tsconfig.build.json(7,5): error TS5053: Option 'declaration' cannot be specified with option 'isolatedModules'.
```
I think you'd need to set to `false` both `allowJs` and `isolatedModules` in the `tsconfig.build.json` files.

DAreRodz:
It should work now.

packages/tiny-router/src/html/server.ts
```diff
@@ -0,0 +1 @@
+export { default } from "..";
```
orballo, line 1: Sexy!

orballo (approved):

luisherranz (approved):

PR #37, opened 2019-05-06
Add connect!

luisherranz:
First implementation. Only actions and derived state.


luisherranz:
Ready for review.

This is the first implementation only. I suggest you to take a look at the last two commits (https://github.com/frontity/frontity/pull/37/commits/8514cbd8cd563884f1a6e282e0501a715985b583 & https://github.com/frontity/frontity/pull/37/commits/9a7abd62b5d7562375c55d73cc597f780532769b) instead of the file changes to view what I've changed from the original `observer-util` library.

Basically I made it work with our derived state, derived functions and actions.

orballo (approved):

DAreRodz (approved):
💖

PR #42, opened 2019-05-08
Fix union types in connect

luisherranz:
This a quick fix for a union problem in the Connect type. Now it omits the original `state` and `actions` from the package and includes only the properly processed ones.

I've also added tests for Connect.


orballo (approved):

luisherranz:
That was quick 😅

PR #45, opened 2019-05-10
Fix type of connect HOC

luisherranz:
The problem was that `connect(MyComp)` was returning the same component with the same props, including the injected props (`state`, `actions` and so on). Now it returns a component but with the injected props omitted.


luisherranz:
@orballo feel free to review, merge if approved and rebase your PR to this.

orballo (approved):

PR #56, opened 2019-05-13
feat(set): parse links to pathnames

orballo:


luisherranz (changes requested):
If it's a URL, I would add this regexp to get the page and path separately:
```js
const path = (/^(.*)\/page\//.exec(url.pathname) || [null, url.pathname])[1]
const page = parseInt(
          (/\/page\/(\d+)\/?$/.exec(url.pathname) || [null, "1"])[1],
          10
        )
```

luisherranz (approved):
Perfect 👍

PR #57, opened 2019-05-13
feat(actions): new action arguments including other actions and libraries

luisherranz:
This PR implements a new API for actions and derived state:

```
const myAction = ({ state, actions, libraries }) => {...};
const myDerivedState = ({ state }) => {...};
```


packages/router/__tests__/index.test.ts
```diff
@@ -6,7 +6,7 @@ const router: Router = {
     router: {
       path: "/some-path",
       page: 2,
-      location: state => new URL(state.router.path)
+      location: ({ state }) => new URL(state.router.path)
```
orballo, line 9: Does this work as expected? Because I think that `URL` will throw if the argument is not a proper url. And in this case is only the `path` part.

luisherranz: Yes, you're right the code is wrong but those tests only test types, not the actual code. So that part is just a dummy implementation of a function that receives state and returns a URL, which is the thing we are interested in :)

orballo: Oh, sorry, I missed it was a test file.

luisherranz: No prob!

orballo (approved):

PR #60, opened 2019-05-15
Add pagination to mars

luisherranz:


packages/mars-theme/src/html/components/link.js
```diff
@@ -1,13 +1,16 @@
...
+  const href = page ? `${path}page/${page}` : path;
+
   const onClick = event => {
     event.preventDefault();
     // Fetch the data if it's not already fetched.
-    actions.source.fetch(href);
+    actions.source.fetch({ path, page });
     // Set the router to the new url.
-    actions.router.set(href);
+    actions.router.set({ path, page });
```
orballo, line 13: I'm afraid this doesn't work. If you pass a full url inside the `path` field, it won't be parsed to `path` and `page` by `router.set()`. You need to either parse it before and pass an object with each part, or send the url as a string parameter, no object.

https://github.com/frontity/frontity/blob/666f6db50ba2e45834266f6b3a90f2916ada4aeb/packages/tiny-router/src/actions.ts#L9-L19

orballo (changes requested):

packages/mars-theme/src/html/components/link.js
luisherranz: There you go. Now router accepts the same type of urls in both the string or the `path` parameter. I also made page optional and if don't passed it will be inferred from the `path`.

orballo (approved):

PR #64, opened 2019-05-16
chore(mars-theme): add 404 page

DAreRodz:


packages/mars-theme/src/html/components/theme.js
```diff
@@ -30,6 +31,7 @@ const Theme = ({ state }) => (
     <Body>
       {state.source.data(state.router.path).isArchive && <List />}
       {state.source.data(state.router.path).isPostType && <Post />}
+      {state.source.data(state.router.path).is404 && <Page404 />}
```
orballo, line 34: I tried this yesterday and `is404` was undefined when I entered a wrong url. When exactly is `is404` populated?

I think it needs some refactoring in `wp-source` in order to cover all the cases where there is no content available.

DAreRodz: @orballo, what URL did you try? The property `is404` is set [here](https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/actions.ts#L35), when a handler throws an exception.

orballo: I tried something invented like `/invented-path/` or something like that. Maybe the error was somewhere else and I was mistaken?

luisherranz (approved):
It works fine in my local environment :)

orballo (approved):

packages/mars-theme/src/html/components/theme.js
DAreRodz: I don't know, but it seems to be working fine now.

PR #83, opened 2019-05-20
Fix default wp.org patterns for permalinks

### LINKED ISSUE #82: wp-source should support common settings for permalinks by default
We need to change the [wp-org](https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/libraries/patterns/wp-org.ts) patterns to match all the permalink structures below (except for the custom one).

Currently, we support the custom one (as appears in the image), and the others don't. 😟

Screenshot from 2019-05-20 11-40-26

DAreRodz:
Fixes #82 

Support for "plain" permalink was not implemented as our current `resolver` implementation cannot differentiate URLs based on the search query params.


luisherranz (approved):
👍

PR #115, opened 2019-06-05
Changes in `source` and `route` API.

DAreRodz:
Summary:
- Rename `getRoute` to `stringify` and `getParams` to `parse` in `source.libraries`.
- Accept only strings in `router.set`, `source.fetch` and `source.get`.
- Get rid of `path`, `page`, `query` in `router` and add `link`.
- Rename `frontity.initial` to `frontity.initialLink` and convert it to a string.
- Adapt `mars-theme` to these changes.
- Extra: fix a bug in `taxonomy` handler.

### DIFF
--- a/packages/mars-theme/src/components/link.js
+++ b/packages/mars-theme/src/components/link.js
@@ -1,10 +1,7 @@
 import React from "react";
 import { connect } from "frontity";
 
-const Link = ({ actions, libraries, link, className, children }) => {
-  // normalizes link
-  link = libraries.source.stringify(link);
-
+const Link = ({ actions, link, className, children }) => {
   const onClick = event => {
     event.preventDefault();
     // Set the router to the new url.
--- a/packages/mars-theme/src/components/link.js
+++ b/packages/mars-theme/src/components/link.js
@@ -1,28 +1,19 @@
 import React from "react";
 import { connect } from "frontity";
 
-const Link = ({
-  actions,
-  libraries,
-  path,
-  page,
-  query,
-  className,
-  children
-}) => {
-  // If page is passed, build final href.
-  // const href = page ? `${path}page/${page}` : path;
-  const href = libraries.source.getRoute({ path, page, query });
+const Link = ({ actions, libraries, link, className, children }) => {
+  // normalizes link
+  link = libraries.source.stringify(link);
 
   const onClick = event => {
     event.preventDefault();
     // Set the router to the new url.
-    actions.router.set({ path, page, query });
+    actions.router.set(link);
     window.scrollTo(0, 0);
   };
 
   return (
-    <a href={href} onClick={onClick} className={className}>
+    <a href={link} onClick={onClick} className={className}>
       {children}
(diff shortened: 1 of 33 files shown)


packages/tiny-router/src/actions.ts
```diff
@@ -6,18 +6,16 @@ export const set: TinyRouter["actions"]["router"]["set"] = ({
...
   actions,
   libraries
-}) => routeOrParams => {
-  const { getParams, getRoute } = libraries.source;
-  const { path, page, query } = getParams(routeOrParams);
-  const route = getRoute(routeOrParams);
+}) => link => {
+  const { stringify } = libraries.source;
+  // normalizes link
+  link = stringify(link);
```
luisherranz: Two things:
- Does it make sense to expose a `libraries.source.normalize` utility for these cases? I don't know if using `stringify` in a `string` makes sense. Seems confusing to me.
- Either way, if `tiny-router` is dependent on `libraries.source.stringify` or `libraries.source.normalize` then those APIs should be in `source/index.ts` instead of `wp-source/index.ts` because any source that don't implement them won't work with `tiny-router`.

Also, if finally `libraries.source.stringify` is exposed, I think `libraries.source.parse` should be exposed as well.

packages/mars-theme/src/components/link.js
```diff
@@ -1,28 +1,19 @@
...
-  query,
-  className,
-  children
-}) => {
-  // If page is passed, build final href.
-  // const href = page ? `${path}page/${page}` : path;
-  const href = libraries.source.getRoute({ path, page, query });
+const Link = ({ actions, libraries, link, className, children }) => {
+  // normalizes link
+  link = libraries.source.stringify(link);
```
luisherranz: Is this really needed?

Because it is used in `actions.router.set(link)` where `link` is normalized internally and in `href` where normalization is not important, right?

luisherranz (changes requested):

packages/mars-theme/src/components/link.js
DAreRodz: You are right, it's not needed!

packages/tiny-router/src/actions.ts
DAreRodz: Actually, now that a `link` is always a string, you would need `stringify` only when you did `parse` previously. I'm going to change `stringify` to accept only objects, and expose another utility called `nomalize` that accept strings.

I think it's a good change.

DAreRodz: > if tiny-router is dependent on libraries.source.stringify or libraries.source.normalize then those APIs should be in source/index.ts instead of wp-source/index.ts

They are already exposed in `source/index.ts`!

luisherranz: Oh, I misread that. Sorry!

DAreRodz:
Ready to review again!

luisherranz (approved):
LGTM!

PR #116, opened 2019-06-05
feat(html2react): first implementation of Html2React

orballo:


luisherranz (dismissed):
Just two things:
- [x] Use 1.0.0 version for `html2react`.
- [x] Try to move the inline style to css props in emotion to see if it works out of the box.

If the `css` prop works:
- Can be removed with a processor?
- Can be changed with a processor?

orballo:
> * Can be removed with a processor?
> * Can be changed with a processor?

Yes they can, the `css` prop is added in the `parse` function, and that happens before going through the processors.

luisherranz (approved):
Looks awesome :)

PR #122, opened 2019-06-10
fix(source): return isFetching  and isReady with false by default in get

DAreRodz:


luisherranz:
@DAreRodz is there a test to assert that reactivity for `isReady` and `isFetching` works? Something like this:

```js
test("...", done => {
  expect(state.source.get("/").isReady).toBe(false);
  observe(() => {
    if (state.source.get("/").isReady === true) done();
  });
  actions.source.fetch("/");
});

```

DAreRodz:
There is not, but it should. I will add it.

packages/mars-theme/src/components/theme.js
```diff
@@ -38,7 +32,12 @@ const Theme = ({ state }) => {
       <HeadContainer>
         <Header />
       </HeadContainer>
-      <Body>{Content}</Body>
+      <Body>
+        {data.isFetching && <Loading />}
```
luisherranz, line 36: Wow, I love that! So simple :)

orballo (approved):

orballo:
I guess this one needs a squash, I'll let you to it @DAreRodz 😂

PR #130, opened 2019-06-14
feat(packages): creates packages @frontity/components, @frontity/hooks, and adds image processor to @frontity/html2react

orballo:
Implements `@frontity/components/image` and `@frontity/hooks/use-in-view`


packages/components/image.tsx
```diff
@@ -0,0 +1,122 @@
...
+      changeAttributes(attributes);
+      return <img {...attributes} />;
+    }
+
+    if (typeof IntersectionObserver !== "undefined") {
+      const [onScreen, ref] = useInView({ onlyOnce: true });
+      if (onScreen) changeAttributes(attributes);
+      return (
+        <>
+          <NoScriptImage {...attributes} />
```
luisherranz, line 102: Why does the `IntersectionObserver` version needs a `<NoScriptImage>`? To comply with the SSR version? Isn't that one of the reasons of `suppressHydrationWarning`?

packages/components/image.tsx
```diff
@@ -0,0 +1,122 @@
...
+    "data-srcset": srcSet,
+    className: "frontity-lazy-image".concat(className ? ` ${className}` : ""),
+    loading: loading || "auto",
+    style: { visibility: "hidden" }
+  };
+
+  if (typeof window !== "undefined") {
+    if (typeof (HTMLImageElement as any).prototype.loading !== "undefined") {
+      changeAttributes(attributes);
+      return <img {...attributes} />;
```
luisherranz, line 94: Is `loading` here `"auto"` by default? Is that the correct default parameter? (instead of `"lazy"`):
Screen Shot 2019-06-18 at 16 18 12

packages/components/image.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+interface Props {
+  src?: string;
+  srcSet?: string;
+  alt?: string;
+  className?: string;
+  loading?: "auto" | "lazy" | "eager";
+}
+
+type Image = React.FC<Connect<Package, Props>>;
```
luisherranz: Why do you need `Connect` here. Is it because in the future the component will be connected?

packages/components/image.tsx
```diff
@@ -0,0 +1,122 @@
...
+`;
+
+// Finds all the images rendered by this component and maps
+// their `data-src` and `data-srcset` attributes to `src` and `srcset`
+// when the browser doesn't support Proxy or IntersectionObserver.
+const noProxyScript = `
+  if (typeof window !== "undefined" && (!("Proxy" in window) || !("IntersectionObserver" in window))) {
+    document.addEventListener("DOMContentLoaded", () => {
+      const images = document.querySelectorAll("img.frontity-lazy-image");
+      for (image of images) {
```
luisherranz: This script needs to be in es5 but I'll do the fix. Please be careful with those things :)

packages/components/image.tsx
```diff
@@ -0,0 +1,122 @@
...
+          <img ref={ref} {...attributes} />
+        </>
+      );
+    }
+  }
+
+  return (
+    <>
+      <Head
+        script={[
```
luisherranz, line 112: I see these scripts only get added once even tho the `<Image>` component is used multiple times. Is that a feature of `react-helmet`?

packages/hooks/use-in-view.ts
```diff
@@ -0,0 +1,79 @@
+import { useState, useEffect, useRef, Dispatch, SetStateAction } from "react";
```
luisherranz, line 1: @DAreRodz could you please take a look at this file? I don't know enough about how `IntersectionObserver` works :(

luisherranz (commented):
Great work. I made some comments with several doubts.

packages/components/image.tsx
orballo: To comply with the SSR version, yes. The `suppressHydrationWarning` currently implemented on the code only runs if `IntersectionObserver` and native load are not supported. In this case I didn't add the `supressHydrationWarning` because the changes are done inside React and the warning is perfectly valid. If we remove it we might miss other eventual errors. In the other case there was no work around as the HTML is changed outside of React.

Also, we should keep in mind that this problem will last until the native lazy load is implemented. The number of users that will go through this code will be less and less from that point in time.

packages/components/image.tsx
orballo: Yep, if not specified the `loading` attribute will have the `auto` value.

packages/components/image.tsx
orballo: Yes, I forgot to remove it. Initially it was there because I was already implementing logic with the state and the media `id`, but then we decided to go without it for this first version.

packages/components/image.tsx
orballo: I guess so, I didn't check further after seeing that it worked as expected.

packages/components/image.tsx
orballo: Maybe we could use `raw-loader` in the future to transpile and minify this kind of scripts? Is it possible? I don't remember from the previous version.

packages/components/image.tsx
luisherranz: Perfect, thanks 👍

packages/components/image.tsx
luisherranz: And what's the point of not using `"lazy"` by default?

I mean... for example, native lazy load won't work right now in Mars:
https://github.com/frontity/frontity/blob/51afabdbe02bdf590d42a9afb0b7d5380ad4ea57/packages/mars-theme/src/components/featured-media.js#L25-L29

Or maybe I'm not getting what `"auto"` means...

packages/components/image.tsx
luisherranz: Ok. Let's leave it there then. Don't worry.

packages/components/image.tsx
luisherranz: Hmm.. I'm not sure if that's possible because `raw` is... well, raw 😆 

My idea a while back was to give packages the possibility of adding an additional entry-point that loads before React for this type of things.

packages/components/image.tsx
luisherranz: Magic :)

packages/components/image.tsx
orballo: The post only said that the decision is taken by the browser. How the browser decides, I have no clue. I just used `auto` as default because of the specification, but maybe we want to change it in the future. It won't be a breaking change until native lazy load is available.

orballo: Also keep in mind that this is an `Image` component, not a `LazyImage` component. Lazy load is just one of the features.

luisherranz: Oh, I see.

If that's the case maybe we should use the `loading` prop to opt-out of the IO implementation as well if the comp receives `"eager"`, shouldn't we?

Not that I think we should have that now, just that it would seem more consistent with how the component is used.

orballo: That's right. That crossed my mind earlier but I forgot it once I was fighting with the implementation 😅

luisherranz: I've made a fix. Let me know what you think.

packages/hooks/use-in-view.ts
```diff
@@ -0,0 +1,79 @@
...
+
+// This observer will be used during the whole life of the process.
+let observer: IntersectionObserver;
+
+// This callback can be called with more than one entry,
+// so we need to filter them and call the corresponding
+// `setIntersected` function for each changed entry.
+let createCallback: IntersectionObserverCallbackCreator = options => {
+  return entries => {
+    const changed = entries.filter(entry => entry.isIntersecting);
```
DAreRodz: If we're considering just those entries that describe a transition from not-intersecting to intersecting, then, when the state of the hook changes to `true`, it will never be `false` again, right?

If that's the case, I don't understand very well what the `onlyOnce` option is for. Uhm 🤔

orballo: Yeah, this is a major mistake. I'll change this.

orballo: Fixed

packages/components/image.tsx
orballo: Changed some things with the next steps of the implementation in mind.

luisherranz (approved):
LGTM!

@DAreRodz could you please take a look at the hook changes as well?

luisherranz:
@orballo could you please add the conventional-changelog title you deem convenient?

orballo:
Guys, I pushed some changes with the implementation of a processor to use `Image` with `html2react` have a look at it again please.

luisherranz:
Seems fine.

Does it work with WP lazy-loading plugins that modify the HTML to use `data-src` instead of `src`?

orballo:
It does now.

luisherranz (approved):
Amazing work. Thanks @orballo!

DAreRodz (approved):
LGTM!

PR #131, opened 2019-06-14
feat(wp-source): add support for subdirectory, redirections, pages as home, category and tag base

DAreRodz:


luisherranz (approved):
Incredible work David!

PR #135, opened 2019-06-18
fix(frontity): allow dots on packages names

### LINKED ISSUE #128: Cannot create a new project with a `.` character
Is there a reason for the limitation of characters when creating a new project? When I try to run the following command: `npx frontity create frontity.local` I get the following error: `Error: The name of the package is not valid. Please enter a valid one (only letters and dashes).`

cbravobernal:
allow_dots

Fixes #128.

### DIFF
--- a/packages/frontity/src/__tests__/utils.tests.ts
+++ b/packages/frontity/src/__tests__/utils.tests.ts
@@ -1,0 +1,34 @@
+import { isPackageNameValid } from "../utils";
+
+describe("isPackageNameValid", () => {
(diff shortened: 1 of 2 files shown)


luisherranz:
I've added failing tests for the requirements of both package.json name and Unix/Window file systems.

cbravobernal:
I will update the code then in order to make them pass :-)

luisherranz:
😍

You can add some positive tests if you want.

orballo (approved):
Excellent contribution @c4rl0sbr4v0 @cluny85! 👏👏👏

PR #155, opened 2019-07-01
feat(cli): Suggest matching commands if the user mistypes

### LINKED ISSUE #153: Suggest matching commands if the user makes a typo
## Proposed API

`frontity creat`

As per now, it warns the user such that the command is not being recognized. It would be great to have a command suggestion logic such that the nearest matching command (`create` in this case) is being suggested.

jamesgeorge007:
Suggest matching commands if the user makes a typo.
Fixes #153 

#### My PR is a:

- [ ] 🐞 Bug fix
- [x] 🚀 New feature
- [ ] 🔝 Improvement

#### Main update on the:

- [ ] Documentation
- [x] Framework

- [x] Yes!
- [ ] Work in progress

### DIFF
--- a/packages/frontity/src/actions/unknown.ts
+++ b/packages/frontity/src/actions/unknown.ts
@@ -1,7 +1,26 @@
+import didYouMean from "didyoumean";
 import { CommanderStatic } from "commander";
 import chalk from "chalk";
 
+/**
+ * @param {String} command - Unknown command
+ * @param {CommanderStatic} program - commander instance
+ * 
+ * @returns {void}
+ */
+
 export default (command: string, program: CommanderStatic) => {
-  console.log(chalk.red(`Unknown command: ${chalk.bold(command)}\n`));
+  console.log(chalk.red(`Unknown command: ${chalk.bold(command)}`));
+  console.log();
+
+  const availableCommands: string[] = program.commands.map(c => c._name);
+  const suggestion: string | string[] = didYouMean(command, availableCommands);
+  
+  if (suggestion) {
+	console.log(chalk.cyan(`Did you mean '${suggestion}'?`));
+	console.log();
+  }
+  
   program.help();
 };
+
--- a/package-lock.json
+++ b/package-lock.json
@@ -1139,6 +1139,12 @@
         "any-observable": "^0.3.0"
       }
     },
+    "@types/didyoumean": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/@types/didyoumean/-/didyoumean-1.2.0.tgz",
+      "integrity": "sha512-3QMjBOgPEXaOkhfQxiGgzn6JpofPSi8Z0RmQq65FIpZv4kCPT/jvYiYrwqaUrfiHZmwwRvyYJjJmoKopbvLPQw==",
+      "dev": true
+    },
     "@typescript-eslint/eslint-plugin": {
       "version": "1.4.2",
       "resolved": "https://registry.npmjs.org/@typescript-eslint/eslint-plugin/-/eslint-plugin-1.4.2.tgz",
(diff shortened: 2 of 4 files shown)


jamesgeorge007:
cc @luisherranz @orballo

DAreRodz (changes requested):
Hey @jamesgeorge007, great idea! 😁

There are some changes to do though. You can test the `frontity` CLI with your changes running `npm run build` command inside `/packages/frontity` directory and then doing `npx frontity [command]` inside `/examples/mars-theme-example`.

DAreRodz:
Oh, I saw a problem on our part. In [this line](https://github.com/jamesgeorge007/frontity/blob/feat/envinfo/packages/frontity/src/commands.ts#L86) the type of the `command` argument that receives the listener function is wrong. It should be `string[]` (it's a list of arguments).

It would be great if you fix it too, @jamesgeorge007!

packages/frontity/src/actions/unknown.ts
```diff
@@ -1,7 +1,26 @@
...
+ * @returns {void}
+ */
+
 export default (command: string, program: CommanderStatic) => {
-  console.log(chalk.red(`Unknown command: ${chalk.bold(command)}\n`));
+  console.log(chalk.red(`Unknown command: ${chalk.bold(command)}`));
+  console.log();
+
+  const availableCommands: string[] = program.commands.map(c => c._name);
+  const suggestion: string | string[] = didYouMean(cmd, availableCommands);
```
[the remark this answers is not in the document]
jamesgeorge007: Oops!

jamesgeorge007:
@DAreRodz It is supposed to be string right? What we intend to receive is a single command and not a list :thinking:

packages/frontity/src/actions/unknown.ts
```diff
@@ -1,7 +1,26 @@
...
 import chalk from "chalk";
 
+/**
+ * @param {String} command - Unknown command
+ * @param {CommanderStatic} program - commander instance
+ * 
+ * @returns {void}
+ */
+
 export default (command: string, program: CommanderStatic) => {
```
jamesgeorge007, line 12: The method signature here is self-explanatory.

DAreRodz:
> @DAreRodz It is supposed to be string right? What we intend to receive is a single command and not a list 🤔

Yeah, I mean, the `unknown` action is right, it's [the listener](https://github.com/jamesgeorge007/frontity/blob/feat/envinfo/packages/frontity/src/commands.ts#L86) which needs to be fixed. It receives a string array that we are treating as if it were a single string.

It should be something like this:

```typescript
program.on("command:*", ([command]: string[]) => unknown(command, program));

// instead of (command: string) => unknown(command, program)
```

packages/frontity/src/actions/unknown.ts
```diff
@@ -1,7 +1,26 @@
...
-  console.log(chalk.red(`Unknown command: ${chalk.bold(command)}\n`));
+  console.log(chalk.red(`Unknown command: ${chalk.bold(command)}`));
+  console.log();
+
+  const availableCommands: string[] = program.commands.map(c => c._name);
+  const suggestion: string | string[] = didYouMean(command, availableCommands);
+  
+  if (suggestion) {
+	console.log(`chalk.cyan(Did you mean '${suggestion}'?)`);
```
DAreRodz: The method call is inside the template string. Just fix this line and it's ready.

```typescript
console.log(chalk.cyan(`Did you mean '${suggestion}'?`));

// instead of console.log(`chalk.cyan(Did you mean '${suggestion}'?)`);
```

DAreRodz (changes requested):
There is one thing left to change.

packages/frontity/src/actions/unknown.ts
jamesgeorge007: Oh didn't notice!

jamesgeorge007:
cc @DAreRodz

DAreRodz (approved):

DAreRodz:
Great, thanks @jamesgeorge007!

jamesgeorge007:
@DAreRodz Would love to know your thoughts on [Mevn-CLI](https://github.com/madlabsinc/mevn-cli)

PR #156, opened 2019-07-02
refactor(redirections): allow to execute several followed redirects in priority order

DAreRodz:

The idea of this PR was to execute redirections sequentially in order to transform a route. For example, given the following route

```/blog/wp-tag/nature/```

instead of creating a redirection like

```javascript
const redirect = {
  name: "change-tag-route",
  priority: 10,
  pattern: "/blog/wp-tag/:slug/amp/",
  handler: ({ slug )} => `/tag/${slug}`
}
```

be able to use three redirections like

```javascript
const redirections = [
  {
    name: "remove-subdir",
    priority: 10,
    pattern: "/blog/:route+",
    handler: ({ route )} => `/${route}`
  },
  {
    name: "remove-amp-suffix",
    priority: 20,
    pattern: "/:route+/amp/",
    handler: ({ route )} => `/${route}`
  },
  {
    name: "change-tag-base",
    priority: 30,
    pattern: "/wp-tag/:slug/",
    handler: ({ slug )} => `/tag/${slug}`
  }
]
```

This approach, although it makes it more direct to create redirects based on settings, has shown to have problems, for example, when you want some routes to be ignored because they start with the wrong prefix (and thus should return `404`).

#### My PR is a:

- [ ] 🐞 Bug fix
- [ ] 🚀 New feature
- [x] 🔝 Improvement

#### Main update on the:

- [ ] Documentation
- [x] Framework

- [x] Yes!
- [ ] Work in progress

### DIFF
--- a/packages/wp-source/index.ts
+++ b/packages/wp-source/index.ts
@@ -8,6 +8,7 @@ export type Pattern<F extends Function = (params: any) => any> = {
   priority: number;
   pattern: string;
   func: F;
+  isRequired?: boolean;
 };
 
 export type Handler = (args: {
--- a/packages/wp-source/src/libraries/__tests__/redirections.tests.ts
+++ b/packages/wp-source/src/libraries/__tests__/redirections.tests.ts
@@ -1,53 +1,41 @@
-import { getMatch } from "../get-match";
-
-// Some redirections
-const redirections: {
-  name: string;
-  priority: number;
-  pattern: string;
-  func: (params: Record<string, string>) => string;
-}[] = [
-  {
-    name: "",
-    priority: 10,
-    pattern: "/",
-    func: () => "/about-us/"
-  },
-  {
-    name: "",
-    priority: 10,
-    pattern: "/posts/",
-    func: () => "/"
-  },
-  {
-    name: "",
-    priority: 10,
-    pattern: "/wp-cat/:subpath+",
-    func: ({ subpath }) => `/category/${subpath}/`
-  },
-  {
-    name: "",
-    priority: 10,
-    pattern: "/category/(.*)",
-    func: () => ""
-  },
-  {
-    name: "",
-    priority: 10,
-    pattern: "/blog/:subpath*/",
-    func: ({ subpath = "" }) => `/${subpath}${subpath ? "/" : ""}`
-  }
-];
-
-const redirect = (path, redirections) => {
-  const match = getMatch(path, redirections);
-  if (!match) return path;
-  return match.func(match.params);
-};
+import { redirect } from "../redirect";
 
 // Add redirects
-describe("getMatch", () => {
-  test("redirects different routes", async () => {
+describe("redirect", () => {
+  test("redirects different routes (same priority)", () => {
+    // Some redirections
+    const redirections: {
+      name: string;
+      priority: number;
+      pattern: string;
+      func: (params: Record<string, string>) => string;
+    }[] = [
+      {
+        name: "",
+        priority: 10,
+        pattern: "/",
+        func: () => "/about-us/"
+      },
+      {
+        name: "",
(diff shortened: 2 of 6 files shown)


packages/wp-source/src/libraries/__tests__/redirections.tests.ts
```diff
@@ -56,7 +44,68 @@ describe("getMatch", () => {
...
+        name: "",
+        priority: 0,
+        pattern: "/:mainpath+/amp/",
+        func: ({ mainpath }) => `/${mainpath}/`
+      },
+      {
+        name: "",
+        priority: 0,
+        pattern: "/(.*)",
+        func: () => ""
```
luisherranz, line 67: I don't understand why you need two of these.

One here 👆

packages/wp-source/src/libraries/__tests__/redirections.tests.ts
```diff
@@ -56,7 +44,68 @@ describe("getMatch", () => {
...
+      {
+        name: "",
+        priority: 10,
+        pattern: "/subdir/:subpath*",
+        func: ({ subpath }) => (subpath ? `/${subpath}/` : "/")
+      },
+      {
+        name: "",
+        priority: 10,
+        pattern: "/(.*)",
```
luisherranz, line 78: And another here 👆

packages/wp-source/index.ts
```diff
@@ -8,6 +8,7 @@ export type Pattern<F extends Function = (params: any) => any> = {
   priority: number;
   pattern: string;
   func: F;
+  isRequired?: boolean;
```
luisherranz, line 11: Why do you need `isRequired`? When is it used? (I don't see it in the rest of the PR)

luisherranz (commented):
I've made a couple of questions

packages/wp-source/src/libraries/redirect.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const redirect = (
+  path: string,
+  redirections: WpSource["libraries"]["source"]["redirections"]
+) => {
+  // sort redirections
+  redirections.sort(({ priority: p1 }, { priority: p2 }) => p1 - p2);
+
+  // apply redirections in order
+  let lastPriority = -1;
```
luisherranz, line 12: Uhm... what happens if people use negative priorities?

packages/wp-source/index.ts
DAreRodz: Oops, I forgot to delete that line.

packages/wp-source/src/libraries/__tests__/redirections.tests.ts
DAreRodz: Ok, I'm going to add a description to the PR and explain there why,

packages/wp-source/src/libraries/redirect.ts
DAreRodz: It  won't work... But `lastPriority` could have a value of  `-Infinity` and this would be fixed. I'll change that.

luisherranz: Maybe we should do the same for the other arrays with priorities? (handlers and processors)

DAreRodz:
I think it is better to change the patterns of the handlers to solve the problem of redirects, and leave the redirects as they are.

@luisherranz, if you agree, we reject this PR and continue in https://github.com/frontity/frontity/pull/173.

luisherranz:
> change the patterns of the handlers 

What do you mean by change the patterns of the handlers?

(you can answer here https://github.com/frontity/frontity/pull/173)

luisherranz:
> if you agree, we reject this PR and continue in #173.

Feel free to close it.

PR #168, opened 2019-07-17
feat(frontity): expose fetch and URL from frontity package

### LINKED ISSUE #166: Expose "fetch" and "URL" in node global

`fetch` and `URL` should be available on node `global`.

Right now they are not.

We are using `cross-fetch` but it adds 15kbs of code.

**Solution**

We can import `URL` from the node module `"url"` and fetch from `"node-fetch"`.
Then import them in `/server/index.ts` like this:
```js
(global as any).fetch = fetch;
(global as any).URL = URL;
```

_It will work on TypeScript because it thinks they come from window._

Also, we should get rid of `cross-fetch` on `wp-source` and the regexps that are not using `URL` yet.

DAreRodz:

Closes #166

#### My PR is a:

- [ ] 🐞 Bug fix
- [ ] 🚀 New feature
- [x] 🔝 Improvement

#### Main update on the:

- [ ] Documentation
- [x] Framework

- [x] Yes!
- [ ] Work in progress

### DIFF
--- a/packages/wp-source/package.json
+++ b/packages/wp-source/package.json
@@ -27,13 +27,13 @@
     "@types/jest": "^24.0.11",
     "frontity": "^1.2.2",
     "jest": "^24.7.1",
+    "node-fetch": "^2.6.0",
     "ts-jest": "^24.0.2",
     "typescript": "^3.4.5"
   },
   "dependencies": {
     "@frontity/connect": "^1.0.1",
     "@frontity/source": "^1.0.11",
-    "cross-fetch": "^3.0.2",
     "normalizr": "^3.3.0",
     "path-to-regexp": "^3.0.0"
   }
--- a/packages/wp-source/src/libraries/api.ts
+++ b/packages/wp-source/src/libraries/api.ts
@@ -1,4 +1,5 @@
 import { fetch } from "frontity";
+
 class Api {
   api = "";
   isWpCom = false;
--- a/packages/wp-source/src/libraries/api.ts
+++ b/packages/wp-source/src/libraries/api.ts
@@ -1,5 +1,4 @@
-import fetch from "cross-fetch";
-
+import { fetch } from "frontity";
 class Api {
   api = "";
   isWpCom = false;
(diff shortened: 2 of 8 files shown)


packages/wp-source/package.json
```diff
@@ -27,13 +27,13 @@
     "@types/jest": "^24.0.11",
     "frontity": "^1.2.2",
     "jest": "^24.7.1",
+    "node-fetch": "^2.6.0",
```
luisherranz, line 30: This shouldn't be in devDependencies or it won't be installed when `wp-source` is installed as a dependency.

packages/wp-source/src/libraries/api.ts
```diff
@@ -1,5 +1,4 @@
-import fetch from "cross-fetch";
-
+import { fetch } from "frontity";
 class Api {
```
luisherranz, line 3: Missing empty line between `import` and `class` :)

luisherranz (changes requested):
Looks good but node-fetch should be added as a dependency.

packages/wp-source/package.json
DAreRodz: Well, `node-fetch` is only needed for tests, and for the other cases `wp-source` uses `fetch` from `frontity` package. That's the reason I left `node-fetch` in `devDependencies`.

packages/wp-source/src/libraries/api.ts
DAreRodz: Oops!

luisherranz:
By the way, "Implement" doesn't look like a GitHub word. You have to use "Fixes" or "Closes": https://help.github.com/en/articles/closing-issues-using-keywords

packages/wp-source/package.json
luisherranz: Ohh, absolutely. But then, it should be added in the `frontity` dependencies, right?

DAreRodz: @luisherranz yeah, you are right. I has been there for a while, though 😅

https://github.com/frontity/frontity/blob/bd8c6dc3d9872f1780e395c952737c867c88c0e8/packages/frontity/package.json#L48

If there's nothing else to change we can resolve this and merge the PR.

luisherranz: Nothing else on my side! Thanks @DAreRodz.

luisherranz (approved):

PR #171, opened 2019-07-31
fix(components): fixes image component

### LINKED ISSUE #169: Images not showing on Safari after navigation

Images should show after navigation in all Safari versions.

Right now, in versions lower than 12.1, are only showing at the first render.

**Additional info**

I guess the problem could be that it is not transforming `data-src` and `data-srcset` into `src` and `srcset`.
Test_Frontity_Blog

After doing some checks, maybe the problem is at the file https://github.com/frontity/frontity/blob/dev/packages/components/image.tsx#L13-L30. This code may not be executing:

```javascript
// Finds all the images rendered by this component and maps
// their `data-src` and `data-srcset` attributes to `src` and `srcset`
// when the browser doesn't support Proxy or IntersectionObserver.
const noProxyScript = `
  if (typeof window !== "undefined" && (!("Proxy" in window) || !("IntersectionObserver" in window))) {
    document.addEventListener("DOMContentLoaded", function() {
      var images = document.querySelectorAll("img.frontity-lazy-image");
      for (i = 0; i < images.length; ++i) {
        var image = images[i];
        image.setAttribute("src", image.getAttribute("data-src"));
        image.setAttribute("srcset", image.getAttribute("data-srcset"));
        image.removeAttribute("data-src");
        image.removeAttribute("data-srcset");
        image.removeAttribute("style");
      }
    });
  }
`;
````

`IntersectionObserver` is not supported by these Safari versions: https://caniuse.com/#search=IntersectionObserver.

Apart from that, at the same file, `useInView` is being used inside an `if` statement checking for `IntersectionObserver`, so it is not being executed in this case:

```javascript
if (typeof IntersectionObserver !== "undefined") {
      const [onScreen, ref] = useInView({ rootMargin: props.rootMargin, onlyOnce: true });
      if (onScreen) changeAttributes(attributes);
      return (
        <>
          <NoScriptImage {...attributes} />
          <img ref={ref} {...attributes} />
        </>
      );
}
```

orballo:

- Fixes https://github.com/frontity/frontity/issues/169 with the solution that @luisherranz proposed.

- Fixes the `changeAttributes` function, because when I wrote it I thought: "Ok, I think that just this one time I can write a non pure function to map an object without messing up". Well, it turns out that I was wrong.

- Fixes the whole hydration thing, because it wasn't working at all. Now, when we need to render a different `img` in client than in server, it will wait till the component has mounted to change it, so the hydration doesn't fail. There is still a `suppressHydrationWarning` prop to avoid the warning when React is hydrating after we changed the `img` from outside with the `script`.

- Implements a new `useDidMount` hook in `@frontity/hooks`.

- Fixes a small bug in the `FeaturedImage` styles that prevented it from loading correctly when scrolling from below.

#### My PR is a:

- [x] 🐞 Bug fix
- [ ] 🚀 New feature
- [ ] 🔝 Improvement

#### Main update on the:

- [ ] Documentation
- [x] Framework

- [x] Yes!
- [ ] Work in progress


luisherranz (changes requested):
First of all, thanks for the PR @orballo :)

----

How did you make native lazy load (NLL from now on) work in Chrome? I always get `Illegal invocation` when checking for `HTMLImageElement.prototype.loading`. And NLL is not working at all.

I'm on Chrome 76, last version. And I'm not the only one:
https://community.frontity.org/t/illegal-invocation-at-image-tsx-1-0-6/325

I thought it was a bug in Chrome itself but I wonder how you were able to made it work.

`"loading" in HTMLImageElement.prototype` returns true, tho.

----

I've taken a good look at the implementation and I like it, but I think there's one thing we can do better.

When using NLL (or no IntersectionObserver), you are using a `didMount` hook to cause a rerender. You are rendering images in two steps:
`data-src` (lazyAttributes) -> `src` (eagerAttributes)

That works great to move from SSR to CSR without a hydration warning.

But two things:
- I don't think it makes sense to maintain that behaviour once we are in CSR.
- And actually, the React team said that in React 16 they improved `React.hydrate` and having different nodes in server than in client is not a performance problem anymore.

So, I'd go directly for `src` (eagerAttributes) if we are in the client, and just suppress the hydration warning. You can check `state.frontity.platform` for `"server"` or `"client"` for that.

What do you think? Do you see any problem with that approach?

orballo:
> How did you make native lazy load (NLL from now on) work in Chrome? I always get `Illegal invocation` when checking for `HTMLImageElement.prototype.loading`. And NLL is not working at all.
> 
> I'm on Chrome 76, last version. And I'm not the only one:
> https://community.frontity.org/t/illegal-invocation-at-image-tsx-1-0-6/325
> 
> I thought it was a bug in Chrome itself but I wonder how you were able to made it work.
> 
> `"loading" in HTMLImageElement.prototype` returns true, tho.

Wow, I didn't know this was working already. I didn't try it, it was written based on the docs of NLL (a blog post or something like that was?). I'll have a look at it.

> So, I'd go directly for `src` (eagerAttributes) if we are in the client, and just suppress the hydration warning. You can check `state.frontity.platform` for `"server"` or `"client"` for that.

This won't work. It won't update the nodes if the props differ, so the server version will prevail. I also tried the trick with `dangerouslySetInnerHtml` and it didn't work either.

But, another way of solving this might be to include this scenario within the range of the script.

orballo:
@luisherranz let me know what do you think about the second solution and I'll probably work on this tonight.

orballo:
I fixed the error with the NLL but I didn't see any lazy load behaviour while using it... Would you like to check it out? See if you notice if it's working or not?

luisherranz:
> Wow, I didn't know this was working already. I didn't try it, it was written based on the docs of NLL (a blog post or something like that was?). I'll have a look at it.

You still need to activate it with a Chrome flag. But I haven't been able to make it work even with the flag. Not even other examples I found in some blog posts 😕

---- 

> This won't work. It won't update the nodes if the props differ, so the server version will prevail. I also tried the trick with dangerouslySetInnerHtml and it didn't work either.

I didn't know that, sorry. Ok then we need the double-render in the step between SSR->CSR, but we should try to avoid it in CSR.

Maybe we need to introduce our `state.frontity.rendering` again, with `"ssr"` on server and first frame of the client. That would substitute your `didMount`. Or do you have any other idea?

orballo:
> I didn't know that, sorry. Ok then we need the double-render in the step between SSR->CSR, but we should try to avoid it in CSR.

Oh, I didn't understood this the first time, sorry. Yeah, it doesn't make any sense that step once we are on the client.

> Maybe we need to introduce our `state.frontity.rendering` again, with `"ssr"` on server and first frame of the client. That would substitute your `didMount`. Or do you have any other idea?

I can't think of any other solution. This might be the way to go.

luisherranz:
> I can't think of any other solution. This might be the way to go.

Done! :) 

Feel free to approve, merge it and rebase this PR https://github.com/frontity/frontity/pull/175

It should be as simple as removing `didMount` and using `state.frontity.rendering`, right? Something like this:
```jsx
<img
  {...(state.frontity.rendering === "csr" ? eagerAttributes : lazyAttributes)}
  suppressHydrationWarning
/>
```

Or am I missing something else?

orballo:
Great! I think that should do it. Do you want me to remove the `useDidMount` hook or do you think it might be useful in the future?

luisherranz:
Initially, I don't see a `useDidMount` as a bad thing to have, even tho its implementation with `useEffect` is quite straightforward.

What I don't like it that you `useDidMount`, as it is right now, causes a double-render effect, because it's using a `setState` on `useEffect`. That was needed for your case, but not for a more general `useDidMount` use.

So I think a more general approach of a `useDidMount` would be something like:
https://github.com/imbhargav5/rooks/blob/dev/packages/shared/useDidMount.ts
which is so simple that pretty much loses the point of having a new hook in the first place...

So... yes. I think it's better to remove it 😆

orballo:
Ok, fixed. @luisherranz I'm not sure about the types on the `Image` component, I used `React.FC<Connect<Package, Props>>`, but I don't know if that's right. Let me know.

luisherranz:
> I used `React.FC<Connect<Package, Props>>`

Yes, that's actually the way to do it, because `<Image>` only needs `state.frontity.rendering` and that's defined in the `Package` types :)

luisherranz:
Code looks great to me!! 

Tomorrow I'll test it out in my local env, but I think it's perfect now.

DAreRodz (approved):
LGTM!

DAreRodz:
I don't know why but GitHub thinks @orballo didn't make the changes you requested, @luisherranz, and so merging is "blocked". I'm going to merge this PR anyway.

PR #173, opened 2019-07-31
Use pattern generation instead of redirections

DAreRodz:

Use `wp-source` settings to modify handler patterns instead of create redirections.

For example, is `state.source.tagBase` is set to `wp-tag`, then change the tag handler pattern to start with `"/wp-tag/"` instead of `"/tag/"`, and not use redirects for those cases.

#### My PR is a:

- 🔝 Improvement

- ❌ No!

### DIFF
--- a/packages/wp-source/src/actions.ts
+++ b/packages/wp-source/src/actions.ts
@@ -53,17 +53,9 @@ const actions: WpSource["actions"]["source"] = {
   },
 
   init: ({ state, libraries }) => {
-    const { api, isWpCom } = state.source;
-
-    libraries.source.api.init({ api, isWpCom });
-
-    // handlers & redirections:
-    const { handlers, redirections } = libraries.source;
-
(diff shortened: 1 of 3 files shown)


luisherranz:
> Use wp-source settings to modify handler patterns instead of create redirections

What do you mean by that?

DAreRodz:
> What do you mean by that?

I've updated the PR description.

luisherranz:
Ok, I get it now. This solution is to avoid having two valid handlers, one for `tag` and another for `wp-tag` when using `tagBase`. If we change the pattern instead of adding a redirection, then only `wp-tag` is valid.

DAreRodz:
That's the idea!

luisherranz:
I think we can close this because it doesn't seem to be a problem for Frontity users and it will change in Source v2. 

What do you think @DAreRodz?

DAreRodz:
Yeah, sure. I'll do it.

PR #191, opened 2019-09-03
feat(mars-theme): add dynamic <title> tags

luisherranz:

I've added a new component called `<Title>` that takes care of rendering the proper `<title>` tags of the `<head>`.

#### My PR is a:

- [ ] 🐞 Bug fix
- [ ] 🚀 New feature
- [x] 🔝 Improvement

#### Main update on the:

- [ ] Documentation
- [x] Framework

- [x] Yes!
- [ ] Work in progress

### DIFF
--- a/packages/mars-theme/src/components/index.js
+++ b/packages/mars-theme/src/components/index.js
@@ -5,6 +5,7 @@ import List from "./list";
 import Post from "./post";
 import Page404 from "./page404.js";
 import Loading from "./loading";
+import Title from "./title";
 
 const globalStyles = css`
   body {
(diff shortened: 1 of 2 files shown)


DAreRodz (commented):
I see that titles for date archives won't be rendered, but doing so is not complicated. It would be a good idea to add it in this PR, right?

luisherranz:
Yeah, I thought about it but then I saw the component got quite big without that and the purpose is to be an example of how to use React for populating `<title>` dynamically.

DAreRodz:
Ok, we can always add it later 👍

DAreRodz (approved):

PR #193, opened 2019-09-04
fix(wp-source): fix handlers, refactor them and improve tests

DAreRodz:

Move `getTotal` and `getTotalPages` functions to `libraries` so custom handlers can use them.

#### My PR is a:

- [ ] 🐞 Bug fix
- [ ] 🚀 New feature
- [x] 🔝 Improvement

#### Main update on the:

- [ ] Documentation
- [x] Framework

- [x] Yes!
- [ ] Work in progress


packages/wp-source/src/libraries/response-utils.ts
```diff
@@ -0,0 +1,7 @@
+export const getTotal = (response: Response): number =>
+  parseInt(response.headers.get("X-WP-Total"));
+
+export const getTotalPages = (response: Response): number =>
+  parseInt(response.headers.get("X-WP-TotalPages"));
```
luisherranz, line 5: @DAreRodz  What happens here if the X-WP-TotalPages or X-WP-Total is not present? Sometimes people delete them in Nginx.

DAreRodz: Uhm, if that is the case, I don't know from which place that values could be obtained.

Taxonomies has a `count` property that gives you the `total`value but not `totalPages`, and authors and other archives don't even have that property.

We could maybe use the `_envelope` param in the `api.get` requests and receive all response data in the body, including headers and status code, but that would change the return value of that method...

@luisherranz, any idea?

DAreRodz: I forgot to link the [docs](https://developer.wordpress.org/rest-api/using-the-rest-api/global-parameters/#_envelope) of the `_envelope` property.

luisherranz: I wasn't asking how to solve but what happens when those headers are not present. `parseInt(undefined);` and `parseInt(null);` are `NaN`. It is something we can expect and it shouldn't break the app.

luisherranz: Oh, I didn't know about the `_envelope` query. Interesting :)

luisherranz: Maybe we can make sure nothing is broken if those headers are not present right now and if people report having problems with this in the future switch to use the `_envelope` query param.

DAreRodz: Ok, I'll check everything works without those headers.

luisherranz: Thanks David.

DAreRodz:
I've fixed some handlers that were using `total` and `totalPages` to check if an out-of-range page was being fetched.

DAreRodz:
Btw, `mars-theme` and `yoast` packages depend on `totalPages` prop for handling pagination. What should we do in those packages if `totalPages` is not present?

packages/wp-source/src/libraries/handlers/author.ts
```diff
@@ -23,13 +23,13 @@ const authorHandler: Handler = async ({ route, params, state, libraries }) => {
     }
   });
 
-  // 3. throw an error if page is out of range
+  // 3. populate response and add page to data
+  const items = await populate({ response, state });
+  if (page > 0 && !items.length) throw new Error("Page doesn't exist.");
```
luisherranz: I don't quite understand what this does. Could you please explain?

luisherranz:
> Btw, mars-theme and yoast packages depend on totalPages prop for handling pagination. What should we do in those packages if totalPages is not present?

I've just checked and if `totalPages` is 0, undefined or null, pagination doesn't appear. So I guess it's fine.

packages/wp-source/src/libraries/handlers/author.ts
DAreRodz: If you fetch a page that is out of range, you get a response that could be an error message or an empty array. When that response is populated, `populate` returns an empty array because nothing was added to the state.

And if you access to the **first page** of an archive that exists but its empty, WP returns a page that says *Nothing Found* with code `200` meaning that archive has no content, but if you access the next page WP returns with code `404`.

So I check these two conditions before throwing an error.

DAreRodz: Oops, I just realized that if the author or taxonomy doesn't exist an error should be thrown. This is not happening right now.

luisherranz:
Great work.

Remember to re-request a review once you've finished.

packages/wp-source/src/libraries/api.ts
```diff
@@ -46,7 +46,7 @@ class Api {
 
   async getIdBySlug(endpoint: string, slug: string) {
     const response = await this.get({ endpoint, params: { slug } });
-    const [entity] = await response.json();
+    const [entity] = await response.clone().json();
```
luisherranz, line 49: Why do you have to clone the response?

DAreRodz: I don't fully understand the problem, but it seems that you cannot  call `json()` twice for the same `Response` object (at least it happens with [`node-fetch`](https://github.com/bitinn/node-fetch/issues/533#issuecomment-429261523)). 😕

I think it would be a good idea to change the return type of `api.get()` in the next version. At first I found useful to work directly with a `Response` object but now I'm not very convinced.

luisherranz: I'm afraid `clone` can be really slow on a big response. We should find an alternative.

luisherranz: Or take a look and see how expensive it really is.

DAreRodz: Anyway, I've been able to get rid of `response.clone` calls.

luisherranz (approved):
Let's merge it! 👍

luisherranz:
@DAreRodz give ti an appropriate name (I guess it's not only `move getTotal and getTotalPages to libraries` anymore) and squash it.

PR #198, opened 2019-09-09
refactor(wp-source): add entity props in data during populate

DAreRodz:

Generate entity data objects with `id` and `type` or `taxonomy` props when calling `populate`.

#### My PR is a:

- [ ] 🐞 Bug fix
- [ ] 🚀 New feature
- [x] 🔝 Improvement

#### Main update on the:

- [ ] Documentation
- [x] Framework

- [x] Yes!
- [ ] Work in progress


packages/wp-source/src/libraries/handlers/author.ts
```diff
@@ -42,8 +41,7 @@ const authorHandler: Handler = async ({ route, params, state, libraries }) => {
   const totalPages = getTotalPages(response);
 
   // 5. add data to source
-  Object.assign(state.source.data[route], {
-    id,
+  Object.assign(state.source.data[route], state.source.data[path], {
```
luisherranz: In case of requesting other than the first page, we are setting `isFetching` to false when it should be true.

Add a test to check that `isFetching` and `isReady` are never false until `fetch` finishes.

packages/wp-source/src/libraries/__tests__/__snapshots__/populate.tests.ts.snap
```diff
@@ -961,7 +974,132 @@ Object {
     },
   },
   "categoryBase": "",
-  "data": Object {},
```
luisherranz, line 964: Move this tests to the new fixtures.

luisherranz (changes requested):

packages/wp-source/src/libraries/handlers/author.ts
DAreRodz: Done!

packages/wp-source/src/libraries/__tests__/__snapshots__/populate.tests.ts.snap
DAreRodz: Done!

luisherranz (approved):
👏

luisherranz:
Amazing work @DAreRodz!

PR #208, opened 2019-09-19
WIP: feature(frontity/connect): Attaching context, path and state

michalczaplinski:

The new version of the state manager will need the `state`,
the `path` and `context` to be nested for each proxified state.
This is a first attempt to add this functionality.

```
proxy = new Proxy(
      {                
        state,       //
        context,   //  <-- Instead of `new Proxy(state, handlers);`
        path         //
      },
      // store the root in handlers possibly
      handlers
    );
```
@luisherranz 
Currently 3/4 of the tests are failing, so there are multiple bugs it seems. I'm pushing this only for you to have a look and see if you can debug it further.

#### My PR is a:

- 🔝 Improvement

- 🚧 Work in progress

### DIFF
--- a/packages/connect/src/__tests__/debug.tests.js
+++ b/packages/connect/src/__tests__/debug.tests.js
@@ -143,31 +143,6 @@ describe("debugger", () => {
     ]);
   });
 
-  it("should debug clear operations", () => {
-    let dummy;
-    const rawMap = new Map();
-    rawMap.set("key", "value");
-    const map = observable(rawMap);
-    const debugSpy = jest.fn(() => {});
-    observe(() => (dummy = map.get("key")), {
-      debugger: debugSpy
-    });
-
-    expect(dummy).toEqual("value");
-    expect(debugSpy.mock.calls.length).toEqual(1);
-    const oldMap = new Map(rawMap);
-    map.clear();
-    expect(dummy).toEqual(undefined);
-    expect(debugSpy.mock.calls.length).toEqual(3);
-    expect(debugSpy.mock.calls[1]).toEqual([
-      {
-        type: "clear",
-        target: rawMap,
-        oldTarget: oldMap
-      }
-    ]);
-  });
-
   it("should not cause infinite loops", () => {
     let receiverDummy;
     const rawCounter = { num: 0 };
--- a/packages/connect/jest.config.js
+++ b/packages/connect/jest.config.js
@@ -1,5 +1,6 @@
 module.exports = {
   transform: { "^.+\\.(t|j)sx?$": "ts-jest" },
   testRegex: "(/__tests__/.*|(\\.|/)(test|spec))\\.(t|j)sx?$",
-  moduleFileExtensions: ["ts", "tsx", "js", "jsx"]
+  moduleFileExtensions: ["ts", "tsx", "js", "jsx"],
+  testPathIgnorePatterns: ["/__tests__/builtIns"]
 };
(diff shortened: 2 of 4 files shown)


packages/connect/src/__tests__/debug.tests.js
```diff
@@ -143,31 +143,6 @@ describe("debugger", () => {
     ]);
   });
 
-  it("should debug clear operations", () => {
```
michalczaplinski, line 146: Removed this test because it tests a built-in object `Map` which we do not support anymore

michalczaplinski:
### Note on the changes and progress

- I haven't yet didn't remove the tests for built-in objects (like `Map`, `Set`, etc) but rather added a param in `jest.config.js` to skip all tests in `__test__/builtIns`.

- We have to be careful with the `receiver` parameter of the proxy handlers as well the `this` in general when we nest the `state` inside of our proxy. For example, in : 
https://github.com/frontity/frontity/pull/208/files#diff-347dcfa4da46e09bce4c6277d5ffa7e3L25
we **shouldn't** actually pass the `receiver` further because it points to the proxy and and not the raw state. This caused many tests to fail

- We were missing the traps for `setPrototypeOf` and `getPrototypeOf`, which caused a few tests to fail, for example:
```javascript
it("should observe properties on the prototype chain", () => {
    let dummy;
    const counter = observable({ num: 0 });
    const parentCounter = observable({ num: 2 });
    Object.setPrototypeOf(counter, parentCounter);
    observe(() => (dummy = counter.num));

    expect(dummy).toEqual(0);
    delete counter.num;
    expect(dummy).toEqual(2); // <-- dummy is `undefined`
  });
```
   
- We had a similar problem with iteration. Had to add the `setPrototypeOf` and `getPrototypeOf`, traps.

- Finally, all the tests are fixed, but it seems that this change has broken snapshot testing in jest with:

```
TypeError: 'getOwnPropertyDescriptor' on proxy: trap reported non-configurability for property 'length' which is either non-existent or configurable in the proxy target
```
which I think is due to how jest uses `this` to get the object from context when using `expect(someObject).toMatchSnapshot()` but I didn't manage to debug this yet.

packages/connect/src/__tests__/debug.tests.js
luisherranz: 👍

luisherranz:
> TypeError: 'getOwnPropertyDescriptor' on proxy: trap reported non-configurability for property 'length' which is either non-existent or configurable in the proxy target

I'm debugging that error.

I've added this test to isolate the problem in `connect`:
```js
  it("return [object Array] for arrays", () => {
    const arr = [];
    const obs = observable(arr);
    expect(obs).toEqual([]);
  });
```

The `toEqual` fails in this line, when comparing the array with the proxified array:
https://github.com/facebook/jest/blob/master/packages/expect/src/jasmineUtils.ts#L100

`Object.prototype.toString.call(obs)` returns `"[object Object]"` and `Object.prototype.toString.call([])` returns `"[object Array]"`.

I don't know yet why the test is throwing with the TypeError instead of just failing because of the assertion.

EDIT: I've just check the previous implementation and `Object.prototype.toString.call(observer([])` returns `"[object Array]"`. My guess is that Object.prototype.toString is inspecting the target. I'll try to find if we can create a trap for that.

michalczaplinski:
The last 2 commits belong to another branch with a different implementation. Closing and continuing with another implementation in #213

PR #211, opened 2019-09-20
chore(type-declarations): create a type declarations package

DAreRodz:

This is an attempt to declare TypeScript modules for imported files like images, fonts and so on, in a way that Frontity users don't have to do it.

Right now is not working.

#### My PR is a:

- 🔝 Improvement

- ✅ Yes!

### DIFF
--- a/packages/frontity/package.json
+++ b/packages/frontity/package.json
@@ -37,6 +37,7 @@
     "@emotion/styled": "^10.0.11",
     "@frontity/connect": "^1.0.2",
     "@frontity/types": "^1.1.1",
+    "@frontity/type-declarations": "^1.0.0",
     "@loadable/component": "^5.10.1",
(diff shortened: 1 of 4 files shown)


luisherranz:
Why do you say it's not working? I can see the module if I add an image to mars-theme:
Screen Shot 2019-09-23 at 11 06 37

Besides that, I think you need to add the package to `frontity` dependencies, so the package is downloaded on an external Frontity project.

luisherranz:
By the way, please use GitHub's "draft PR" feature when a PR is still a WIP.

luisherranz:
Once this is ready for review please remove the WIP from the title and description and re-request a review or ping me in the comments.

luisherranz (approved):
Let's see if this works in a Frontity project!

PR #213, opened 2019-09-23
WIP: (@frontity/connect) use deep overwrite 

michalczaplinski:
Update the `@frontity/connect` implementation to use a "deepOverwrite" strategy described in https://community.frontity.org/t/frontity-connect-middleware-support/593/15

WIP, further things to do in order

1. [x] ~~store root on the `proxy[ROOT]` https://community.frontity.org/t/frontity-connect-middleware-support/593/16 (I'm not sure if this is a good idea, because it seems that it messes up with object serialization (at least in jest tests, see the failing test case). Let's just store it in a global? 🤪 (I'm actually serious :) )~~
2. [x] fix the failing tests for `wp-source`
3. [ ] concatenate the object keys in order to get the full path for each reaction https://community.frontity.org/t/frontity-connect-middleware-support/593/19?u=mmczaplinski
4. [ ] add a benchmark to compare with mobx-state-tree / react-easy-state
5. [ ] possibly rewrite with plain `for` loops for additional extra performance
6. [ ] add tests for:
    - attaching actions and context
    - whether the reactions are triggered correctly for each leaf in the state tree (for each data type of primitive, Object & array)
    - array item deletion (splice)
    - substitution of arrays for objects and objects for arrays
7. [ ] emit a warning if user puts anything else than object, array or primitive in the state
8. [ ] remove code related to built-ins

## Improvements:
As is, the code probably will need some perf improvements. For example, in the worst case scenario, we will have to visit every node in the state tree for a single (deeply nested) update. This would happen if the user would replace the root of the tree! Such traversal is `O(N)` where where N is the number of nodes. We could compute a hash for frequently visited nodes and then compare the hashes instead of recursing on every branch of the tree. Or some variation of this technique. 

At the end of the day, I think it is probably not that bad and if the user chooses to return **new** objects instead of mutating, then we cannot stop them from shooting themselves in the foot :D And even in that case, I think we're basically at the same place where pure redux would leave us and a user can always implement selectors or `memo` themselves (if for whatever reason they cannot just mutate the state). 

#### My PR is a:
- 🔝 Improvement

- 🚧 Work in progress

### DIFF
--- a/packages/connect/src/__tests__/observable.tests.js
+++ b/packages/connect/src/__tests__/observable.tests.js
@@ -84,6 +84,70 @@ describe("observable", () => {
     expect(obs.nested1.prop3(3)).toEqual(4);
     expect(obs.nested1.prop4(2)).toEqual(8);
   });
+
+  it("iterating an object", () => {
+    const obj = { a: "a", b: "b", c: "c" };
+    const obs = observable(obj);
+
+    for (let key in obs) {
+      expect(obs[key]).toEqual(obj[key]);
+    }
+  });
+
+  it("should allow making arrays observable", () => {
+    const arr = [];
+    const obs = observable(arr);
+    expect(isObservable(obs)).toEqual(true);
+  });
+
+  it("should update the observable when adding new properties", () => {
+    const arr = [];
+    const obs = observable(arr);
+
+    obs.push(1);
+    expect(obs[0]).toEqual(1);
+
+    obs.push({ name: "Jon Snow" });
+    expect(isObservable(obs[1])).toEqual(true);
+    expect(obs[1].name).toEqual("Jon Snow");
+  });
+
+  it("should allow iterating array", () => {
+    const arr = [{ name: "a" }, { name: "b" }, { name: "c" }];
+    const obs = observable(arr);
+
+    let i = 0;
+    for (let key of obs) {
+      expect(key).toEqual(arr[i]);
+      i++;
+    }
+  });
+
+  it("length works correctly", () => {
+    const arr = [];
+    const obs = observable(arr);
+
+    expect(obs.length).toEqual(0);
+
+    obs.push("hello");
+    expect(obs.length).toEqual(1);
+  });
+
+  it("state can access the property `state`", () => {
+    const obj = { state: { user: "test" } };
+    const obs = observable(obj);
+
+    expect(obs.state).toEqual(obj.state);
+  });
+
+  it("Object.keys works", () => {
+    const obj = { a: 1, b: 2 };
+    const obs = observable(obj);
+
+    const keys = Object.keys(obs);
+
+    expect(keys).toEqual(["a", "b"]);
+  });
 });
 
 describe("isObservable", () => {
--- a/packages/connect/src/__tests__/observable.tests.js
+++ b/packages/connect/src/__tests__/observable.tests.js
@@ -183,3 +183,104 @@ describe("raw", () => {
     expect(raw(12)).toEqual(12);
   });
 });
+
+describe("Test object references", () => {
+  const context = {
+    id: 1,
+    action: "action-1",
+    path: "action.action-1",
+    triggeredBy: {
+      type: "component",
+      name: "Post"
+    }
+  };
+
+  let proxy, state;
+  beforeEach(() => {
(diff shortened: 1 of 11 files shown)


luisherranz:
Looks promising 👏👏

I'll take a look at the failing tests of `wp-source` tomorrow.

luisherranz:
> (I'm not sure if this is a good idea, because it seems that it messes up with object serialization (at least in jest tests, see the failing test case).

I think we can store it as a non-enumerable prop. Let's try for now and see if it gives us any problem in the future.

> fix the failing tests for wp-source (partly done)

This is really interesting because we need to make a decision about asynchronism. I'll comment that in the [forum topic](https://community.frontity.org/t/frontity-connect-middleware-support/593/20).

michalczaplinski:
> I think we can store it as a non-enumerable prop. Let's try for now and see if it gives us any problem in the future.

Right! Although I thought that Symbols were supposed to be non-enumerable by default (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Symbols_and_for...in_iteration) I guess there's a caveat... 🤷‍♂

michalczaplinski:
@luisherranz I have come across an issue that we've missed previously while trying to get the peformance tests to work. We are currently iterating over arrays/objects and recursively reassigning values to original keys with the `deepOverwrite`. The problem is that we have to be very careful about what data types we operate on. If 

First of all, we need to `arr.splice(key, 1)` when removing things from arrays because `delete arr[key]` will result in sparse arrays and incorrect `arr.length`

Secondly, if `A` is an array and `B` is an object, or vice versa, we are bound to end up with inconsistent data because we will have arrays with "stringy" keys, etc.

I started working on fix but this is as far as I got for today - it's a work in progress.

michalczaplinski:
I think I've fixed the issue I mentioned in https://github.com/frontity/frontity/pull/213#issuecomment-534818360 but in general this needs some refactoring for clarity and edge cases

packages/connect/src/handlers.js
```diff
@@ -164,9 +164,12 @@ function set(target, key, value, receiver) {
   }
 
   // If both the old value and the new value are objects OR if both are arrays, deepOverwrite.
+  // TODO: this should probably be refactored
   if (
-    isObject(oldValue) === isObject(value) ||
-    Array.isArray(oldValue) === Array.isArray(value)
+    !isPrimitive(oldValue) &&
+    !isPrimitive(value) &&
```
luisherranz, line 170: Why do we need this?

luisherranz:
Good work with the arrays 👍 

We should add tests for that in `connect`.

packages/connect/src/handlers.js
michalczaplinski: I think this should make it a bit clearer:
A and B are the old and new values respectively.

| A  \ B   | primitive | array    | object   |
| ------- | -------   | ------- | --------|   
|primitive|   reassign|  reassign| reassign|
| array  |   reassign |  **deepOverwrite** |reassign|
| object  |   reassign|  reassign|**deepOverwrite**|

previously, we used to also do a deepOverwrite if one was an array and the other was an object and vice-versa, but this is wrong for the reasons I mentioned in https://github.com/frontity/frontity/pull/213#issuecomment-534818360

packages/connect/src/__tests__/observable.tests.js
```diff
@@ -183,3 +183,104 @@ describe("raw", () => {
...
+    proxy.allUsers.push({ name: "Arya", surname: "Stark" });
+    expect(state.allUsers[1].name).toBe("Arya");
+    expect(proxy.allUsers[1].name).toBe("Arya");
+  });
+
+  test("ownKeys work", () => {
+    expect(Object.keys(state)).toEqual(["number", "allUsers", "user"]);
+    expect(Object.keys(state.user)).toEqual(["name", "surname", "location"]);
+  });
+});
```
michalczaplinski: We can add those test cases to the new branch

michalczaplinski: @michalczaplinski

luisherranz:
Closing this for now. We'll review them later once we are back with a final API for the connection middleware.

PR #217, opened 2019-10-02
fix(core): delete package name from merged store

### LINKED ISSUE #95: Delete `frontity.name`
When we merge all the stores together the name also gets merge and one random name ends up in the store. We should remove it.

luisherranz:

After the merge, the name of a random package was appearing in the store and showing when people use `frontity` in the browser console.

It doesn't appear now.

Fixes #95.

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes!


DAreRodz (approved):
👍

PR #225, opened 2019-10-16
Add comments handler

SantosGuillamot:

The idea of this PR is to add a comments handler so users can fetch the comments of a post easily. In this implementation users should use `fetch("comments/:postId")` and it'll populate `state.source.comment[commentId]` with all the comments and `state.source.data["comments/:postId"]` with an array of the comments of the post.

I'm just starting and take a deeper look. In fact, it didn't pass the test :smile: Here are some questions I have right now and things I have to consider:
- Pass the tests.
- Should we create a schema for `comments` or reuse `postType`? I've created a new one, but not sure if it's necessary.
- At `handlers/comments.ts` I have to understand better and modify the 1st point, where we search id in state. Should we do something similar looking for the postId?
- Should we add hierarchy at data so it's easier to know the replies of comments?
- Still have to test it at WP.com.
- Have to take a look at pages and number of comments.
- Have to take a look at the errors.
- I wasn't sure what data to add: isComments / areComments / nothing?
- I wasn't sure when to use comment or comments while declaring the variables :)

#### My PR is a:

- 🚀 New feature

- 🚧 Work in progress

### DIFF
--- a/packages/source/src/data.ts
+++ b/packages/source/src/data.ts
@@ -4,6 +4,8 @@ export type EntityData = {
   type: string;
   id: number;
   link: string;
+  parent: number;
+  children: Array<Object>;
 };
 
 export type Data =
--- a/packages/wp-source/src/libraries/handlers/comments.ts
+++ b/packages/wp-source/src/libraries/handlers/comments.ts
@@ -7,47 +7,45 @@ const commentsHandler: Handler = async ({
   libraries
 }) => {
   const { api, populate, parse, getTotal, getTotalPages } = libraries.source;
-  const { path, page, query } = parse(route);
+  const { page, query } = parse(route);
   const { postId } = params;
 
-  // 1. search id in state or get it from WP REST API
-  let { id } = state.source.get(path);
-  if (!id) {
-    // Request author from WP
-    const response = await api.get({
-      endpoint: "comments",
-      params: { post: postId }
-    });
-    const [entity] = await populate({ response, state });
-    if (!entity)
-      throw new Error(
-        `entity from endpoint "comments" with postId "${postId}" not found`
-      );
-    id = entity.id;
-  }
-
-  // 2. fetch the specified page
+  // 1. fetch the specified page
   const response = await api.get({
     endpoint: "comments",
     params: {
       post: postId,
       search: query.s,
       page,
-      _embed: true,
       ...state.source.params
     }
   });
 
-  // 3. populate response
-  const items = await populate({ response, state });
+  // 2. populate response
+  const rawItems = await populate({ response, state });
+  const parentItems = {};
+  rawItems.map(item => {
+    parentItems[item.parent] = [...(parentItems[item.parent] || []), item];
+  });
+  const items = parentItems[0];
+  const addChildren = arr => {
+    arr.map(item => {
+      if (parentItems[item.id]) {
+        item.children = parentItems[item.id];
+        addChildren(item.children);
+      }
+    });
+  };
+  addChildren(items);
+
   if (page > 1 && items.length === 0)
     throw new Error(`comments for post "${postId}" don't have page ${page}`);
 
-  // 4. get posts and pages count
+  // 3. get posts and pages count
   const total = getTotal(response);
   const totalPages = getTotalPages(response);
 
-  // 5. add data to source
+  // 4. add data to source
   const currentPageData = state.source.data[route];
 
   Object.assign(currentPageData, {
@@ -55,7 +53,7 @@ const commentsHandler: Handler = async ({
     items,
     total,
     totalPages,
-    isComments: true
+    areComments: true
   });
 };
 
--- a/packages/wp-source/src/libraries/handlers/comments.ts
+++ b/packages/wp-source/src/libraries/handlers/comments.ts
@@ -24,12 +24,12 @@ const commentsHandler: Handler = async ({
   // 2. populate response
   const rawItems = await populate({ response, state });
   const parentItems = {};
-  rawItems.map(item => {
+  rawItems.forEach(item => {
     parentItems[item.parent] = [...(parentItems[item.parent] || []), item];
   });
   const items = parentItems[0];
   const addChildren = arr => {
-    arr.map(item => {
+    arr.forEach(item => {
       if (parentItems[item.id]) {
         item.children = parentItems[item.id];
         addChildren(item.children);
--- a/packages/wp-source/src/libraries/handlers/comments.ts
+++ b/packages/wp-source/src/libraries/handlers/comments.ts
@@ -8,7 +8,7 @@ const commentsHandler: Handler = async ({
 }) => {
   const { api, populate, parse, getTotal, getTotalPages } = libraries.source;
   const { page, query } = parse(route);
-  const { postId } = params;
+  const postId = parseInt(params.postId);
 
   // 1. fetch the specified page
   const response = await api.get({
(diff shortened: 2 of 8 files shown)


luisherranz:
Code looks great so far!

## Handler

Maybe this handler could belong to the `@frontity/wp-comments` package because people not using native wp comments won't need it, right? @DAreRodz what's your opinion?

## Schema

> Should we create a schema for comments or reuse postType? I've created a new one, but not sure if it's necessary.

Yes, a comments schema is needed because comment is not a post type.

@DAreRodz how can schemas be extended? Or how is the supposed way to use your own schemas in your custom handler?

About `_embed`: I see you haven't added it in the handler or the schema. Comments seem to have two embeds: `author` and `up` and neither of those are needed because the author info is already included in the comment and `up` is the parent comment, which you are going to have anyway. So, I think it's ok as it is.

EDIT: Maybe this is not needed because there's nothing to normalize!

## Types

We need to take a look at the `source` types and see how we can extend those with the new types.

## Pagination

> Have to take a look at pages and number of comments.

I'd try to be as compatible as possible to WP. 

WP has a setting in "Discussion" to break comments in pages:
Screen Shot 2019-10-20 at 13 42 52

It looks like that that param is not used in the REST API and the default 10 is used:
http://wptest.io/demo/2013/01/03/comments/
http://wptest.io/demo/wp-json/wp/v2/comments?post=149 (only 10 comments)

So I'd make it our own setting:
```js
{
  name: "wp-comments",
  state: {
    breakInPages: 50
  }
}
```

- If `breakInPages` is not present, `actions.source.fetch("comments/:postId")` should contain all the comments for that post.
We can use `per_page: 100` for that, but it also means that if there are more than 100 comments, we'd have to fetch the next pages because `per_page` cannot be greater than 100.
The headers `total` and `totalPages` are present so that should be no problem.
- If `breakInPages` is present we should use it to populate `per_page` and allow fetching additional pages with `actions.source.fetch("comments/:postId/page/:page")`.

Either way, if the post with that ID is already fetched (quite probable) then the first 10 comments will be already in `_embeds.replies`. We could populate those and fetch the next ones using the `offset` param.

## State

> Should we add hierarchy at data so it's easier to know the replies of comments?

Yes! Something like this:
```javascript
state.source.data["comments/123"] = {
  areComments: true
  items: [
    {  type: "comment", id: 2 },
    {  type: "comment", id: 3 },
    {  type: "comment", id: 4, children: [
      { type: "comment", id: 5 },
      { type: "comment", id: 6 },
      { type: "comment", id: 7, children: [
        { type: "comment", id: 8 },
        { type: "comment", id: 9 },
      ] },
    ] },
    {  type: "comment", id: 10 },
    {  type: "comment", id: 11 },
  ]
};
```

> I wasn't sure what data to add: isComments / areComments / nothing?

`isComments` is semantically wrong so I guess something like `areComments` or `arePostComments`.

## Errors

> Have to take a look at the errors.

@DAreRodz right now, if there is an error fetching that data, it'll end up having a 404, which is wrong. 

Have you thought how to deal with errors for non-URL handlers?

EDIT: Maybe we can deprecate `data.is404` for a more general `data.isError` and add the error message in `data.error`.

DAreRodz:
> Maybe this handler could belong to the @frontity/wp-comments package because people not using native wp comments won't need it, right? @DAreRodz what's your opinion?

Agree, it's better to have this code in a different package. Installing a new Frontity package is quite easy.

> @DAreRodz how can schemas be extended? Or how is the supposed way to use your own schemas in your custom handler?
>
> EDIT: Maybe this is not needed because there's nothing to normalize!

Maybe you are right! The main idea of schemas is to normalize the REST API responses, but `postType` schema just returns the same object it receives. It doesn't change anything. So, unless we want to normalize some prop of comment objects I would use the same schema (changing the name from `postType` to something more general, I guess).

> @DAreRodz right now, if there is an error fetching that data, it'll end > up having a 404, which is wrong.
> Have you thought how to deal with errors for non-URL handlers?
>
> EDIT: Maybe we can deprecate data.is404 for a more general data.isError and add the error message in data.error.

I hadn't thought about that yet, but a `data.isError` looks like a good idea to me. I'll think about it.

SantosGuillamot:
I've worked a bit more on the code following your suggestions. I am not used to GitHub (neither to code :smile: ), so it's possible I've messed up something. If so, just tell me please. This is what I've done and some other things:

* In the `comments` handler I've deleted the part where you usually search the id in state. This is not needed in this case right?

https://github.com/frontity/frontity/pull/225/commits/e315e970d4b7e15281fa0a21388520fa3ae0a3b5#diff-caf762976285c7a6b87a27cd4d01287aL13-L27
* I've deleted the `_embed`, it was included and it seems it's not necessary.

> About _embed: I see you haven't added it in the handler or the schema. Comments seem to have two embeds: author and up and neither of those are needed because the author info is already included in the comment and up is the parent comment, which you are going to have anyway. So, I think it's ok as it is.
* I've had some troubles to do the children items nesting. I've came up to this solution but it doesn't seem optimize, so any feedback here is welcomed. This is the code I wrote:
```javascript
 // 2. populate response
  const rawItems = await populate({ response, state });
  const parentItems = {};
  rawItems.forEach(item => {
    parentItems[item.parent] = [...(parentItems[item.parent] || []), item];
  });
  const items = parentItems[0];
  const addChildren = arr => {
    arr.forEach(item => {
      if (parentItems[item.id]) {
        item.children = parentItems[item.id];
        addChildren(item.children);
      }
    });
  };
  addChildren(items);
``` 
In order to do so, I had to add `parent` and `children` to `EntityData`, not sure if that's the correct way to do it.
* I've also changed the prop `postId` to be a number. Before it was a string.
* I've changed this to `areComments`.
> isComments is semantically wrong so I guess something like areComments or arePostComments.
* WordPress shows older comments at the top of the page by default, and there is a setting to change it to the newer ones. However, the API shows the newer at the top. I think we should change that behavior to adapt to WordPress. We could add a default param `order=asc` by default and make our own setting, as we do with `breakInPages`.
* I took a quick look at the tests the code isn't passing and I am not sure if adding `parent` and `children` to `EntityData` is causing some of that errors. Apart from that, I think that not having something as `isArchive` or `isPage` in the data is generating errors too. Is this possible?
* If we finally move all the code to a new package it'd be great some guidance on how to do so :grin:

luisherranz:
> In the comments handler I've deleted the part where you usually search the id in state. This is not needed in this case right?

I think it is needed, because if you don't check if those comments have already been fetched, you'll fetch them over and over again each time the `actions.source.fetch` action is called.

We want the refetch to be explicit, like this:
```js
actions.source.fetch("comments/123", { force: true });
```

This is not implement yet, but it's the way it should work.

> I've also changed the prop postId to be a number. Before it was a string.

Why do you need that? @DAreRodz, is `postId` usually a string?

> I've had some troubles to do the children items nesting. I've came up to this solution but it doesn't seem optimize, so any feedback here is welcomed.

The best way to work on this type of algorithm is with unit testing. We can help you out to set up some tests if you want, then you can try to refactor/improve your algorithm until the tests pass again.

> WordPress shows older comments at the top of the page by default, and there is a setting to change it to the newer ones. However, the API shows the newer at the top. I think we should change that behavior to adapt to WordPress. We could add a default param order=asc by default and make our own setting, as we do with breakInPages.

Agree. Let's mimic what WP does by default and add a setting for that.

> If we finally move all the code to a new package it'd be great some guidance on how to do so

Sure, we'll help you out.

luisherranz:
@darerodz, now that you're working on this on another branch, please feel free to get what you need from here and then close this PR.

DAreRodz:
Closing this as it was finally implemented in https://github.com/frontity/frontity/pull/521.

PR #227, opened 2019-10-23
fix(core): add labels and source maps

luisherranz:

It looks like emotion labels and source maps depend on finding the emotion imports in the files, like this:
```js
import styled from "@emotion/styled";
import { css } from "@emotion/core";
```

So I have created a babel plugin that runs before emotion changing from our imports to their imports.

From this:
```js
import { styled } from "frontity";
```
to this:
```js
import styled from "@emotion/styled";
```
with all the emotion utils: `Global`, `css`, `styled` and `keyframes`.

#### My PR is a:

- 🐞 Bug fix

- ✅  Yes!

## Tasks

- [x] Create `babel-plugin-frontity`.
- [x] Add unit tests. 
- [x] Comment code.
- [x] Add e2e tests.
- [x] Add types.
- [x] Remove `extractCritical` from `@frontity/core`.

Source maps don't work with `extractCritical` so I think we can remove it by default, and let the new `@frontity/amp` package add it manually again.


luisherranz:
@DAreRodz could you please take a look at this? I'd love to release a new version tomorrow.

DAreRodz (approved):
Amazing job, @luisherranz! 👌

jcklpe:
@luisherranz  Hey I know this got merged yesterday but just wanted to run a suggestion by ya'll. I think this is actually more of an emotionjs issue and I'll log a feature request there but thought I'd say it here too:

Instead of `class="css-jdkslf-Item"` :

image
Is it possible to make it something more like `class="Item-kfjdslf"`?

It's just easier to scan for the name of the element at the beginning of the class name rather than at the end. 

And also further, is it possible to do something other than the "css-jfkdslj". I understand that's probably done for namespacing reasons to make sure that if someone has multiple stateless components in their project named something like "container" they don't end up with css conflicts, but I feel like there's got to be a more user friendly output than a bunch of random letters and numbers. It introduces a lot of extra noise to DOM visual scan.

luisherranz:
Yes, I think that can be configured with the `labelFormat` option.
https://emotion.sh/docs/babel-plugin-emotion#labelformat

@DAreRodz do you want to take a look into this? I think it's a good idea.

You don't need to add the `babel-plugin-emotion`, you can add the `labelFormat` setting in the `babel-preset-css-prop` we are already using, like explained here:
https://emotion.sh/docs/@emotion/babel-preset-css-prop

Once added, you have to build `@frontity/core` again with `npm run build` to test it out in the mars example. And update the snapshots of `core` with `npm test`.

DAreRodz:
Yup, I'll take a look.

DAreRodz:
I just took a look at the [`labelFormat`](https://github.com/emotion-js/emotion/tree/master/packages/babel-plugin-emotion#labelformat) option and it doesn't allow to modify the class name but the _label_ part only:

> This format only affects the label property of the expression, meaning that the css prefix and hash will be prepended automatically.

We don't seem to be able to fix that ourselves.  🙁

luisherranz:
Ok, thanks for checking it out @darerodz.

@jcklpe, maybe you can ask in the emotion repository instead.

jcklpe:
thanks @luisherranz  and @DAreRodz  have done!

PR #230, opened 2019-10-30
Fix handlers if a query is present

### LINKED ISSUE #229: Pages with queries return 404
**_Bug report_**

This URL should return the about-us page:
https://mars.frontity.org/about-us/?test=test

It returns a 404.

----

We should add e2e tests for this and ensure it doesn't happen in other type of URLs.

DAreRodz:

Closes #229.

#### My PR is a:

- 🐞 Bug fix

- Yes

### DIFF
--- a/packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/author.tests.ts.snap
+++ b/packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/author.tests.ts.snap
@@ -461,6 +461,182 @@ Object {
 }
 `;
 
+exports[`author is requested with any query param 1`] = `
+Object {
+  "api": "https://test.frontity.io",
+  "attachment": Object {
+    "1": Object {
+      "author": 1,
+      "id": 1,
+      "link": "/post-1/attachment-1/",
+      "slug": "attachment-1",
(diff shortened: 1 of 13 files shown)


DAreRodz:
I've just fixed the post type handler and added tests, but the others need to be checked as well. I'm working on it right now.

DAreRodz:
The other handlers work as expected. 👍

DAreRodz:
> @luisherranz said: We should add e2e tests for this and ensure it doesn't happen in other types of URLs.

That means a _dockerized_ WordPress or something is needed for the e2e tests, right? Or could we use https://test.frontity.io instead?

luisherranz (dismissed):
> The other handlers work as expected. 👍

Even tho they work as expected, could you please add unit tests so we don't break them in the future?

>  That means a dockerized WordPress or something is needed for the e2e tests, right? Or could we use https://test.frontity.io instead?

You can use Cypress network stubbing with fixtures for that.

https://docs.cypress.io/api/commands/server.html
https://docs.cypress.io/api/commands/route.html#Fixtures

luisherranz:
I see! Ok, don't worry then. Merge this and we'll figure it later.

PR #235, opened 2019-11-12
Add Twenty Twenty theme

segunadebayo:
Add Twenty Twenty theme to package, taking ideas and styles from `ma.tt` site

#### My PR is a:
- 🚀 New feature

- 🚧 Work in progress


luisherranz:
I've pushed some commits to fix some CSS problems and React warnings.

Other suggestions:
- I'd add `max-width: 100%;` to content images (this will also be done by the `@frontity/gutenberg` package, but it's a good idea to support non-gutenberg images as well).
Screen Shot 2019-11-26 at 09 41 17

- A lot of components from `list` are used in `Post`. I'd refactor those in its own folder. All these:
Screen Shot 2019-11-26 at 09 43 47

- Pages don't work because `getTags` tries to access `items.tags.length` and `items.tags` is undefined. I'd actually remove `getTags` and `getCategories` and I'd try to be more explicit in the code, even if the final code is longer. It'd be easier to read for a new user.

For example, here, the new users have to look for the `hasTags` variable, then find `getTags` and finally look inside the function and understand what it gets and what it returns; whereas the `item.tags` has all the info they need:

```jsx
// I would do
{item.tags && <PostTags tags={item.tags} />}
// instead of
{hasTags && <PostTags tags={tags} />}
```

_I'd check for `item.tags.length > 0` inside `<PostTags>`._

- I'd wrap the `Header` and `Main` with `min-height: 100vh;` to leave the `Footer` below the fold when the `Loader` appears. Like we did with the [TwentyNineteen](https://github.com/imranhsayed/frontity-twentynineteen/blob/master/packages/twentynineteen-theme/src/components/index.js) theme. Real theme [here](https://twentynineteen.frontity.org/).

- I'd change the `Loader` for another one, so not all the Frontity themes have the same loader 😆 

These are just suggestions. Feel free to follow only the ones you like.

luisherranz:
More suggestions:

- Add an `autoPrefetch` feature to the theme. We can automatically prefetch all the links available in the page at a given moment, using a `useEffect` in the `<Link>` component. That way, *every click will be instant*.

- Add code-splitting for both lists and posts. Then, prefetch the post code on the list component and the list code on the post component.

- Add height to `FeaturedImage` so text in the home doesn't jump when the image loads.

segunadebayo:
I've completed most of the suggestions. Code splitting is the only thing left.

luisherranz:
> I've completed most of the suggestions.

Amazing work Segun! :)

> Code splitting is the only thing left.

I've found the bug and I'll work today on the fix.

But in order to make it work, the code needs some refactoring. I'd also take the opportunity to better organize the files.

I've pushed a [new branch](https://github.com/frontity/frontity/tree/twentytwenty-codesplitting-refactoring/packages/twenty-twenty-theme/src/components) with an [example of how you could organize the code](https://github.com/frontity/frontity/commit/7be6882f82e957152e3650be1bd8777097dd8c6f). It's not meant to be merged, just to serve as an example. Things that, in my opinion, could be improved:
- General styled-components, like `SectionContainer`, can live in their own files. It feels to me like that component lives now in the first place where it was used it, and the rest of the app is coupled to that decision. In the branch, I only moved this component but similar logic can be applied to other general styled-components. The global styles could also be moved to a new `styles` folder location. Another option would be to have `styles.js` files on each folder, with the general styled-components of those components. Then a `styles.js` file on the root for the most general ones (used by components in several folders).
- Most of the post components are in the list folder. I feel that's somehow counterintuitive and I'd move them to either a shared folder (between list and post) or to the post folder.
- The file of the component we want to code-split (`list`) contains a named export `Header`. We need to move that to its own file, `list-header` for example because statically importing it breaks the code-split.
- The difference between `Article` and `Post` is not clear. I'd either make it more apparent (maybe with its own folder) or I'd remove the extra "article" concept.

luisherranz:
I've made a PR for the code splitting. Once it's merged we can rebase this PR. https://github.com/frontity/frontity/pull/238

luisherranz:
I've merged the PR with the fix. You can rebase now.

luisherranz:
Awesome work. Code looks much more organized now. And I love the new loader! This is going to be like the best theme ever :)

Please rebase the branch on top of dev and then, if you want, we can jump to pair-programming to finish a few things:
- Add a setting for `autoPreFetch`: `"no"`, `"hover"` and `"view"`, `"all"`.
- Add a setting for `fontSubSet`: `"us-ascii"`, `"latin"`, `"all"`.
- Add code-splitting to `archive`, `post` and `search` and prefetch of code.
- Make `showAllContentOnList` and `showSearchInHeader` work.

luisherranz (approved):
Awesome work @segunadebayo. This is the best Frontity theme ever 🎉 🎉 🎉

PR #242, opened 2019-12-04
Create @frontity/head-tags package

DAreRodz:

This PR adds a new package called `@frontity/head-tags` that integrates with **REST API - Head Tags by Frontity** WordPress plugin and insert head tags for each page using the `<Head>` component.

The URLs pointing to WordPress blog pages (i.e. pages that not contain `wp-json`,  `wp-content`, `wp-admin`, etc.), would be converted to Frontity URLs.

#### My PR is a:

- 🚀 New feature

- ✅ Yes!


packages/head-tags/src/__tests__/state.tests.ts
```diff
@@ -0,0 +1,540 @@
...
+    // Populate router state.
+    state.router.link = "/post-1/";
+  };
+
+  test("returns an empty array if `head_tags` is undefined", () => {
+    const headTags = undefined;
+    // Populate all state.
+    setUpState(store.state, headTags);
+    // Test current head tags.
+    expect(store.state.headTags.current).toMatchSnapshot();
```
luisherranz: I would use `toEqual([])` here to be more explicit in what it expects.

packages/head-tags/src/components/index.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import { connect, Head } from "frontity";
+import { Connect } from "frontity/types";
+import HeadTagsPackage from "../../types";
+
+// Render all head tags from the current entity.
+const Root: React.FC<Connect<HeadTagsPackage>> = ({ state }) => {
+  return (
+    <Head>
+      {state.headTags.current.map(
```
luisherranz: You can use `useMemo` for this computation (https://reactjs.org/docs/hooks-reference.html#usememo).

packages/source/src/entities.ts
```diff
@@ -7,7 +10,15 @@ interface BaseEntity {
   _links?: any;
```
luisherranz: I think it's safer to use `unkwown` than `any`.

.changeset/serious-bears-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontity/wp-source": patch
```
luisherranz: This should be a minor.

luisherranz (changes requested):
Great work @DAreRodz. I've added a few comments in the code but looks great :)

packages/head-tags/src/utils/index.ts
```diff
@@ -0,0 +1,153 @@
...
+};
+
+const getWpUrl = (api: string, isWpCom: boolean): URL => {
+  const apiUrl = new URL(api);
+  if (isWpCom) {
+    const { pathname } = apiUrl;
+    return new URL(pathname.replace(/^\/wp\/v2\/sites\//, "https://"));
+  }
+  // Get API subdirectory.
+  const apiSubdir = apiUrl.pathname.replace(/\/wp-json\/?$/, "/");
```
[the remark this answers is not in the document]
DAreRodz: I think we misunderstood `state.source.subdirectory` the other day, because it's the Frontity site's subdirectory and not the WordPress site's, so we would have to add a new setting for that.

packages/head-tags/src/utils/index.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export const useFrontityLinks = ({
+  state,
+  headTags
+}: {
+  state: State;
+  headTags: HeadTags;
+}) => {
+  // The site URL.
+  const frontityUrl = state.frontity.url;
```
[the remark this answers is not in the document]
luisherranz: What about using a default export with a function, like:
```js
import msg from "@frontity/error";

console.warn(msg("Something is not quite right"));
throw new Error(msg("Something went really bad"));
```

luisherranz (approved):
Awesome! Looks great :)

Let's remove `@frontity/message` package and add it to the roadmap to think more about the different use cases.

luisherranz (approved):

PR #244, opened 2019-12-08
Add GitHub Actions

luisherranz:

In this PR I'm going to experiment with GH actions. If I'm able to make it work, we'll switch from Travis.

#### My PR is a:

- 🔝 CI Improvement

- Yes!


.github/workflows/e2e-tests-all.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Install dependencies (root)
+        run: npm ci
+        env:
+          CI: true
+
+      - name: Run e2e tests (all)
+        uses: cypress-io/github-action@v1
+        with:
+          start: npm run e2e:prod
```
DAreRodz: The three `e2e` test files are identical except for this line. Is it possible to use a single file and just change the command that will be executed? Maybe using 

```yaml
strategy:
  matrix:
    test-command: [prod, prod:module, prod:es5]
```
and then, in `steps`

```yaml
- name: Run e2e tests (${{ matrix.test-command }})
  uses: cypress-io/github-action@v1
  with:
    start: npm run e2e:${{ matrix.test-command }}
```

like in `unit-tests`, or something like that?

I know nothing about GitHub Actions, BTW. 🙈

DAreRodz (approved):
Great job.⭐
I have a question about `e2e` files, tho. Let me know what you think.

.github/workflows/e2e-tests-all.yml
luisherranz: Maybe it can be done with `ci_node_total` and `ci_node_index` like [this guy does for cypress](https://gist.github.com/ArturT/3535fc24bf2fac577462b9162c12c1d6), but it is not documented anywhere. I'll give it a try.

segunadebayo (approved):
I think this looks good. Great job, Luis 💫.

.github/workflows/e2e-tests-all.yml
luisherranz: It worked using `ci_node_total` and `ci_node_index` https://github.com/frontity/frontity/pull/244/commits/f4b10d0d300fc51e0c5e436339cff2b05036bf5f :)

Thanks for the suggestion @DAreRodz!

PR #250, opened 2019-12-17
TwentyTwenty transitions

luisherranz:

We are going to add transitions between pages to the TwentyTwenty theme.

#### My PR is a:

- 🔝 Improvement

- 🚧 Work in progress


luisherranz:
I have removed the `package-lock` of `twentytwenty-theme` because in our monorepo only the root can have a package-lock. Maybe we could add a GH action to check for that in the future as it's not intuitive :)

The problem with the `archive` failing to get the `data.items` is not because of prefetching or code-splitting, but because of the way it gets the data right now:

```jsx
{(data.isFetching && <Loading />) ||
  (isSearch && <SearchResults />) ||
   (data.isArchive && <Archive />) ||
   (data.isPostType && <Post />) ||
   (data.is404 && <Page404 />)}
```

```jsx
const Archive = ({ state, showExcerpt, showMedia }) => {
  // Get the data of the current list.
```

When we add transitions, the components cannot get the data anymore by themselves, because they do so against the current URL (`state.router.link`). They need to get their `data` from the parent.

If you take a look at the [example with transitions on mars-theme](https://codesandbox.io/s/frontity-page-animations-1tchc), you can see that data access is managed inside the transitions map of `react-spring` and passed down to the components:

```jsx
{transitions.map(({ item, props, key }) => {
        const data = state.source.get(item);
        return (
          <animated.div key={key} style={props}>
            <Absolute>
              <Body>
                {data.isFetching && <Loading />}
                {data.isArchive && <List data={data} />}
                {data.isPostType && <Post data={data} />}
                {data.is404 && <Page404 />}
                {data.content}
              </Body>
            </Absolute>
          </animated.div>
        );
      })}
```

This ensures that each component maintains the same data when they are "disappearing".

Finally, components use the `data` received by props instead of getting their own:
```jsx
const Archive = ({ state, showExcerpt, showMedia, data }) => {
   // Use data...
```

I don't know if that is possible `react-pose`, but it is with `react-spring` as shown in the codesandbox. Besides, it looks like [`react-pose` size](https://bundlephobia.com/result?p=react-pose@4.0.10) is 62kB and [`react-spring` size](https://bundlephobia.com/result?p=react-spring@8.0.27) is 26kB so I would try to make it work with `react-spring` first.

luisherranz:
The LightHouse test is falling because we set the maximum script size to 165kB, which is its current size. Once you add the animation library, you can update the `budget.json` file accordingly:
https://github.com/frontity/frontity/blob/twenty-twenty-transitions/.github/workflows/budget.json#L7

We thought it'd be a good idea to adjust manually the size to make sure it doesn't get too big without us noticing. Similar to the failing snapshots you have to manually update when something changes :)

luisherranz:
By the way, you are right, `autoPrefetch: "all"` is downloading too much data.

We have switched to `"hover"` in the [LightHouse improvements pull request](https://github.com/frontity/frontity/pull/249) as well.

michalczaplinski:
You're right. It looks substantially better with the changes you suggested! Thanks!
I will change how the <Archive /> component animates so that the white background does not fade in and out and only the text does so and I think it's done.

michalczaplinski:
@luisherranz I think the transition looks quite nice - I guess the slight jank should disappear in the faster production build, but waiting for the ZEIT keys to confirm.

Otherwise, LGTM 🙂

luisherranz:
Maybe we can try introducing `will-change` to improve the performance, but I'm not 100% sure about it.

Other than that, the search is broken but before fixing it here, I am going to add `data-isSearch` in `@frontity/wp-source` in another PR. So don't worry about it, I'll take care of it.

michalczaplinski:
I've attempted to fiddle with the will-change property but that makes no
discernible difference so I'm gonna just leave it as is for now.

PR #252, opened 2019-12-18
WIP: Pagination and Infinite Scrolling

michalczaplinski:
#### My PR is a:

- 🔝 Improvement

- 🚧 Work in progress

### DIFF
--- a/packages/twentytwenty-theme/package.json
+++ b/packages/twentytwenty-theme/package.json
@@ -21,6 +21,7 @@
     "@frontity/hooks": "^1.1.14",
     "@frontity/html2react": "^1.1.14",
     "frontity": "^1.4.2",
-    "react-spinners": "^0.5.4"
+    "react-spinners": "^0.5.4",
+    "react-intersection-observer": "^8.25.2"
   }
 }
--- a/packages/twentytwenty-theme/src/components/archive/archive.js
+++ b/packages/twentytwenty-theme/src/components/archive/archive.js
@@ -67,7 +67,7 @@ const Archive = ({
       })}
       {state.theme.infiniteScrolling ? (
         <InfiniteScroller
-          previousPageVisible={archiveIsVisible}
+          currentPageVisible={archiveIsVisible}
           currentLink={link}
         />
       ) : (
--- a/packages/twentytwenty-theme/src/components/archive/archive.js
+++ b/packages/twentytwenty-theme/src/components/archive/archive.js
@@ -2,13 +2,14 @@ import { connect } from "frontity";
 import React, { Fragment, useEffect } from "react";
 import Article from "../post/post-item";
 import ArchiveHeader from "./archive-header";
-import Pagination from "./archive-pagination";
+// import Pagination from "./archive-pagination";
 import PostSeparator from "../post/post-separator";
 import Post from "../post";
+import InfiniteScroller from "./infinite-scroll-pagination";
 
-const Archive = ({ state, showExcerpt, showMedia }) => {
+const Archive = ({ state, showExcerpt, showMedia, data }) => {
   // Get the data of the current list.
-  const data = state.source.get(state.router.link);
+  data = data || state.source.get(state.router.link);
   const { primary } = state.theme.colors;
 
   // Whether the show the excerpt instead of the full content
@@ -52,7 +53,8 @@ const Archive = ({ state, showExcerpt, showMedia }) => {
           </Fragment>
         );
       })}
-      <Pagination />
+      {/* <Pagination /> */}
+      <InfiniteScroller />
     </>
   );
 };
--- a/packages/twentytwenty-theme/src/components/archive/archive.js
+++ b/packages/twentytwenty-theme/src/components/archive/archive.js
@@ -7,9 +7,9 @@ import PostSeparator from "../post/post-separator";
 import Post from "../post";
 import InfiniteScroller from "./infinite-scroll-pagination";
 
-const Archive = ({ state, showExcerpt, showMedia, data }) => {
+const Archive = ({ state, showExcerpt, showMedia, data, link }) => {
   // Get the data of the current list.
-  data = data || state.source.get(state.router.link);
+  data = data || state.source.get(link || state.router.link);
   const { primary } = state.theme.colors;
 
   // Whether the show the excerpt instead of the full content
--- a/packages/twentytwenty-theme/src/components/archive/archive.js
+++ b/packages/twentytwenty-theme/src/components/archive/archive.js
@@ -2,14 +2,21 @@ import { connect } from "frontity";
 import React, { Fragment, useEffect } from "react";
 import Article from "../post/post-item";
 import ArchiveHeader from "./archive-header";
-// import Pagination from "./archive-pagination";
+import Pagination from "./archive-pagination";
 import PostSeparator from "../post/post-separator";
 import Post from "../post";
 import InfiniteScroller from "./infinite-scroll-pagination";
+import { useInView } from "react-intersection-observer";
 
-const Archive = ({ state, showExcerpt, showMedia, data, link }) => {
+const Archive = ({
+  state,
+  showExcerpt,
+  showMedia,
+  data,
+  link = state.router.link
+}) => {
   // Get the data of the current list.
-  data = data || state.source.get(link || state.router.link);
+  data = data || state.source.get(link);
   const { primary } = state.theme.colors;
 
   // Whether the show the excerpt instead of the full content
@@ -20,8 +27,13 @@ const Archive = ({ state, showExcerpt, showMedia, data, link }) => {
     Post.preload();
   }, []);
 
+  // Keep track of whether the Archive is visible or not.
+  // We will pass this as prop to the <InfiniteScroll /> to keep the URL
+  // in sync with whichever page number is currently visible
+  const [archiveRef, archiveIsVisible] = useInView();
+
   return (
-    <>
+    <div ref={archiveRef}>
       {/* If the list is a taxonomy, we render a title. */}
       {data.isTaxonomy && (
         <ArchiveHeader labelColor={primary} label={data.taxonomy}>
@@ -53,9 +65,15 @@ const Archive = ({ state, showExcerpt, showMedia, data, link }) => {
           </Fragment>
         );
       })}
-      {/* <Pagination /> */}
-      <InfiniteScroller />
-    </>
+      {state.theme.infiniteScrolling ? (
+        <InfiniteScroller
+          previousPageVisible={archiveIsVisible}
+          currentLink={link}
+        />
+      ) : (
+        <Pagination />
+      )}
+    </div>
   );
 };
 
--- a/packages/twentytwenty-theme/src/components/archive/infinite-scroll-pagination.js
+++ b/packages/twentytwenty-theme/src/components/archive/infinite-scroll-pagination.js
@@ -1,15 +1,21 @@
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import { connect, styled } from "frontity";
-import useInView from "@frontity/hooks/use-in-view";
+import { useInView } from "react-intersection-observer";
 
 import PostSeparator from "../post/post-separator";
 import Loading from "../loading";
 import Archive from "./archive";
 
-const InfiniteScroller = ({ state, actions, libraries }) => {
+const InfiniteScroller = ({
+  state,
+  actions,
+  libraries,
+  previousPageVisible,
+  currentLink
+}) => {
   // Get the total posts to be displayed based for the current link
-  const { totalPages } = state.source.get(state.router.link);
-  const { path, page, query } = libraries.source.parse(state.router.link);
+  const { totalPages } = state.source.get(currentLink);
+  const { path, page, query } = libraries.source.parse(currentLink);
 
   // Check if we can go to next page within the pagination
   const isThereNextPage = page < totalPages;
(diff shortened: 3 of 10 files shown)


michalczaplinski:
If I understand correctly that the fixes for pagination should include both:
- The pagination on the landing page (the list of all posts) - this is what I currently updated.
- The pagination on the page for an individual post (where you are able to move just 1 post back and forth). This one I'm not sure how to add with frontity or if it's technically possible

@luisherranz

luisherranz:
I'll just do the first one (list of posts) for now.

By the way, we usually mark the PRs that aren't meant to be merged yet with either `[WIP]` at the beginning of the title or starting them as "drafts" here in GitHub. Preferably the second :)

michalczaplinski:
right, gotcha. I'm working on the infinite scroll in that case

luisherranz:
@christianoliveira one SEO question.

When we add infinite scroll:
- Would it be great to maintain the normal pagination so Google can still see links to different pages?

And if the answer is yes:
- What should we do with those links once the user starts scrolling down and the next page appears?
- Can we remove the links?
- Is it better if we keep them in the footer?
- Is it relevant at all?

Maybe the second part is more about UX but I'm sure you know better.

michalczaplinski:
@luisherranz 
This is still WIP - I don't think that I can change the PR type to "draft" now, so I ll just keep it as it is until it's ready to merge.

luisherranz:
It's looking great so far! I think you have to pass down `data` or `link` instead of using `state.router.link` inside `<InfiniteScroller>` but you probably already know that :)

Other than that, the approach looks simple and easy to understand.

michalczaplinski:
I ran into some issues today when working with the useInView hook and trying to get the infinite scrolling to work...

It seems that as you scroll, react updates the refs to the `<ScrollObserver />` component and it causes only the last `<InifiniteScroll />` on the page to work. 
I makes it more clear when you follow the console logs [here](https://github.com/frontity/frontity/pull/252/files#diff-e0d022fe7a6b7df8c172ff619da7d113R35). You can see that at any given time there is only 1 intersection observer callback for each <ScrollObserver /> , even though if you scroll past the end, a second should be added.

If this is confusing to explain, I think I can make this clearer on a call 🙂 

PS. The react devtools extension causes frontity to crash, which might be because of how I recursively nest the `<Archive />` component, so disable it if you check out the code. I found this out almost accidentally and it made me waste some time trying to find the issue with my own code :sweat_smile:

luisherranz:
I don't know, could be the problem because you are nesting the observers inside one another?

Here `nextPageRef` contains the two new refs that are inside the new `Archive` component.

```jsx
      <ArchiveContainer>
        {data && data.isFetching && <Loading />}
        <ArchiveContainer ref={nextPageRef}>
          {data && data.isArchive && (
            <Archive data={data} link={nextPageLink} />
          )}
        </ArchiveContainer>
      </ArchiveContainer>
```

Maybe it'd be cleaner to position the refs outside (or surrounding) `Archive` instead of inside.

luisherranz:
> PS. The react devtools extension causes frontity to crash, which might be because of how I recursively nest the <Archive /> component, so disable it if you check out the code. I found this out almost accidentally and it made me waste some time trying to find the issue with my own code 😅

Yes, React DevTools freeze from time to time. I don't know if it's a Frontity issue or not.

packages/twentytwenty-theme/src/components/archive/archive.js
```diff
@@ -52,8 +65,15 @@ const Archive = ({ state, showExcerpt, showMedia }) => {
           </Fragment>
         );
       })}
-      <Pagination />
-    </>
+      {state.theme.infiniteScrolling ? (
```
michalczaplinski, line 68: @luisherranz I think I'm using the state / settings correctly here? I tested it and it seems okay

packages/twentytwenty-theme/package.json
```diff
@@ -21,6 +21,7 @@
     "@frontity/hooks": "^1.1.14",
     "@frontity/html2react": "^1.1.14",
     "frontity": "^1.4.2",
-    "react-spinners": "^0.5.4"
+    "react-spinners": "^0.5.4",
+    "react-intersection-observer": "^8.25.2"
```
michalczaplinski, line 25: @luisherranz I will update our code to use https://github.com/thebuilder/react-intersection-observer/ in other places in a separate PR

packages/twentytwenty-theme/src/components/archive/infinite-scroll-pagination.js
```diff
@@ -0,0 +1,91 @@
...
+        setNextPageData(state.source.get(nextPageLink));
+      });
+    }
+  }, [scrollObserverVisible, isThereNextPage]);
+
+  // Keep the URL in sync with the pages, as the user keeps scrolling
+  useEffect(() => {
+    // The user has scrolled to the next page
+    if (nextPageVisible && !previousPageVisible) {
+      history.replaceState(null, null, nextPageLink);
```
michalczaplinski: @luisherranz Will update our `tiny-router` in a separate PR. For now this works.

michalczaplinski:
This adds some slightly updated styling for pagination and most importantly adds infinite scrolling. 

Personally, I think it works pretty nice as it is but waiting for feedback 🙂

packages/twentytwenty-theme/src/components/archive/archive.js
luisherranz: Yes, sure. That's the way settings work :)

packages/twentytwenty-theme/package.json
luisherranz: Great 👍

luisherranz (changes requested):
Looks simply and neat. Amazing work :)

I've done some minor comments but otherwise I think this is good to go. I'll wait to merge it until Chris shares with us if we should still add pagination for Google or not tho.

packages/twentytwenty-theme/src/components/archive/infinite-scroll-pagination.js
```diff
@@ -0,0 +1,91 @@
...
+  // Get the link for the next page
+  const nextPageLink = libraries.source.stringify({
+    path,
+    page: page + 1,
+    query
+  });
+
+  // Create a ref to ScrollObserver and observe it's visiblility.
+  // ScrollObserver is a "hit area" - as soon as it's visible in the viewport
+  const [scrollObserverRef, scrollObserverVisible] = useInView();
```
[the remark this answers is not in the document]
michalczaplinski: I thought about this, but this means that we would have 2 refs:
a) `scrollObserverRef` (with 500px rootMargin) and  
b) `nextPageRef` (the same as we have now)
And then we pass BOTH of them to `<ArchiveContainer />` 
Since you cannot pass 2 refs to the same component, we'd have to use something like: https://github.com/seznam/compose-react-refs, which I think is an overkill for the example.

michalczaplinski:
@luisherranz Uff, I've realized that the "issue" that we've been having with the router is really because of the fact that infinite scrolling has a slightly different conceptual model than the routing + data fetching in frontity. Let me try to explain.

In the infinite scrolling example, we load the components more or less like this:

```
1. <Archive>                      link = state.router.link  =  "/"  
2.   <InfiniteScroll>             load the nextPageLink. nextPageLink = "/page/2"
3.      <Archive link="/page/2">
4.        <InfiniteScroll>        load the nextPageLink again, nextPageLink = "/page/3"
5.          <Archive link="/page/3">
```

So, this works if we just use the `window.history` api directly. 

But when I call `actions.router.set()` the `state.router.link` gets updated and this causes component `1.` to rerender with the new data, basically destroying our nice recursive structure. 

Of course, I think that I could "save" the initial value of the `link` in component `1.` but I think this would be getting away from an "easy" example. I would just leave the example as it is to avoid extra work just for the sake of this example. Any thoughts?

christianoliveira:
Hi @luisherranz and @michalczaplinski !

> * Would it be great to maintain the normal pagination so Google can still see links to different pages?

Yes, definitely. The best way of thinking about this in order to make it "SEO-friendly" is to design it to work without JS. That means:

- Google needs to see plain html links (`<a href="...">`) to discover each of the URLs of the pagination
- Each URL has to work when accessed directly, and show only the content that it is supposed to show (only the posts that correspond to that specific page)

From an SEO point of view I think it is safe for now to hide the links if JS is available, although this may change in the future, depending on how Googlebot deals with differences between rendered and non-rendered content. If you decide to hide them, I would keep them on the html anyway so Googlebot can still "see" them after rendering the page.

From a UX point of view it may make sense to leave the links visible, visually marking the current page at all times as a way of informing the user "where s/he is" inside the pagination. You may consider configuring those links as a "floating element".

You should also update the URL on the browser as the user scrolls the page, both up and down.

You can check [Google's official recommendations](https://webmasters.googleblog.com/2014/02/infinite-scroll-search-friendly.html) here (although written in 2014, they are still relevant) and see a [SEO friendly example here](http://scrollsample.appspot.com/items), built by John Mueller from Google

Another thing to consider is what to do if the site has a footer: it can be very frustrating for users to try to get to the footer and never achieve that while there are more pages available loading automatically as they scroll down. One common solution is to stop automatically loading pages after 2 or 3 pages and include after that a "Load more" link/button that will allow the user to reach the footer or continue loading pages ([Google Images does this](https://www.google.com/search?q=cheesecake&tbm=isch&source=lnms)). Other solution is to make the entire footer a "floating element", but this is tricky as it depends on the size of the footer. 

In any case, as a Frontity user I think that infinite scroll should be an optional feature I can activate and even configure its behaviour for my specific site.

Let me know if you have any doubts!

luisherranz:
You're right. We would have to do something like this, which looks quite horrible:

```jsx
<Main id="main">
  {((data.isArchive || (prevData.isArchive && data.path === prevData.path)) && (
    <Archive />
  )) ||
    (data.isFetching && <Loading />) ||
    (isSearch && <SearchResults />) ||
    (data.isPostType && <Post />) ||
    (data.is404 && <Page404 />)}
</Main>
```

I'll think about it.

luisherranz:
Maybe we introduce the `Switch` component to clean this a bit:

```jsx
<Switch>
  <Archive when={data.isArchive} when={prevData.isArchive && data.path === prevData.path} />
  <PostType when={data.isPostType} />
  <Search when={data.isSearch} />
  <Loading when={data.isLoading} />
  <Page404 />
</Switch>;
```

But I guess if there's a better solution than checking the previous `data` object.

luisherranz:
Thanks for the insights @christianoliveira!

> If you decide to hide them, I would keep them on the html anyway so Googlebot can still "see" them after rendering the page.

So the best idea is to keep the pagination element in the HTML even after the JS has loaded, and just hide it with CSS, is that correct?

> You may consider configuring those links as a "floating element".

That's a great idea! Here we are replicating an existing theme, so I think we will just hide the pagination. But we can do the floating pagination in other themes :)

> You should also update the URL on the browser as the user scrolls the page, both up and down.

We are working on that 👍 

> One common solution is to stop automatically loading pages after 2 or 3 pages and include after that a "Load more" link/button that will allow the user to reach the footer

Ok, we'll add it here. Actually, it's how it worked in Frontity PRO but I forgot about that!

> In any case, as a Frontity user I think that infinite scroll should be an optional feature I can activate and even configure its behaviour for my specific site.

Yes, we have added a feature of the theme. I can be deactivated.

christianoliveira:
> So the best idea is to keep the pagination element in the HTML even after the JS has loaded, and just hide it with CSS, is that correct?
> 
> >

Yes, that's correct!

luisherranz:
There's yet another problem we haven't took into account: we need to make it work with the back and forward buttons of the browser.

But I think I've found a way to solve all the problems while keeping your "recursive pattern". 

We need to add a `state.router.history` array with information about the previous visited URLs, including if they were `push` or `replace`.

```js
state.router.history = [
  {
    link: "/some-post",
    method: "push"
  },
  {
    link: "/",
    method: "push"
  },
  {
    link: "/page/2",
    method: "replace"
  },
  {
    link: "/page/3",
    method: "replace"
  },
]
```

We are going to need a little `<Archive>` wrapper that extracts the last `push` link from `state.router.history` and passes it down. After that, I think the `<ArchivePage>` component should be able to take care of everything, including adding the next pages.

Also, we cannot trigger an `action.router.set` until the `data.isReady` of that `link` is true, to avoid destroying the parent `<Archive>` component.

Steps:

1. The `Archive` wrapper extracts the last `push` link from `state.router.history`:

```jsx
const Archive = ({ state }) => {
  const link = state.router.history.reverse().find(({ method }) => method === "push");
  return <ArchivePage link={link} />;
}
```

2. Inside <ArchivePage /> a `useInView` with a `rootMargin` of `600px` triggers the fetch of the next page.

3. `<ArchivePage>` adds a new `<ArchivePage>` with `link` pointing to the next page when `data.isReady` and either that link it's in the `replace` links of `state.router.history` or we have triggered the fetch `useInView`.

4. To avoid changing the route to `data.isLoading`, each page needs to take care of its own `actions.router.set`. This is an easy way to configure the `useInView` hooks to do so:
https://codesandbox.io/s/wandering-paper-z0hbx
_(the intersection observer doesn't work in codesandbox inline browser, you have to open it in another tab: https://z0hbx.csb.app/)_

5. Always show the pagination after  `data.page > 3`.

```jsx
const ArchivePage = ({ state, actions, link }) => {
  const data = state.source.get(link);

  const [fetchTrigger, fetchTriggerRef] = useInView({ rootMargin: "600px", onlyOnce: true });
  const [routerTrigger, routerTriggerRef] = useInView({ rootMargin: "-80% 0% -19.999999999999% 0%" });

  // Get all the link that have been visited via scrolling (replace) since the last click (push).
  const replaceLinks = takeLastWhile(({ method }) => method === "replace", state.router.history);
  const nextData = state.source.get(data.next);
  // Show the next page only when the data is ready, page is less than 3 AND one of these two things happen:
  // - The user has scrolled pass the fetchTrigger point.
  // - That page has already been visited (the user hit the back or forward button of the browser).
  const showNextPage =
    data.page <= 3 &&
    nextData.isReady &&
    (replaceLinks.includes(data.next) || fetchTrigger);

  React.useEffect(() => {
    if (data.next) actions.source.fetch(data.next);
  }, [fetchTrigger]);

  React.useEffect(() => {
    if (routerTrigger && nextData.isReady) actions.router.set(link);
  }, [routerTrigger, nextData.isReady]);

  return (
    <>
      <div ref={routerTriggerRef}>
        {data.items.map( /* ...  */)}
        <PostSeparator ref={fetchTriggerRef} />
      </div>
      {showNextPage
         ? <ArchivePage link={data.next} />
         : {nextData.isLoading ? <Loading /> : <Pagination />}
      }
    </>
  );
};
```

I'll try this implementation as soon as I can 👍

michalczaplinski:
@luisherranz What's the reason for closing this? I thought that this work is still supposed to be continued because of #292 and #293

luisherranz:
Oh, it was closed because I deleted the `beta` branch. I'm sorry, I'll fix it tomorrow.

michalczaplinski:
@luisherranz are you able to reopen this PR? Otherwise, I can open the new one, but would would be better to keep the conversation in one place :)

PR #253, opened 2019-12-18
Add search links and pagination to wp source

### LINKED ISSUE #312: Access pagination and search in the data objects
**Feature:** [Search links and pagination in data objects](https://community.frontity.org/t/search-links-and-pagination-in-data-objects-of-wp-source/1230)
**User Story**: Access pagination and search in the data objects

I started a PR (https://github.com/frontity/frontity/pull/253) but you can take over and continue the work.

Let's improve searches and pagination by adding:
- For all entities:
   - `data.link`
   - `data.query`
- For searches:
   - `data.isSearch`
   - `data.searchQuery`
- For archives:
   - `data.page`
   - `data.path`
   - `data.next`
   - `data.previous`

Tasks:
- Add new types to Data
- TypeScript guards tests 
- TypeScript object tests
- Add new logic to handlers
- Update docs

luisherranz:

Let's improve searches and pagination by adding:
- For all entities:
   - [x] `data.link`
   - [x] `data.query`
- For searches:
   - [x] `data.isSearch`
   - [x] `data.searchQuery`
- For archives:
   - [x] `data.page`
   - [x] `data.route`
   - [x] `data.next`
   - [x] `data.previous`

Tasks:
- [x] Add new types to Data
- [x] TypeScript guards tests 
- [x] TypeScript object tests
- [x] Add new logic to handlers
- [x] Update docs
- [x] Change `path` for `route` in `libraries.source.stringify`
  - [x] While keeping `path` for backwards compatibility
- [x] Change `path` for `route` in `libraries.source.parse`
  - [x] While keeping `path` for backwards compatibility
- Refactor themes pagination (use `data` instead of `parse` and `route` instead of `path`):
  - [x] mars-theme
  - [x] twentytwenty
- Refactor themes to use `isSearch` and `searchQuery`:
  - [x] ~mars-theme don't have search~
  - [x] twentytwenty
- [x] Change `route` to `link` in handlers callback
  - [x] While keeping `route` for backwards compatibilitiy
  - [x] Also, rename it internally inside `fetch`
- [x] Duplicate tests for both `route` and `path` in `parse` and `stringify`
- [x] Duplicate tests for both `route` and `link` in handler callbacks

`link`, `query`, `route` and `page` can be added even before the fetch and they should be also returned by `source.get` when `source.data` is undefined.

#### Documentation

https://app.gitbook.com/@frontity-tech/s/frontity/~/diff/drafts/-M0yjLG_zUV-XTqIlBCK/learning-frontity/state/@drafts

#### My PR is a:

- 🔝 Improvement

- 🚧 Work in progress


luisherranz:
We should check if this solves this bug:
- go to 404 page
- switch to offline
- search for something

packages/wp-source/src/libraries/handlers/author.ts
```diff
@@ -38,27 +38,34 @@ const authorHandler: Handler = async ({
...
+  const total = getTotal(response, items.length);
+  const totalPages = getTotalPages(response, 0);
 
-  // 5. add data to source
+  // 5. Add data to source..
   const currentPageData = state.source.data[route];
   const firstPageData = state.source.data[path];
 
   Object.assign(currentPageData, {
+    link: route,
```
michalczaplinski: @luisherranz The `link` property is redundant, right? But we still want to have it on the data object for each route...?

luisherranz: I've found that when you pass down `data` like this:

```js
<Post data={data} />
```

it feels a bit weird to know everything about the post (the `type`, the `id`, the `isPost`...) but to get the `link` you have to go somewhere else to get it:

```js
const id = data.id;
const type = data.type;
const isPost = data.isPost;
const link = state.source[data.type][data.id].link // <-- feels overcomplicated.
```

or you have to pass it down along with `data`:

```js
<Post data={data} link={link} />
```

so I thought it made sense to store it in `data.link` along with the rest of the properties.

----

By the way maybe this type of discussion, not actually related to the implementation but to the design, could happen in the [feature topic](https://community.frontity.org/t/search-links-and-pagination-in-data-objects-of-wp-source/1230). What do you think?

michalczaplinski: You're right, this is more of a discussion for the forum. I ll copy it there :) 

Just for reference, I agree with your reasoning here.

packages/source/types/data.ts
```diff
@@ -49,6 +58,16 @@ export type BaseData = {
...
+
+  // search attributes
+  isSearch?: false;
+  searchQuery?: string;
+
+  // pagination atributes (for Archive)
+  page?: number;
+  path?: string;
+  next?: string;
+  previous?: string;
```
michalczaplinski, line 70: @luisherranz this is what I'm not 100% sure about.
Should it be necessary to declare all the new properties on the `BaseData` ?

packages/wp-source/src/libraries/handlers/taxonomy.ts
```diff
@@ -69,7 +72,15 @@ const taxonomyHandler = ({
     isArchive: true,
     isTaxonomy: true,
     [`is${capitalize(firstPageData.taxonomy)}`]: true
-  });
+  } as TaxonomyData);
```
michalczaplinski: adding `as TaxonomyData` seems like the best (only?) way to ensure typesafety when doing `Object.assign()`

packages/source/types/data.ts
luisherranz: I don't think so because you won't access those props until you make sure that that is an archive (using `data.is..`), but not 100% sure.

packages/wp-source/src/libraries/handlers/author.ts
```diff
@@ -13,6 +13,15 @@ const authorHandler: Handler = async ({
   const { path, page, query } = parse(route);
   const { slug } = params;
 
+  // Add the attributes that should be present even if fetch fails or we throw a ServerError below
```
luisherranz: Good one 👍 

Yes, after this PR I expect people to use
```js
const { path } = state.source.get(someLink);
```
instead of
```js
const { path } = libraries.source.parse(someLink);
```

luisherranz:
I've been thinking about the problem of not having type safety inside handlers.

I've done this test:

```js
  const currentPageData = state.source.data[route];
  
  const newPageData: DateData = {
    isArchive: true,
    isDate: true,
    year,
    month,
    day,
    items,
    isReady: currentPageData.isReady,
    isFetching: currentPageData.isFetching,
    total,
    totalPages
  };

  Object.assign(currentPageData, newPageData);
```

And it works. We can cast `DateDate` by declaring `newPageData` as a new variable.

This doesn't work yet though:
```js
  state.source.data[route] = {
    ...currentPageData,
    ...newPageData
  };
```

Because of the incompatible types of `isArchive` and `isDate` which is `false` in `BaseData` and `true` in `ArchiveData/DateData`. 

I never remember why we need those `false` types in `BaseData` although I know @DAreRodz has explained it to me several times 😓I guess we'll have to ask him again once he's back.

In the meantime, you can use `const newPageData: DateDate = { ... }` and `Object.assign(currentPageData, newPageData);`.

packages/wp-source/src/libraries/handlers/taxonomy.ts
michalczaplinski: not relevant anymore, we decided to forgo the typesafety for the time being in this case until we do a proper refactor of wp-source :)

michalczaplinski:
@luisherranz 

nice, you're right, it works 🙂 I've updated the PR accordingly.

I still have a couple outstanding questions (check comments in code) and I still want to add some new tests.

packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/author.tests.ts.snap
```diff
@@ -195,8 +197,10 @@ Object {
       "isFetching": false,
       "isReady": true,
       "items": Array [],
-      "total": NaN,
-      "totalPages": NaN,
+      "next": undefined,
```
michalczaplinski: so, prev and next can be `undefined` if there is no prev or next page.

packages/wp-source/src/libraries/handlers/date.ts
```diff
@@ -43,20 +44,42 @@ export const dateHandler: Handler = async ({
...
-    isDate: true
-  });
+    isDate: true,
+    isFetching: currentPageData.isFetching,
+    isReady: currentPageData.isReady,
+    prev: hasOlderPosts ? getPageLink(page - 1) : undefined,
+    next: hasNewerPosts ? getPageLink(page + 1) : undefined
+  };
+
+  Object.assign(currentPageData, newPageData);
```
michalczaplinski: not super happy that this is basically repeated 4 times in `taxonomy`, `data`, `author` and `archivePostType` handlers, but extracting it to a function is bad too because we depend on the state from:
```
 const { api, populate, parse, getTotal, getTotalPages } = libraries.source;
 const { path, page, query } = parse(route);
```
so we'd end up passing a lot of function arguments around

packages/wp-source/src/actions.ts
```diff
@@ -20,10 +20,20 @@ const actions: WpSource["actions"]["source"] = {
...
     const data = source.data[route];
 
+    // Add the attributes that should be present even if fetch fails or we throw a ServerError below
+    state.source.data[route] = {
+      ...state.source.data[route],
+      link: route,
+      path: routeParams.path,
+      query,
+      page
+    };
```
michalczaplinski: This is the reason that the tests are failing. I'm not 100% sure if we should adjust the test or if this is a problem.

One of the example test cases: 
```
 // Add category to the store
    await store.libraries.source.populate({
      state: store.state,
      response: mockResponse([cat1])
    });
    // Mock Api responses
    api.get = jest.fn().mockResolvedValueOnce(
      mockResponse(cat1PostsPage2, {
        "X-WP-Total": "5",
        "X-WP-TotalPages": "2"
      })
    );
    // Observe changes in isFetching and isReady properties
    const dataState = [];
    observe(() => {
      const { isFetching, isReady } = store.state.source.get(
        "/category/cat-1/page/2/"
      );
      dataState.push({ isFetching, isReady });
    });
    // Fetch entities
    await store.actions.source.fetch("/category/cat-1/page/2/");
    expect(api.get).toHaveBeenCalledTimes(1);
    expect(store.state.source).toMatchSnapshot();
    // Values history of isFetching and isReady
    expect(dataState).toEqual([
      { isFetching: false, isReady: false }, // first values are from a different object
      { isFetching: false, isReady: false }, // initial values from the data object
      { isFetching: true, isReady: false }, // fetch starts
      { isFetching: false, isReady: true } // fetch ends
    ]);
```

Because we don't set the `isFetching` and `isReady` here, the are undefined in the test, after the initial first value are set.

Not sure if we need to set them as well, or if that's fine?

packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/author.tests.ts.snap
luisherranz: We don't have a particular guideline for this although so far Frontity is more aimed to "missing properties" instead of "undefined properties" o "null properties".

michalczaplinski:
@luisherranz Could you review this?

Notes: 
1) I want to add a couple of tests for `previous` and `next` but I'm not sure about how the previous and next links are supposed to be generated and where they come from. (maybe better to explain in a call)
2) We have `path`, `link` and now we want to add the new param `route` which is supposed to be the same as `path`, correct? (We talked about this I know, I forgot to write it down, sorry)

docs: https://app.gitbook.com/@frontity-tech/s/frontity/~/diff/drafts/-M0yjLG_zUV-XTqIlBCK/learning-frontity/state/@drafts

packages/wp-source/src/libraries/handlers/author.ts
```diff
@@ -76,8 +76,10 @@ const authorHandler: Handler = async ({
     isAuthor: true,
     isReady: currentPageData.isReady,
     isFetching: currentPageData.isFetching,
-    prev: hasOlderPosts ? getPageLink(page - 1) : undefined,
-    next: hasNewerPosts ? getPageLink(page + 1) : undefined
+
+    // Add those keys if hasOlderPosts / hasNewerPosts === true
+    ...(hasOlderPosts && { prev: getPageLink(page - 1) }),
```
luisherranz: n

packages/wp-source/src/libraries/handlers/author.ts
```diff
@@ -76,8 +76,10 @@ const authorHandler: Handler = async ({
...
+    ...(hasNewerPosts && { next: getPageLink(page + 1) })
```
luisherranz: Nice trick!

luisherranz (commented):
> We have path, link and now we want to add the new param route which is supposed to be the same as path, correct? (We talked about this I know, I forgot to write it down, sorry)

I have added the remaining tasks to the PR description. Let me know if that is enough 👍 

> I'm not sure about how the previous and next links are supposed to be generated and where they come from. (maybe better to explain in a call)

I think what you did is fine, although you're right that feels quite repetitive. Maybe we can move that logic to `fetch` and move [this logic](https://github.com/frontity/frontity/blob/add-search-links-and-pagination-to-wp-source/packages/wp-source/src/libraries/handlers/author.ts#L59-L64) (and [this](https://github.com/frontity/frontity/blob/add-search-links-and-pagination-to-wp-source/packages/wp-source/src/libraries/handlers/author.ts#L81-L82)) to [this point](https://github.com/frontity/frontity/blob/add-search-links-and-pagination-to-wp-source/packages/wp-source/src/actions.ts#L64). If you think it's a good idea, please add a new task so we can keep track of everything :)

michalczaplinski:
@luisherranz 

> - [ ] Change `route` to `link` in handlers callback
>   - [ ] While keeping `route` for backwards compatibilitiy
>   - [ ] Also, rename it internally inside `fetch`

Sorry, I need to ask again, but are you sure this is correct? I thought that we need to change the name of `path` to `route` like:

**previous**
```js
// wp-source/src/actions.ts
source.data[route] = {
        link: route,
        path: routeParams.path,
        query,
        page,
        isFetching: true,
        isReady: false
      };
```

**next**
```js
source.data[route] = {
        link: route,
        path: routeParams.path,
        route: routeParams.path
        query,
        page,
        isFetching: true,
        isReady: false
      };
```

luisherranz:
Yeah, we are passing `route` to the handlers callbacks, like indicated here: https://docs.frontity.org/api-reference-1/wordpress-source#handlers

We should rename that to `link` (but keep passing `route` as well so we don't break the current handlers). This is the handler call:
https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/actions.ts#L53

So we have to change this:
```js
await handler.func({
        route,
        params: handler.params,
        state,
        libraries,
        force
      });
```
for this:
```js
await handler.func({
        route,
        link: route,
        params: handler.params,
        state,
        libraries,
        force
      });
```
(or better rename `route` internally to `link`)

And then change the handlers like this: https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/libraries/handlers/author.ts#L5

From this:
```js
const authorHandler: Handler = async ({
  route,
  params,
  state,
  libraries,
  force
}) => {
```

to this:
```js
const authorHandler: Handler = async ({
  link,
  params,
  state,
  libraries,
  force
}) => {
```

We don't know how much it'll take to release the new version so we better fix the naming right now so people can get used to it :)

michalczaplinski:
Still missing for now:

- I haven't checked if the docs need to be updated as well e.g for `stringify` and `parse`
- Add the pagination for mars-theme

The rest should be good to go, but I havent reviewed all the diffs yet.

packages/twentytwenty-theme/src/components/archive/archive-pagination.js
```diff
@@ -41,13 +41,10 @@ const paginate = (totalPages, currentPage) => {
 };
 
 const Pagination = ({ state, actions, libraries }) => {
-  const { totalPages } = state.source.get(state.router.link);
-  const { path, page, query } = libraries.source.parse(state.router.link);
+  const { path, query, totalPages, next, previous, page } = state.source.get(
```
luisherranz: Shouldn't this be `route` instead of `path`?

packages/twentytwenty-theme/src/components/search/search-modal.js
```diff
@@ -8,10 +8,9 @@ import useFocusTrap from "../hooks/use-trap-focus";
 import useFocusEffect from "../hooks/use-focus-effect";
 import Button from "../styles/button";
 
-const SearchModal = ({ state, actions, libraries }) => {
+const SearchModal = ({ state, actions }) => {
   // Improvement: Discuss with David about making a custom hook to grab query params
```
luisherranz: You can remove this comment.

packages/wp-source/src/actions.ts
```diff
@@ -12,27 +12,28 @@ import { ErrorData } from "@frontity/source/types/data";
...
     const force = options ? options.force : false;
 
     if (!data || force) {
       // Add the attributes that should be present even if fetch fails or we throw a ServerError below
-      source.data[route] = {
-        link: route,
-        path: routeParams.path,
+      source.data[link] = {
+        link,
+        path: linkParams.path,
```
luisherranz: I guess this would be `route` as well.

packages/wp-source/src/actions.ts
```diff
@@ -44,7 +45,7 @@ const actions: WpSource["actions"]["source"] = {
 
     // get and execute the corresponding handler based on path
     try {
-      let { path } = routeParams;
+      let { path } = linkParams;
```
luisherranz: Same here.

luisherranz (commented):
I see you have changed `next` and `previous` so:
- `next` of page 2 is page 1.
- `previous` of page 2 is page 3.

Let's investigate how is the norm in WordPress and follow that.

luisherranz:
Ok, this is what WordPress does:
- It adds two meta tags in the `<head>` to the HTML of https://test.frontity.io/page/2/ in the front-end:
```html
<link rel="prev" href="https://test.frontity.io/" />
<link rel="next" href="https://test.frontity.io/page/3/" />
```
- It adds a `link` header to the REST API response of https://test.frontity.io/wp-json/wp/v2/posts?page=2
```
link: 
  <https://test.frontity.io/wp-json/wp/v2/posts?page=1>; rel="prev",
  <https://test.frontity.io/wp-json/wp/v2/posts?page=3>; rel="next"
```

So it looks like in both cases:
- `next` of page 2 is page 3.
- `prev` of page 2 is page 1.

It also looks like the naming is `prev` instead of `previous` although that is not that important I think :)

luisherranz:
@michalczaplinski you're gonna kill me but I've added another task to duplicate the tests that use `path` in `parse` and `stringify` so we have tests for both `path` and `route`. The same for `route` and `link` in handlers.

I think it's important to have duplicate tests for both names to make sure we don't accidentally break one or the other. We can delete those tests in the future, once we release a major version and remove the old names.

packages/twentytwenty-theme/src/components/archive/archive-pagination.js
michalczaplinski: I thought that no? It's the `path` attribute of [`Data`](https://github.com/frontity/frontity/pull/253/files#diff-879b131cdb78f509b45efc097fac4ab7R68)

 If I understood it correctly, it was not supposed to change?

packages/wp-source/src/actions.ts
michalczaplinski: right, correct!

michalczaplinski: wait, no I think not!

packages/twentytwenty-theme/src/components/archive/archive-pagination.js
michalczaplinski: oh, nevermind, you did mean to change the `path` to `route` there as well. All clear.

michalczaplinski:
@luisherranz 
> @michalczaplinski you're gonna kill me but I've added another task to duplicate the tests that use path in parse and stringify so we have tests for both path and route. The same for route and link in handlers.

I dont think there is a test that I can duplicate for the handlers. All of our handlers are now using the `link` parameter instead of the `route` internally. I think I would have to write a custom handler for this that uses the `link` parameter instead and test that. Does that make sense?

> link, query, route and page can be added even before the fetch and they should be also returned by source.get when source.data is undefined.

I've added the `link`, `query`, `route` and `page` [here in the `get()`](https://github.com/frontity/frontity/pull/253/files?file-filters%5B%5D=.js&file-filters%5B%5D=.snap&file-filters%5B%5D=.ts#diff-f150c68432e61538c53b76bb7ab7ade4R5-R17). However, this way they are added for all entities. Earlier, you mentioned that `page` and `route` should only be added for the Archives: 

> - For archives:
>    - [x] `data.page`
>    - [x] `data.route`
>    - [x] `data.next`
>    - [x] `data.previous`

Which one is it? 🙂 

> So it looks like in both cases:
> - `next` of page 2 is page 3.
> - `prev` of page 2 is page 1.

okay, also fixed now 👍

packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/author.tests.ts.snap
```diff
@@ -469,6 +487,193 @@ Object {
...
+          "type": "post",
+        },
+        Object {
+          "id": 3,
+          "link": "/post-3/",
+          "type": "post",
+        },
+      ],
+      "link": "/author/author-1/?s=findAuthor",
+      "next": "/author/author-1/page/2/?s=findAuthor",
```
michalczaplinski, line 557: @luisherranz should `next` and `previous` include the query string?

packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/date.tests.ts.snap
```diff
@@ -626,7 +638,14 @@ Object {
           "type": "post",
         },
       ],
+      "link": "/2019/?some=param",
       "month": undefined,
```
michalczaplinski: @luisherranz This is not related to the current PR, but `month` and `day` are undefined if the route does not include them, for example the route `/2019/page/2/?some=param` in this test will return `undefined` month and day.

Should we include `day` and `month` conditionally in the final data?

Like we do with `previous` and `next`:

```js
{
  ...(day && { day }),
  ...(month && { month }),
}

luisherranz: Yes, good idea!

I'll do the same for `query.s`.

luisherranz (commented):
Looks perfect to me 👍 

Consider it approved (I can't because I opened it, but approve it yourself).

- I've added an improved changeset on your behalf.
- I've added type guard tests for searches.
- I've used your spread trick for search and day/month properties.

Feel free to merge this.

michalczaplinski (approved):

PR #256, opened 2019-12-23
Intersection observer hook update

michalczaplinski:

- [x]  Create a new  `@frontity/invariant` package
- [x]  Add a depreciation notice to our `@frontity/hooks/useInView` package. 
- [x]  Test and update the `<Image />` component to make sure that it works correctly with the new hook
- [x] Remove commented out old code that used the "old" hook

#### My PR is a:

- 🔝 Improvement

- ✅ Ready to be merged


michalczaplinski:
@luisherranz I think that I would need your help on how to `add` new packages with lerna. It seems that the other packages do not "see" the new `@frontity/errors` packages.

packages/core/src/config/babel/index.ts
```diff
@@ -74,7 +74,8 @@ export default ({ mode }: { mode: Mode }): BabelConfigs => {
         {
           include: ["CWD"]
         }
-      ]
+      ],
+      "babel-plugin-dev-expression"
```
luisherranz: Add a comment here about what's this for.

packages/error/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+// Strip out error messages for production
+// > Not providing an inline default argument for message as the result is smaller
+export default function invariant(condition: boolean, message?: string) {
+  if (condition) {
+    return;
+  }
+  // Condition not passed
+
+  if (isProduction) {
+    // In production we strip the message but still throw
```
luisherranz: I really like the idea of using `babel-plugin-dev-expression` but if we are using it, why do we need to strip the message here? https://www.npmjs.com/package/babel-plugin-dev-expression#invariant

packages/error/src/index.ts
```diff
@@ -0,0 +1,21 @@
+const isProduction: boolean = process.env.NODE_ENV === "production";
+const prefix = "\nVisit https://community.frontity.org for help! 🙂\n";
+
+// Throw an error if the condition fails
+// Strip out error messages for production
+// > Not providing an inline default argument for message as the result is smaller
+export default function invariant(condition: boolean, message?: string) {
```
luisherranz: As it looks like we are going to have two exports (`invariant` and `warning`), what about making them named exports?

That way we can also ensure that `babel-plugin-dev-expression` always works because it seems like it only works when it finds `invariant` and `warning` functions, and it doesn't work for other names like `error` or `warn`.

packages/error/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (condition) {
+    return;
+  }
+  // Condition not passed
+
+  if (isProduction) {
+    // In production we strip the message but still throw
+    throw new Error(prefix);
```
luisherranz: Do we want to throw the "visit community" message in production? (if so, I guess the babel plugin won't help).

packages/hooks/use-in-view.ts
```diff
@@ -55,6 +56,13 @@ const useInView: UseInView = ({ rootMargin, onlyOnce } = {}) => {
   const [isIntersecting, setIntersecting] = useState(false);
   const ref = useRef();
 
+  useEffect(() => {
+    error(
+      true,
+      "The @frontity/hooks package is deprecated.\nThe Frontity team recommends that you use https://github.com/thebuilder/react-intersection-observer instead."
+    );
```
luisherranz: This has to be a deprecation warning, not an error. Also, it's not the `@frontity/hooks` package, but only the `useInView` hook.

packages/error/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@frontity/error",
+  "version": "1.0.0",
```
luisherranz: We have to start with a `0.0.1` or `0.1.0` version so when we do a `changeset version` with a major change for this package it will release the `1.0.0` version. If we start in `1.0.0` it releases the `2.0.0`.

luisherranz (commented):
I really like the idea of using `babel-plugin-dev-expression`! It's much simpler that way.

> I think that I would need your help on how to add new packages with lerna.

Just go to the root and do `npm install` again to trigger the lerna bootstrap again.

I've added some comments to the code, but my main concern is... should this package be named `@frontity/error` or maybe there's a better name for it? `@frontity/invariant`? Is there a name that can cover both invariants and warnings?

packages/core/src/config/babel/index.ts
luisherranz: You also have to update the `@frontity/core` snapshots.

packages/error/src/index.ts
michalczaplinski: So, as far as I understand, the use case is something like in React: In dev you get a full error message, but in production you just get an error code which you can look up to see what the full error message is. Maybe it's a bit overengineered 😅. We don't actually need to use it to strip out whatever is in `__DEV__` I realized.

packages/error/src/index.ts
michalczaplinski: yes, that's what I was thinking as well.

packages/error/src/index.ts
michalczaplinski: No, I think not! I think we don't show the warning, and if it's an error then we just throw the error and only in __DEV__ we have a full error message.

packages/hooks/use-in-view.ts
michalczaplinski: So, I was thinking that we could follow the facebook strategy by the way, and show all warnings as errors (like react does), but I don't have a strong opinion on that.

luisherranz: Yes, that was my thought too but this deprecation message should be a warning and not an error, shouldn't it?

michalczaplinski:
Yes yes, it's still a work in progress 😛

packages/error/src/index.ts
michalczaplinski: Actually, I think we can throw it. I guess that's as good as throwing any other message.

packages/error/src/index.ts
michalczaplinski: Actually, I removed the babel plugin for now. I realized that we probably won't be using the `invariant` function very much. If I'm wrong, we can always add it back in 🙂 .

luisherranz:
I've been thinking that instead of deprecating the hook, why not just use the `react-intersection-observer` npm package for our own implementation. That way we can be sure that everyone is using the same implementation and that each package doesn't include his own version of useInView. And we can add more features on top if we need it in the future.

michalczaplinski:
@luisherranz I think that we should encourage using the ecosystem and I would be against maintaining our own frontity packages with such narrow use cases for two reasons. 

1. If we have an "official" package, people will use it but eventually a newer, better package will come up (probably with a slightly incompatible API). What to do then? We can use the new solution under the hood and warn the users that it will change in a new version or just stay stuck on "old" package. Neither is ideal.

2. This is more philosophical, but I think that frontity should not be a huge framwork with a solution for everything. I think it's important to make a hard distinction between problems we ARE solving and what is left for the user to figure out. If we have a separate package like that, realistically, some users will know about it and use it, others will not know that it exists or prefer something else and it will lead to more confusion about what is the "right way". 

In this context, I'm thinking about what happened with Meteor as it was trying to have a "Meteor" way of doing things and eventually got surpassed in almost every aspect by various npm packages.

packages/hooks/use-in-view.ts
```diff
@@ -55,6 +56,10 @@ const useInView: UseInView = ({ rootMargin, onlyOnce } = {}) => {
   const [isIntersecting, setIntersecting] = useState(false);
   const ref = useRef();
 
+  warning(
+    "The @frontity/hooks package is deprecated and will be removed in the future version of frontity.\nThe Frontity team recommends that you use https://github.com/thebuilder/react-intersection-observer instead."
+  );
+
```
luisherranz: I'd move this inside a `useEffect` with no deps (`[]`) so we trigger it only once per component.

packages/invariant/index.ts
```diff
@@ -0,0 +1,29 @@
...
+export const warning = (() => {
+  if (isProduction) {
+    return;
+  }
+  let warned = false;
+
+  return (message?: string) => {
+    if (!warned) console.warn(`${prefix}${message || ""}`);
+    warned = true;
+  };
```
luisherranz: I think this closure is wrong because it will trigger the warning only once even for different warnings.

Add tests for more than one warning with the same message and more than one warning with different messages.

packages/invariant/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+// Throw an error if the condition fails
+export function invariant(condition: boolean, message?: string) {
+  if (condition) {
+    return;
+  }
+  // Condition not passed
+
+  if (isProduction) {
+    throw new Error(prefix);
```
luisherranz: Are we sure we want all our error messages in production to just say `"Visit https://community.frontity.org for help"`? Those are the messages that are going to be thrown in production sites, and production site users should not be redirected to our community.

Apart from that, I've just realized that there's not much point in removing the message to save Kbs because the message is not inside a `process.env.NODE_ENV === "production"` conditional and it won't be remove by Webpack.

@SantosGuillamot what do you think? What type of error messages should we show in development and production modes?

luisherranz (changes requested):
It looks like you forgot to add tests and a changest. I've also added some other comments.

luisherranz:
> I think that we should encourage using the ecosystem and I would be against maintaining our own frontity packages with such narrow use cases

Ok, let's recommend `react-intersection-observer` then 🙂

packages/invariant/index.ts
michalczaplinski: I think you're right @luisherranz. Not sure how I missed that.

packages/hooks/use-in-view.ts
michalczaplinski: It was not necessary, because we check if the warning was already triggered in the closure. But now I think it's best to just simplify the code, remove the closure and use `useEffect` otherwise, we'd have to keep track of which warnings have been fired (sounds over-engineered to me already)

luisherranz: Your right!

packages/invariant/index.ts
SantosGuillamot: Not sure how developer tools usually work, but I feel that messages redirecting them to the community should be only included in development mode. I will take a deeper look at other tools, but this is what React codebase says about [warnings and invariants](https://reactjs.org/docs/codebase-overview.html#warnings-and-invariants), just in case it's useful somehow.

packages/hooks/use-in-view.ts
michalczaplinski: I put the warning inside of `useEffect` now :)

packages/invariant/index.ts
michalczaplinski: fixed.

michalczaplinski:
I've updated it to do this:

### Invariant
In **production** we throw a generic error: 
```
Minified exception occurred; use the non-minified dev environment for the full error message and additional helpful warnings.
```
The `babel-plugin-dev-expression` actually changes the signature of the `invariant` function so that we don't ship the whole error message there.

In **development** we just throw the full error. 

This is much closer to what FB is doing with https://github.com/facebook/react/blob/v0.13.3/src/vendor/core/invariant.js

### Warning
Warning just console.warns, and in production does nothing. Simple as that.

@luisherranz Let me know if this is fine with you because I'd like to go ahead and merge it :)

.changeset/happy-camels-obey.md
```diff
@@ -0,0 +1,14 @@
+---
+"@frontity/hooks": minor
+"@frontity/components": patch
+"@frontity/core": patch
+"@frontity/invariant": patch
```
luisherranz: This should be 3 separate changesets instead of one because each change should have its own changeset:

1. Deprecate useInView hook
2. Switch to `react-intersection-observer` in Image
3. Add `@frontity/invariant` package.

Maybe 1 and 2 could use the same changeset because they are related, but 3 has nothing to do.

Apart from that, `@frontity/invariant` should be a major (from 0.1.1 to 1.0.0).

packages/invariant/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export function invariant(condition: boolean, message: string | undefined) {
+  if (!condition) {
+    let error: Error;
+    if (message === undefined) {
+      error = new Error(
+        "Minified exception occurred; use the non-minified dev environment " +
+          "for the full error message and additional helpful warnings."
+      );
+    } else {
+      error = new Error(prefix + message);
```
luisherranz: I'm wondering if this should be the other way around... first the error and then the "Visit frontity community" message.

cc: @SantosGuillamot

luisherranz (changes requested):
Looks great Michal, thanks. I like the idea of using babel-plugin-dev-expression.

My main concern is with the different API of warning and invariant. From what you showed us, FB uses the "condition API" for both:
```js
invariant(condition, message);
warning(condition, message);
```
and we are using it only for invariant:
```js
invariant(condition, message);
warning(message);
```

Maybe we can add the condition to the warning as well. But then, when you work with try-catch a lot, the condition API doesn't make much sense:
```js
try {
  somethingThatThrows();
} catch (e) {
  invariant(true, message)
}
```

So maybe we could add an invariant API that doesn't have a conditional.

```js
try {
  somethingThatThrows();
} catch (e) {
  error(message)
}
```

But then:
- the babel package won't strip the message
- we would need two different names for the warning functions (one for the conditional API and another for the non-conditional API).

Thoughts?

By the way, two additional things:
- I don't mind if we don't use the exact same API of FB, only the core React developers are used to use invariant that way, it's not a popular API.
- We can add the code of `babel-plugin-dev-expression` to our own babel plugin to add more names. I took a look and it's super simple. I prefer to add this ourselves than to have a suboptimal API just to keep using the original package.

----

Oh, apart from that, please remove `packages/invariant/package-lock.json`.

.changeset/happy-camels-obey.md
michalczaplinski: got it 🙁 .

packages/invariant/index.ts
michalczaplinski: I have no opinion 🙂 Whichever works

michalczaplinski:
> Thoughts?

My main thought is that I'm literally happy with whatever API we settle on in the interest of having a short discussion about this 🙂. I agree that this API: 

```
try {
  somethingThatThrows();
} catch (e) {
  invariant(true, message)
}
```
is not perfect but I also don't think anyone would be confused about it for more than 3 seconds to be perfectly honest 🙂 

So, we have the following options:

1. Keep it as is although not perfect 😛

2. Fork the babel-dev-expression-plugin. 
Add the extra functions with signature: `invariant(message)` and `warning(message)` to the package. So, we have a total of 4 functions in `@frontity/invariant`. 
Then extend the babel plugin to handle those as well. 

3. Fork the babel-dev-expression-plugin. 
Change the `invariant` and `warning` to handle a variable number of arguments. So, we have 2 functions in the package in total.
Then extend the babel plugin to handle those cases as well. 

Do you wanna make a decision @luisherranz ?

michalczaplinski:
For some reason 1 test fails about 50% of the time:

```
  1) Image lazy-loading (with Intersection Observer) should work scrolling from right to left:
     CypressError: Timed out retrying: expected '[ <img.frontity-lazy-image.top.left.css-1klzqdy-Root>, 1 more... ]' not to be 'visible'
      at Object.cypressErr 
```

It seems to have passed the last time I pushed, however.

I'm not 100% sure if it's okay to merge it like this, but I also don't know if it's a problem with cypress or on our end.

luisherranz:
> For some reason 1 test fails about 50% of the time:

Yes, for some reason that test is not reliable. I guess it's a bug in Cypress. I have removed it in `beta` (https://github.com/frontity/frontity/commit/379fda87b7786b1b9716ce01af9dd36b974193e2), you can merge the branch if you want.

----

> Do you wanna make a decision @luisherranz?

Ok, let's make some decisions.

1. I don't feel comfortable with `invariant` because I feel it's something quite related to low-level JS libraries. WordPress doesn't have an invariant API and even the FB team doesn't use invariant for the external API (https://github.com/facebook/react/pull/15170): they use `error` and `warn`. So I would stick with `error` and `warning` which are simpler and don't require explanation.

2. I like the idea of https://github.com/facebook/react/pull/15170. We can expose `error` and `warning` directly from our package and avoid the need to install and know about an additional npm package.

```js
import { error, warning } from "frontity";
```

3. I don't see a rush in forking the `babel-plugin-dev-expression` package. It would be nice if it worked out of the box, but it doesn't so I think it's not a priority because right now the number of errors and warnings is so low that it won't make a difference.

The `error` and `warning` functions can be something like:
```js
const error = msg => { 
  if (process.env.NODE_ENV !== "production") {
    throw new Error(msg + sufix);
  } else {
    throw new Error("Minified error...");
  }
}

const warning = msg => {
  if (process.env.NODE_ENV !== "production") {
    console.warn(msg + sufix);
  }
}
```

In the future we can create a production-only plugin that transpiles errors and warnings like this:
```js
// from...
error("My super long message...")
// to...
error()

// from 
warning("My super long message...")
// to
// nothing
```

michalczaplinski:
Yeah, those are good points Luis. I think it's clearly the better way :) 

It appears that the FB themselves might move away from the `invariant` API and seems like people find it confusing: https://github.com/facebook/react/pull/15170#issuecomment-475358712
> Tbh our experience with warning / invariant APIs is people find them confusing (easy to mess up the condition, and you can't do args lazily without a build transform). We might be eventually migrating away from them in React as well.

So, let's remove the babel plugin for now and simplify the API to something like you mentioned: 
```
const error = msg => { 
  if (process.env.NODE_ENV !== "production") {
    throw new Error(msg + sufix);
  } else {
    throw new Error("Minified error...");
  }
}

const warning = msg => {
  if (process.env.NODE_ENV !== "production") {
    console.warn(msg + sufix);
  }
}
```

michalczaplinski:
We can still keep the code as a separate "internal" package right?

And then just import it in `packages/frontity/src/index.ts` and expose `warning` and `error` as named exports?

luisherranz:
> We can still keep the code as a separate "internal" package right?

Oh yes :)

michalczaplinski:
I wonder if `error` should throw an error or just wrap `console.error`. In react, it's just a wrapper around `console.error/warn`.

If it actually throws the error, then the inside a `try/catch`, the code becomes unnecessarily confusing in my opinion.

```js
try {
  if (problem) {
    throw new Error('theres a problem');
  }
} catch (e) {
  // then we rethrow the error...?? 🤷‍♂️
  error(e);
  // I think it's a very rare use case when we want to puropsefully break the application. It's much more likely that we just want to log an error with an extra link to the forum, etc.
}
```

I'm becoming more convinced that it makes no sense to the the API for throwing the errors, just for logging them.

luisherranz:
I thought a lot about that... and I'm not convinced either, but it seems like the less bad solution.

My main concern with doing a `console.log` is that this will duplicate the message in the console:

```js
throw error("some msg...") // logs "some msg..." twice
```

And my final thought was:
- if you don't want to break the app, then it should be a `warning` instead of an `error`.
- `invariant` does throw an error.

But as I said, I'm not 100% convinced either. If you come up with an elegant and easy to understand API that lets you choose between a throw or a `console.error` I'm in.

michalczaplinski:
> if you don't want to break the app, then it should be a warning instead of an error.
I don't think this is necessarily true. I can think of many situations where we can catch the generic JS errors like `somethingsomething is undefined` and `console.error` it with some added context. 

For example if a user tries to access some piece of state that does not exist in `frontity.state`, we could say "Hey, are you sure you have fetched the data for this piece of state". I don't think it would make sense to show a warning there... Of course, this is a useless example, but just to illustrate the idea...

I'm still not sure if we should then throw that error or console.error it 😅 .

In any case, I feel like we've had a lot of discussion about this and this is ultimately not that impactful right now. We can always change the API because no code depends on it yet :) 

I would just like to merge this as is even if the API is stupid. I feel like if we have some more insight from future use cases, we can change it then. 

For your convenience this is `error/index.ts`:
```
const isProduction: boolean = process.env.NODE_ENV === "production";
const suffix = "Visit https://community.frontity.org for help! 🙂";

export const error = (message: string) => {
  if (!isProduction) {
    console.error(message + "\n" + suffix);
  }
};

export const warn = (message: string) => {
  if (!isProduction) {
    console.warn(message + "\n" + suffix);
  }
};
```

luisherranz:
Ok, allow me a last try at this. If you want  `error` not to throw, which is going to be less common, you do: 
`error("Something went wrong but I only want to inform", false)`.

```js
const error = (msg, doThrow = true) => { 
  if (process.env.NODE_ENV !== "production") {
    doThrow ? throw new Error(msg + sufix) : console.error(msg + sufix);
  } else {
    doThrow ? throw new Error("Minified error...") : console.error("Minified error...");
  }
}

const warning = msg => {
  if (process.env.NODE_ENV !== "production") {
    console.warn(msg + sufix);
  }
}
```

michalczaplinski:
Okay, updated the implementation and I think it's good to merge 🙂

PR #257, opened 2019-12-24
2020 Theme CSS fixes

michalczaplinski:

- Fixes the hover styling on Category tags
- Fixes the hover styling on Author and Post Date
- Adds animation to the Search bar
- Fixes the "jumping" when the Search Bar is openend 
- Fixes the overflowing on the document body on the Post page

This is ready to be merged I think, but more fixes or updates might follow.

#### My PR is a:

- 🔝 Improvement

- ✅ Yes!


packages/twentytwenty-theme/src/components/post/post.js
```diff
@@ -114,7 +114,6 @@ const FeaturedImage = styled(FeaturedMedia)`
 
   > div {
     position: relative;
-    width: 100vw !important;
```
michalczaplinski, line 117: This was causing overflow on the Post page and I could not see a good reason why it was there. Am I wrong @segunadebayo 🙂?

michalczaplinski: > This was causing overflow on the Post page and I could not see a good reason why it was there. Am I wrong @segunadebayo 🙂?

Could you get back to me on this when you get a chance @segunadebayo ? 🙂

segunadebayo: Hey Michal. You're right, there's no reason to have it there. It was a carryover code from the mars-theme. Thanks for catching that.

Sorry for my late reply. I'm available today 😀

michalczaplinski: Thanks Segun! :) No worries, I know that you were on holiday yesterday! 🙂

packages/twentytwenty-theme/src/components/search/search-modal.js
```diff
@@ -19,6 +21,12 @@ const SearchModal = ({ state, actions, libraries }) => {
   const inputRef = useRef();
   const containerRef = useRef();
 
+  const transitions = useTransition(isSearchModalOpen, null, {
```
luisherranz, line 26: - I am wondering (out of ignorance) if we can achieve the same animation with CSS and avoid `useTransition`?
- In that case: does it make sense or it is harder to create and/or harder to understand?

packages/twentytwenty-theme/src/components/search/search-modal.js
```diff
@@ -51,46 +59,72 @@ const SearchModal = ({ state, actions, libraries }) => {
...
+        onClick={closeSearchModal}
+      />
+      <div
+        style={{
+          position: `absolute`,
+          top: 0,
+          right: 0,
+          left: 0,
+          zIndex: 9999
         }}
```
luisherranz: Is there a reason to use inline `style` here instead of Emotion?

luisherranz (commented):
LGTM, although I have a couple of questions.

luisherranz:
Oh, another thing: we already have a beta version of twentytwenty, so I think you can add a patch changeset for this :)

packages/twentytwenty-theme/src/components/search/search-modal.js
luisherranz: I think both @DAreRodz and @orballo used to do this type of animations using Emotion's `keyframes`:
https://emotion.sh/docs/keyframes

Is that right guys?

michalczaplinski:
Yes, it's possible with keyframes and you're right, it's probably more
understandable for people who are not super into react but know some CSS.

michalczaplinski:
I've done something like this before: https://czaplinski.io/blog/super-easy-animation-with-react-hooks/ but looking at it again now, I'm not so sure if it's easier than `useTransition`. Will try to simplify it now.

luisherranz:
Awesome :)

michalczaplinski:
@orballo @DAreRodz I've spent too much time trying to add the animations / transitions with just react and emotion and it seems like it's much harder than I thought without using `react-spring`. If you have a solution in mind, then Id really like you to have a look. 

What I want to achieve is animating the search bar opening like in https://wordpress.org/themes/twentytwenty/, so:
1. Have a "dimmer" overlay which gently fades in the `opacity`
2. Have the search bar slide down when "open" and slide back up when "closed"

orballo:
Hum, a similar behavior that I remember to have implemented was the share modal, and it was using `react-transition-group/Transition`:

https://github.com/wp-pwa/saturn-theme/blob/dev-stable/src/pwa/components/Share/index.js

Which I guess is doing something similar to the `Fade` component from your post.

I really don't remember if using `keyframes` instead of transitions had some effect on performance, but if not, `transitions` always seemed simpler to me than using `keyframes`, unless you need a loop animation of some kind?

luisherranz:
I only asked out of ignorance. If there's no easy way to that with keyframes and CSS only, then add the `react-script` transition 👍 

@DAreRodz is back today, maybe he can add his opinion too.

michalczaplinski:
TransitionGroup is nice but the idea was just to NOT add another library to the example to make it more digestable for the first-time user 🙂. I ll wait for @DAreRodz and otherwise use `react-spring`.

DAreRodz:
I remember writing something like your `Fade` component but using transitions and `onTransitionEnd` instead of `onAnimationEnd`. For me, transitions make sense in this case. You could do, for example:

```jsx
shouldRender && (
  <div
    css={{
      filter: `opacity(${show ? '100%' : '0%'})`,
      transform: `translateY(${show ? '0%' : '-100%'})`,
      transition: "filter 1s, tranform 1s"
    }}
    onTransitionEnd={onTransitionEnd}
  >
    {children}
  </div>
)
```

And so you don't need to define animations using keyframes. 

Apart from that, if the component should be unmounted just after the animation/transition and we have to write your own code I would decide to use a third-party library 😂.

PS: maybe the [`Transition`](https://www.react-spring.io/docs/props/transition) component from `react-spring` is more "understandable" than `useTransition`.

michalczaplinski:
OK, thanks @DAreRodz 

>Apart from that, if the component should be unmounted just after the animation/transition and we have to write your own code I would decide to use a third-party library 😂.

That's exactly the problem: We want to either unmount the component right after the animation/transition  or (preferably) hide it with `display: none`. 

However, you cannot transition between `display: none` and `display: block`. You can do it with keyframes but it's not that straighforward in my opinion. 

If there's not veto, I'm gonna bring back the `react-spring`

luisherranz:
Thanks @DAreRodz and @orballo for the insights!

michalczaplinski:
I rebased this branch incorrectly it seems and most of the changes ended up in the 65f7098 commit... 🤷‍♂
I'm not quite sure how to "unfuck" right now I'm just gonna leave it as is 😬

PR #258, opened 2019-12-26
Add bundlewatch

michalczaplinski:

Add bundlewatch github action.

Why?
- The `budget.json` file and the corresponding github action are annoying to remember to edit for almost every PR.
- The lighthouse output is hidden in github behind several clicks on the PR.
- With bundlewatch, we get a badge directly on the PR, no more clicking around.
- What is ACTUALLY important IMO is seeing the bundle size increase for each change added package, etc. and **not** trying to be under some artificially prescribed size.

#### My PR is a:

- 🔝 Improvement

- ✅ Yes!

### DIFF
--- a/.github/workflows/bundlewatch.yml
+++ b/.github/workflows/bundlewatch.yml
@@ -25,7 +25,7 @@ jobs:
           CI: true
 
       - name: Check bundle size
-      - uses: jackyef/bundlewatch-gh-action@master
+        uses: jackyef/bundlewatch-gh-action@master
         with:
           build-script: npm run lerna build
           bundlewatch-github-token: ${{ secrets.BUNDLEWATCH_GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


luisherranz:
A summary of our conversation about this PR: The whole point of limiting the JS file size is having a way to acknowledge that a PR has increased the file size by X Kbs. The current approach of editing the bundle.json file is not straightforward and bundlewatch seems to have a nicer interface, but it doesn't solve the problem either.

We agreed that the best solution would be this type of workflow: Somehow the PR tells you "You are increasing X Kbs the file size, are you sure this is all necessary?" where you can approve the changes. It also needs to be exposed to the person doing the code review, because he/she will be also judging if there's another way to do the same with less Kbs. The most similar example we found of this type of workflow is snapshot testing, where you have to use "U" to update the snapshots and everything is stored in file changes (the PR).

We'll leave this PR on hold for now.

PR #243, opened 2019-12-06
Remove the typescript option from create-package

luisherranz:
#### Description of what you did:

Typescript support in `frontity create-package` it is not implemented yet. Once it is implemented, we can add this back.

#### My PR is a:

- 🐞 Bug fix

#### Is the PR ready to be merged?

- ✅ Yes!

PR #262, opened 2020-01-03
CLI refactor

michalczaplinski:

This is a big PR with many improvements. You have been warned 😄 
- [x] Organize and rename files according to: https://community.frontity.org/t/expose-frontity-commands-to-be-used-programmatically/977. Basically, divide the CLI in to `steps`, `commands` and `interfaces`, with the only interface currently being the CLI interfface.
- [x] Move the event emitter to it's own module and import it wherever it's being used.
- [x] Add new TESTS
- [x] Strip out all the CLI-related code from the steps and commands (so that we can use the same code in the future web interface):
  - [x] `create`
  - [x] `create-package`
  - [x] `subscribe`
  - [x] `info`
- [x] Re-use the `steps` in multiple commands
- [x] Use environmental variables in the `commands` if present

#### My PR is a:

- 🔝 Improvement

- 🚧 Work in progress


package.json
```diff
@@ -23,6 +23,7 @@
     "@typescript-eslint/parser": "^2.0.0",
     "cross-env": "^5.2.1",
     "eslint": "^6.2.2",
+    "eslint-import-resolver-ts": "^0.4.2",
```
michalczaplinski: This package helps resolve the type imports like in https://github.com/frontity/frontity/pull/262/files#diff-d1b3474f889aadcfb57de4d0c9c26b08R2.

michalczaplinski:
@luisherranz Could you see if the deploy failures are some kind of build caching problem? The files build fine locally, but I see `chmod: cannot access '/home/runner/work/frontity/frontity/packages/frontity/dist/cli/index.js': No such file or directory` errors on the CI.

packages/frontity/src/steps/create/index.ts
```diff
@@ -1,80 +1,213 @@
...
-    await step;
-  } catch (error) {
-    if (typeof dirExisted !== "undefined")
-      await revertProgress(dirExisted, options);
-
-    if (emitter) emitter.emit("error", error);
-    else throw error;
+// This function removes the files and directories created
+// with `frontity create`.
+export const revertProgress = async (
```
luisherranz: I'd rename this step so it can be reused easily. Something like `removeDir` for example. And it could receive an optional array of `allowedExistingContent` passed by the command.

luisherranz (commented):
Nice progress :)

I left some comments, sorry if I'm stepping in too early.

luisherranz:
By the way, `dev`, `build` and `serve` are a different kind of command because they belong to `@frontity/core`. Don't worry about them, we'll figure out how to integrate them with the new system when everything is clear.

luisherranz:
> Could you see if the deploy failures are some kind of build caching problem?

When that happens, remove the temp folders (`dist`, `build`, `node_modules`...) and try again.

It looks like the `frontity` package build is now creating an `src` folder inside `dist`:

```
// previously:
frontity/dist/index.js
// now:
frontity/dist/src/index.js
```

I'm not sure why it is doing that in this case, because it doesn't seem you are importing the root `types.ts` file anywhere.

Anyway, you can change the references in the `package.json` to the main, bin, types... and it should work.

package.json
michalczaplinski: Good catch. I saw the package as the solution to the type import problem and didn't notice it ws a fork. The main repo should be used.

packages/frontity/src/commands/create.ts
```diff
@@ -1,97 +1,80 @@
...
+import {
+  normalizeOptions,
+  ensureProjectDir,
+  createPackageJson,
+  createFrontitySettings,
+  cloneStarterTheme,
+  installDependencies,
+  downloadFavicon,
+  revertProgress
+} from "../steps/create";
```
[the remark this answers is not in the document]
michalczaplinski: I thought the same. Just haven't done it yet :)

packages/frontity/src/commands/subscribe.ts
```diff
@@ -1,19 +1,17 @@
...
-  emitter.on("error", errorLogger);
-  emitter.on("subscribe", (message, action) => {
+  emitter.on("cli:subscribe:error", errorLogger);
+  emitter.on("cli:subscribe", (message, action) => {
     if (action) ora.promise(action, message);
     else console.log(message);
   });
 
-  await subscribe(email, emitter);
+  await subscribe(email);
```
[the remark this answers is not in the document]
michalczaplinski: I know you suggested using `event-promised` for this but for the reasons that I've outlined here: https://community.frontity.org/t/expose-frontity-commands-to-be-used-programmatically/977/6?u=mmczaplinski using a promise will not work. 

Unless you had a different idea in mind that I did not understand, of course 😅 

My thinking was the following: 
1. A command calls some step.
2. A step performs some work, whatever it needs to do.
3. A step finished work and wants to log a message to the user in the terminal. For example to let them know that they subscribed successfully
4. A step should not `console.log` anything directly, instead, it will emit an event. For example: "cli:user:subscribed"
5. The event will be caught in the interface and the interface will print the message on the terminal

packages/frontity/src/steps/create/index.ts
```diff
@@ -1,80 +1,213 @@
...
 };
 
-export default async (passedOptions?: Options, emitter?: EventEmitter) => {
-  // This functions will emit an event if an emitter is passed in options.
-  const emit = (message: string, step?: Promise<void>) => {
-    if (emitter) emitter.emit("create", message, step);
-  };
+// This function ensures the path and checks if it's empty or it's a new repo.
+// Also returns a boolean indicating if the dir existed already.
+export const ensureProjectDir = async ({ path }: Options): Promise<boolean> => {
```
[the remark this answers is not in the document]
michalczaplinski: Yup, gotcha, I just didn't get there yet 🙂

packages/frontity/src/commands/create.ts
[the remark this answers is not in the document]
luisherranz: Sorry for overstepping 😅

packages/frontity/src/commands/subscribe.ts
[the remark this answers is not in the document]
luisherranz: Here I'm talking about steps, not commands.

In my opinion, steps should return simple promises, not event emitters because they should either succeed or fail. It should not be a concern of the step to log a message (or to have the intent of logging a message).

packages/frontity/src/steps/create/index.ts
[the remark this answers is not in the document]
luisherranz: Yep, sorry for that then.

michalczaplinski:
Oh right, okay, this makes sense. As long as the above assumption is true
of course (that steps always only succeed or fail and don't return any
information back to the command).

Should we just try it?

packages/frontity/src/commands/subscribe.ts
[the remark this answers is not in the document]
michalczaplinski: Oh right, okay, this makes sense. As long as the above assumption is true
of course (that steps always only succeed or fail and don't return any
information back to the command).

Should we just try it?

michalczaplinski:
> > Could you see if the deploy failures are some kind of build caching problem?
> 
> When that happens, remove the temp folders (`dist`, `build`, `node_modules`...) and try again.
> 
> It looks like the `frontity` package build is now creating an `src` folder inside `dist`:
> 
> ```
> // previously:
> frontity/dist/index.js
> // now:
> frontity/dist/src/index.js
> ```
> 
> I'm not sure why it is doing that in this case, because it doesn't seem you are importing the root `types.ts` file anywhere.
> 
> Anyway, you can change the references in the `package.json` to the main, bin, types... and it should work.

I realized that this was caused because I imported the `package.json` in the `create` command. I think it's not a problem, I can just change the paths like you suggested.

packages/frontity/src/commands/subscribe.ts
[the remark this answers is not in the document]
luisherranz: Sure, go ahead and let's see what happens! :)

By the way, if they succeed, they can return a message using the result of the promise. And if they fail, they can return a message with information about the error. 

```js
try {
  const availablePort = await checkForAvailablePort({ default: 3000 });
  // => Maybe availablePort is 3001 because 3000 is used
} catch (error) {
  console.error(error.message);
  throw new Error(error);
  // => Maybe error.message is "You are not going to believe this but
  // there's not a single available port in your computer! 😱".
}
```

luisherranz:
Ah, you are right, that's the reason. I don't think it's a problem and it happens in other packages when we import `types.ts` from the root anyway...

michalczaplinski:
@luisherranz I have a problem which I'm not sure how to solve at the moment.

I want to use a data structure which is a combination of `EventEmitter` and `Promise` in order to be able to communicate the progress of each step from "inside" the command, as well as be able to listen to events and emit new ones from "outside" the steps (for example in the interface).

I have used https://github.com/Mfron-42/EventPromised, but I realised that it misses the `emit` method on the instance returned from `new EventPromised`. So, I forked the code and added the method myself: https://github.com/frontity/frontity/pull/262/files#diff-d1b3474f889aadcfb57de4d0c9c26b08
(The `EventPromised2` in the same file is a different version that I tried to create myself)

Now, I think that because of how we bind to `this`, the events from the outside are not being listened to. You can verify it by running the test case: https://github.com/frontity/frontity/pull/262/files#diff-22baf25df232f8da4213d53d75aaa932

Any new idea on how to approach it?

michalczaplinski:
@luisherranz I hit upon a weird use case that I didn't know about before which impacts the API a bit:

https://nodejs.org/api/events.html#events_error_events
>If an EventEmitter does not have at least one listener registered for the 'error' event, and an 'error' event is emitted, the error is thrown, a stack trace is printed, and the Node.js process exits.

I wasted some time on this this afternoon unfortunately. It seems like we have to require users to use the `on('error', ... )` callback because it doesn't seem possible to bubble up the error and allow the user to catch it.

Here's the test case to prove it:
```js
  test("Bubbling up the error fails", async () => {
    const options = {
      name: "random-name",
      path: "/path/to/project"
    };

    try {
      const error = new Error("Mocked Error");
      mockedSteps.ensureProjectDir.mockImplementation(async () => {
        throw error;
      });

      const emitter = create(options);

      // Now the test fails but if you uncomment the next line, it works
      // emitter.on("error", err => console.log(err.message));
      await emitter;
    } catch (err) {
      // This will never actually be caught
      expect(err.message).toBe("Mocked Error");
    }
  });
```

luisherranz:
> If an EventEmitter does not have at least one listener registered for the 'error' event, **and an 'error' event is emitted,** the error is thrown, a stack trace is printed, and the Node.js process exits.

No problem, a further argument for using the promise's `reject` instead of emitting an `error` event.

luisherranz:
This works for me:

```js
import EventPromised from "event-promised";

describe("EventPromised", () => {
  it("should throw an error with reject", async () => {
    const myCommand = () =>
      new EventPromised((resolve, reject, emit) => {
        reject("some error");
      });

    await expect(myCommand()).rejects.toBe("some error");

    try {
      await myCommand();
    } catch (error) {
      expect(error).toBe("some error");
    }
  });
});
```

https://codesandbox.io/s/blue-meadow-hnqv3 (`__tests__` folder)

michalczaplinski:
It's fine if you explicitly `reject()` inside of the EventPromised callback. But if there is an unexpected error thrown inside of the `create` command, the user cannot catch it with a regular `try/catch`.

They have to explicitly listen for the event:`on('error', (err) => {})

Let me know if that's clear, otherwise I can explain in a quick call. 

API-wise, It's a bit annoying to have this. Because the user would have to do something like this:

```
  const ee = create(options);
  ee.on('error', (err) => {
    ...do something with the error
  }
  await ee;
```

luisherranz:
Yes, I agree. It would be a super annoying API!

But I don't fully understand what you mean yet. Why would you have unexpected errors inside the `EventPromised` callback? Why wouldn't you be able to catch them and reject properly like this?

```js
import EventPromised from "event-promised";

describe("EventPromised", () => {
  it("should throw an error with reject", async () => {
    const someError = new Error("some error");

    const myCommand = () =>
      new EventPromised((resolve, reject, emit) => {
        try {
          throw someError;
        } catch (error) {
          reject(error);
        }
      });

    await expect(myCommand()).rejects.toBe(someError);

    try {
      await myCommand();
    } catch (error) {
      expect(error).toBe(someError);
    }
  });
});
```

https://codesandbox.io/s/tender-napier-up6th

Could you please create a sandbox so I can understand what you mean?

michalczaplinski:
I think it should work if we pass the `reject` from the EventPromised to the `create` command like you suggested.

I just wanted to avoid that because the command should not need to know about the rejecting of the Promise that wraps it. I guess there's no avoiding it if we want a nicer API :)

packages/frontity/src/__tests__/create-package.cli.test.ts
```diff
@@ -1,19 +1,17 @@
...
+import * as utils from "../utils";
 
-jest.mock("../../functions/create-package");
 jest.mock("inquirer");
-jest.mock("events");
-jest.mock("../../utils");
 
 const mockedInquirer = inquirer as jest.Mocked<typeof inquirer>;
-const mockedUtils = utils as jest.Mocked<typeof utils>;
+const mockedUtils = utils as any;
```
michalczaplinski, line 11: It's just easier this way to fix: `Cannot assign to 'isFrontityProjectRoot' because it is a read-only property`

.eslintrc
```diff
@@ -23,6 +23,9 @@
     "import/resolver": {
       "node": {
         "extensions": [".js", ".jsx", ".ts", ".tsx"]
+      },
+      "typescript": {
+        "alwaysTryTypes": true
```
michalczaplinski, line 28: option for https://github.com/alexgorbatchev/eslint-import-resolver-typescript

packages/frontity/package.json
```diff
@@ -24,13 +24,13 @@
...
-    "start": "node dist/commands.js",
-    "dev": "ts-node src/commands.ts",
+    "start": "node dist/src/cli/index.js",
+    "dev": "ts-node src/cli/index.ts",
     "ts": "ts-node",
     "test": "jest --watch",
     "test:ci": "jest --colors",
     "test:inspect": "node --inspect ../../node_modules/jest/bin/jest.js --watch --no-cache --runInBand",
-    "build": "tsc --project ./tsconfig.build.json",
+    "build": "tsc --project ./tsconfig.build.json && cp -a templates dist/templates",
```
michalczaplinski, line 33: I'm not happy that we do this but I couldn't think of a better way.

packages/frontity/src/__tests__/create.test.ts
```diff
@@ -42,7 +42,8 @@ describe("create", () => {
     await create(options);
     expect(mockedSteps.createFrontitySettings).toHaveBeenCalledWith(
       "js",
-      options
+      options.name,
+      options.path
```
michalczaplinski, line 46: Now, we pass individual options instead of an `Options` object to steps.

packages/frontity/src/__tests__/create.test.ts
```diff
@@ -78,23 +91,44 @@ describe("create", () => {
...
+
+    try {
+      const error = new Error("Mocked Error");
+      mockedSteps.createPackageJson.mockImplementation(() => {
+        throw error;
+      });
+
+      const emitter = create(options);
+      await emitter;
+      throw new Error("This should not be reached");
```
michalczaplinski, line 103: Is there a more elegant way?

packages/frontity/src/cli/create-package.ts
```diff
@@ -2,53 +2,53 @@ import ora from "ora";
...
+import {
+  errorLogger,
+  isFrontityProjectRoot,
+  isThemeNameValid,
+  log
+} from "../utils";
+import { Options } from "../steps/create-package";
 
 //  Command:
-//    create-package [name] [--typescript]
```
michalczaplinski, line 11: Typescript support here was supposed to be removed in an earlier PR #243 but it was incomplete

packages/frontity/src/cli/create-package.ts
```diff
@@ -2,53 +2,53 @@ import ora from "ora";
...
 //    1. validate project location
 //    2. ask for the package name if it wasn't passed as argument and validate
 //    3. ask for the package namespace if it wasn't passed as argument
 //    4. create package
 
-export default async (
-  name: string,
-  {
-    namespace,
-    typescript
```
michalczaplinski, line 23: see comment above.

packages/frontity/src/cli/index.ts
```diff
@@ -26,26 +26,19 @@ tsNode.register({
...
-const { version } = JSON.parse(
-  readFileSync(resolve(__dirname, "../package.json"), { encoding: "utf8" })
-);
+import { default as create } from "./create";
+import { default as createPackage } from "./create-package";
+import { default as subscribe } from "./subscribe";
+import { default as unknown } from "./unknown";
+import { default as info } from "./info";
+
+import packageJson from "../../package.json";
```
michalczaplinski, line 37: We can just import the JSON file, the only drawback is that it creates `src` inside the `dist`folder.

packages/frontity/src/cli/index.ts
```diff
@@ -56,14 +49,16 @@ program
   .command("create [name]")
   .option("-t, --typescript", "Adds support for TypeScript")
   .option("-c, --use-cwd", "Generates the project in the current directory.")
+  .option("-n, --no-prompt", "Skips prompting the user for options")
   .description("Creates a new Frontity project.")
-  .action(create);
+  .action((name, { ...args }) => create({ name, ...args }));
```
michalczaplinski, line 55: It's not ideal that we have to remap the arguments here, but there's a mismatch between the proposed API and the callback args from `commander`.

packages/frontity/src/cli/subscribe.ts
```diff
@@ -0,0 +1,35 @@
...
+        message: "Please, enter your email:"
+      }
+    ];
+    const answer = await prompt(subscribeQuestion);
+    email = answer.email;
+  }
+
+  const emitter = subscribe(email);
+
+  emitter.on("cli:subscribe:error", errorLogger);
```
michalczaplinski: Whoops forgot to remove the namespaces here still

packages/frontity/src/commands/subscribe.ts
```diff
@@ -0,0 +1,21 @@
+import { EventPromised } from "../utils/eventPromised";
+import { subscribe } from "../steps";
+
+// TODO:  make param an object
+const subscribeCommand = async (
+  email: string,
+  emit: (event: string, ...value: any[]) => void
+) => {
+  try {
+    emit("cli:subscribe:message", "Subscribing to frontity");
```
michalczaplinski: namespaces to remove

michalczaplinski:
@luisherranz Please review 🙂 Some things to note:

1. There are a couple tests that I would still like to add for `create-package`
2. Currently tests don't pass because because snapshots are different in CI and locally. It's a small fix.
3. I havent removed the all the `on(error)` that we talked about because of time, but it's a small fix.
4. I haven't really touched the `dev`, `serve` and `build` commands but from a brief look I think that they can follow more or less the same structures as the other commands.
5. I have **not** removed any existing tests and only added new ones and adapted the existing ones.

luisherranz (approved):

PR #266, opened 2020-01-15
Implement script component and processor

### LINKED ISSUE #283: Implement script processor
**Feature:** [Execute script tags found in content](https://community.frontity.org/t/execute-script-tags-found-in-content/1060)
**User Story**: Script tags included in the content should be executed.

- [x] Create the script processor.
- [x] Unit tests.
- [x] Update docs.

Branch: `script-processor`
[Sprint Card](https://github.com/orgs/frontity/projects/2#card-32044424)

### LINKED ISSUE #284: Implement script component
**Feature:** [Execute script tags found in content](https://community.frontity.org/t/execute-script-tags-found-in-content/1060)
**User Story**: Script tags included in the content should be executed.

- [x] Create the script component.
- [x] e2e tests.
- [x] Update docs.

Branch: `script-processor`
[Sprint Card](https://github.com/orgs/frontity/projects/2#card-32044424)

iamuchejude:

Implement a script processor to execute script tags from embeds.

#### My PR is a:

- 🔝 Improvement

- ✅ Yes!

#### Feature Request

https://community.frontity.org/t/execute-script-tags-found-in-content/1060

#### Docs Changes
https://app.gitbook.com/@frontity-tech/s/frontity/~/drafts/-LzGMLsfZI_8PDDtUgAv/api-reference-1/frontity-components/@drafts

----

Fixes #284.
Fixes #283.


luisherranz (changes requested):
Looks great to me so far @iamuchejude :)

We use [`changesets`](https://github.com/atlassian/changesets) for versioning. You need to [add a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) to this PR. Take a look at the changeset docs and if you have any doubt let us know.

I've also left a comment about a change I think we should make.

Apart from that, maybe we can have a quick call tomorrow to see how to add e2e tests for this.

packages/components/script.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useRef, useEffect } from "react";
+import { connect } from "frontity";
+import { Connect, Package } from "frontity/types";
+
+type ScriptType = React.FC<Connect<Package, { src: string }>>;
+
+const generateId = () => Math.trunc(Math.random() * 9999999).toString();
+
+const Script: ScriptType = ({ state, ...props }) => {
+  const scriptId = useRef(generateId());
```
luisherranz: Instead of `generateId` which will break hydration I would try to keep the reference to the `script`. Something like this:

```js
const Script: ScriptType = ({ state, ...props }) => {
  let script = null;
  const { src } = props;

  useEffect(() => {
    if (state.frontity.rendering === "csr") {
      script = window.document.createElement("script");
      script.src = src;
      script.async = true;
      script.charset = "utf-8";

      window.document.body.appendChild(script);
    }

    return () => {
      if (script) window.document.body.removeChild(script);
    };
  }, []);

  return <script async src={src} charSet="utf-8" />;
};
```

packages/html2react/processors/image.tsx
```diff
@@ -23,6 +23,8 @@ const image: Processor = {
 export default image;
 
 const ContentImage: React.FC<any> = props => {
+  console.log(props);
```
luisherranz: Careful with the console logs.

packages/mars-theme/src/index.js
```diff
@@ -44,7 +45,7 @@ const marsTheme = {
        * Add a processor to `html2react` so it processes the `<img>` tags
        * inside the content HTML. You can add your own processors too
        */
-      processors: [image]
+      processors: [image, script]
```
luisherranz: Instead of adding the processor on each theme, I think we can add it by default in `html2react`:
https://github.com/frontity/frontity/blob/dev/packages/html2react/src/index.ts#L11

because we are "fixing something we broke". It should be fixed by default.

luisherranz (changes requested):

packages/html2react/processors/script.tsx
```diff
@@ -0,0 +1,20 @@
+import { Processor, Element } from "../types";
+import Script from '@frontity/components/script';
+
+const script: Processor = {
+  test: node => node.type === "element" && node.component === "script",
```
luisherranz: We need to add a name (I guess `"script"` it's fine) and a priority, which I think it should be at least 20 so other people can overwrite this.

luisherranz (changes requested):

luisherranz:
Right now it looks like the best idea is to `return null` in the `<Script>` component because if not it will be executed/loaded twice in SSR. We need to figure out how to make it work with SSR when Frontity doesn't load.

examples/mars-theme-example/frontity.settings.js
```diff
@@ -32,7 +32,7 @@ export default {
       name: "@frontity/wp-source",
       state: {
         source: {
-          api: "https://test.frontity.io/wp-json"
+          api: "http://frontity.local/wp-json"
```
luisherranz: You need to revert this change.

packages/components/script.tsx
```diff
@@ -0,0 +1,35 @@
...
+type ScriptType = React.FC<
+  Connect<Package, { src?: string; code?: string; dataId?: string }>
+>;
+
+const Script: ScriptType = ({ state, ...props }) => {
+  let script = null;
+  const { src, code, dataId } = props;
+
+  useEffect(() => {
+    if (state.frontity.rendering === "csr") {
```
luisherranz: I think there's not much sense to use `state.frontity.rendering` now that we are returning null.

packages/components/script.tsx
```diff
@@ -0,0 +1,35 @@
...
+      script = window.document.createElement("script");
+
+      if (src) script.src = src;
+      if (code) eval(code);
```
luisherranz: I think we don't need to add a script tag if what we are going to do is to use `eval` for the code.

packages/components/script.tsx
```diff
@@ -0,0 +1,35 @@
...
+      if (dataId) script.id = dataId;
```
luisherranz: Is there any reason to use `dataId` instead of `id` directly?

packages/components/script.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+      script.async = true;
+      script.charset = "utf-8";
```
luisherranz: We can remove `charset`. It looks like it's been deprecated: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

luisherranz (changes requested):
Looks great @iamuchejude :)

Last things we can explore:
- Do e2e tests for the processor.
- Pass down other `script` props, like `type`.

luisherranz:
Let's only apply the processor if type is missing or it's javascript (text/javascript, application/javascript...).

luisherranz (approved):
Great work @iamuchejude, ready for merge :)

Is the code finished?

luisherranz:
I've merged `dev` to solve a problem with the cypress library not working in Ubuntu 18.04.

iamuchejude:
@luisherranz The PR is ready to be merged. :)

luisherranz:
Awesome. Thanks Jude!

If we merge this PR it will close the related issues and the "update docs" tasks are not finished yet so we need to figure out how to wait and review the documentation. I'm going to wait until I talk with @SantosGuillamot about this.

iamuchejude:
Thanks, Luis! 🙌

I'd have a meeting with @SantosGuillamot on Monday to talk about documentation. I think that's all that is left for the issues to be resolved and closed.

luisherranz (changes requested):
Great work @iamuchejude.

- I've added an e2e test to check that the script is actually being loaded.
- I've moved the react testing library dependency to `devDependencies`.
- I've added a unit test to check if the valid types are being picked up fine, but it fails. Can you check it out?

I'll review the docs next 👍

iamuchejude:
Okay @luisherranz

luisherranz (changes requested):
Docs are great, thanks. Could you please add a couple of examples of usage of the `<Script>` component? One with `src` and another with the `code` prop. Thanks! :)

luisherranz (approved):
Thanks @iamuchejude!

I've added TypeScript props to the Html2react processor but other than that it's fine, thanks!

Be careful with committing the lerna logs and package-lock files (we must add those to the .gitignore file).

luisherranz:
One test is failing. I'll fix it later.

luisherranz:
Fixed. `script` was missing because I accidentally removed it when I merged with the latest `dev` branch.

Issue #237, opened 2019-11-26
Array elements from package state are duplicated on `csr`


orballo:
**_Bug report_**

- [X] I've searched for existing issues
- [X] I've checked documentation: https://docs.frontity.org
- [ ] I've checked community forum: https://community.frontity.org

**Describe the expected behavior**

When I define an array inside a theme `index.js` file like this:
```js
const theme = {
  name: "noticias-coches-theme",
  roots: {
    theme: Root
  },
  state: {
    theme: {
      nav: [ { label: "One", link: "/one" }, { label: "Two", link: "/two" } ]
    }
  }
};
```

The menu should be rendered once, as if defined in `frontity.settings.js`.

**Describe the observed behavior**

Instead the menu is rendered once in `ssr` and duplicated in every `csr`.

DAreRodz:
I've done some research and the problem is happening when calling [mergePackages](https://github.com/frontity/frontity/blob/f6de112114132f1758ee4a12ad09860e247372d3/packages/core/src/client/store.ts#L15) during Client Side Rendering (more specifically in [this line](https://github.com/frontity/frontity/blob/f6de112114132f1758ee4a12ad09860e247372d3/packages/core/src/utils/merge-packages.ts#L33L35)).

luisherranz:
`deepmerge` accepts an `arrayMerge`. The problem is that is we go back to overwrite, you cannot add elements to arrays in the state declaration because the next array will delete the first ones.

luisherranz:
Let me think about it...

luisherranz:
Ok, I guess we can try to do an overwrite at that point because the merge of the arrays already happened on the server and we are only hydrating the state.

PR #267, opened 2020-01-15
Create @frontity/google-analytics package

DAreRodz:

This package will allow any frontity project to integrate with Google Analytics.

~Note: work is not completed yet, this PR is intended to be merged with beta.~

#### My PR is a:

- 🚀 New feature

- ✅ Yes!


packages/analitics/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@frontity/analytics",
+  "version": "0.1.0",
+  "description": "Analytics utilities for »Frontity",
```
luisherranz: I'd say more something like: `This package contains the base types and actions to build analytics packages for Frontity`

packages/analitics/types.ts
```diff
@@ -0,0 +1,40 @@
...
+export type Pageview = {
+  page: string;
+  title: string;
+};
+
+export type Event = {
+  category: string;
+  action: string;
+  label: string;
+};
```
luisherranz: I'm not sure if we should use `action`, `category` and  `label`. Feels too coupled to Google Analytics. Most analytics tools have something like `name` and `payload` for events.

packages/analitics/types.ts
```diff
@@ -0,0 +1,40 @@
...
+      sendPageview: Action<Analytics, Pageview>;
+      sendEvent: Action<Analytics, Event>;
+    };
+  };
+  state: {
+    analytics: {
+      namespaces: Record<string, string>;
+    };
+    router?: {
+      link: string;
```
luisherranz: I'm yet not sure if it's a better idea to hardcode external types like this (for the sake of simplicity) or to recommend importing the original type like this.

```js
router?: {
  link: Router["state"]["router"]["link"];
}
```

One scenario when this could be relevant is if you are updating this package to be used with `@frontity/router@^2.0.0` and in that version `router` has changed `link` from a string to an object for example. It's not a big deal.

Any other scenarios where this could be more important?

packages/analitics/index.ts
```diff
@@ -0,0 +1,49 @@
+import { observe } from "frontity";
+import Analytics from "./types";
+
```
luisherranz: I think that apart from the named exports, I would also export the default export (containing `state` and `actions`) in this file.

Did you think about a reason not to do so? Do you think it can be misleading?

packages/analitics/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  observe(async () => {
+    const { link } = state.router;
+
+    // Wait data to be ready.
+    if (state.source.get(link).isReady) {
+      /**
+       * Execute action in the next tick so we can
+       * ensure the title has changed by React.
+       */
+      setTimeout(() => {
```
luisherranz: I think this can be kind of dangerous... How can we make sure this is not going to break? What happens if React takes more time to update the `document.title`? Maybe we need to think of a better technique to extract the title.

packages/google-analytics/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+              eventAction: event.action,
+              eventLabel: event.label
+            });
+          });
+      }
+    }
+  },
+  state: {
+    analytics: {
+      namespaces: { ga: "ga" }
```
luisherranz: What's the advantage of using an object over an array `["ga"]`. Is it just to avoid [this bug](https://github.com/frontity/frontity/issues/237) or does it have a real advantage?

packages/google-analytics/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const googleAnalytics: GoogleAnalytics = {
+  roots: { ga: Root },
+  actions: {
+    analytics: {
+      sendPageview,
+      sendEvent,
+      afterCSR
+    },
+    ga: {
```
luisherranz: I'm not sure if the namespaces should be `googleAnalytics`, `gtmAnalytics`, `mixpanelAnalytics`, `amplitudeAnalytics` and so on... thoughts?

packages/google-analytics/src/components/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { getTrackerName } from "..";
+
+const Root: React.FC<Connect<GoogleAnalytics>> = ({ state }) => {
+  // Get Tracking ids from state.
+  const { trackingIds, trackingId } = state.ga;
+  const ids = trackingIds || (trackingId && [trackingId]) || [];
+
+  // Render Tracker code.
+  return (
+    <script
```
luisherranz: It seems that this script should go in the `<head>` section:
> Copy and paste the global site tag to each web page you want to measure. Add the tag right after the opening <head> tag on each page.

https://support.google.com/analytics/answer/1008080?hl=en

luisherranz:
I've made a bunch of comments but it looks great David. I see you've tried to use `MutationObserver` to extract the `document.title`. Maybe we can have a quick call to discuss that.

luisherranz:
**Important:** I've changed the base of this PR to `beta`. Please rebase the branch.

packages/analitics/index.ts
DAreRodz: I wanted to clearly show that this is not a package but a library, just looking at the code.

packages/analitics/index.ts
DAreRodz: Here I (hastily) assumed that React would render syncronously as soon as `data.isReady` was `true`, so it would be `safe` to use `setTimeout` in this case.

Maybe I assumed wrong 😅

packages/analitics/package.json
DAreRodz: Much better!

packages/google-analytics/src/components/index.tsx
DAreRodz: Good catch 👍

packages/google-analytics/src/index.ts
DAreRodz: Using `ga` as the namespace didn't convince me. I prefer your idea. 👍

packages/google-analytics/src/index.ts
DAreRodz: It's just to avoid the bug, ofc.

PR #288, opened 2020-01-24
CLI: Add theme param to create

### LINKED ISSUE #132: `npx frontity create` adds mars-theme's settings even if a different theme has been set
By default, `npx frontity create` uses `@frontity/mars-theme` as the project's theme, but when you set a different theme (e.g. `osw-theme-example`), the generated `frontity.settings.js` file doesn't contain the theme package that has been set but the default `mars-theme` and their settings.

The problem is [here](https://github.com/frontity/frontity/blob/frontity%401.0.1/packages/frontity/src/functions/create/steps.ts#L124L141).

michalczaplinski:

This PR adds the ability to pick a theme when running the `frontity create`.

- A new option `--theme` which allows specifying the starter theme on the command line. 
- If the theme is not specified, the user can pick a theme from an interactive prompt
- If the command is run with `--no-prompt`, fall back to the default theme (mars-theme)

(A few unrelated and small improvements to the tests for the CLI also made it here, by the way)

#### My PR is a:

- 🚀 New feature

- 🚧 Work in progress

----

Fixes #132

### DIFF
--- a/.changeset/tidy-pets-yawn.md
+++ b/.changeset/tidy-pets-yawn.md
@@ -1,0 +1,5 @@
+---
+"frontity": minor
+---
+
+Add a new option `--theme` option which allows specifying the starter theme on the command line. If the theme is not specified, the user can pick a theme from an interactive prompt.
--- a/.changeset/tidy-pets-yawn.md
+++ b/.changeset/tidy-pets-yawn.md
@@ -2,4 +2,4 @@
 "frontity": minor
 ---
 
-Add a new option `--theme` option which allows specifying the starter theme on the command line. If the theme is not specified, the user can pick a theme from an interactive prompt.
+Add a new option `--theme` which allows specifying the starter theme on the command line. If the theme is not specified, the user can pick a theme from an interactive prompt.
--- a/packages/frontity/package.json
+++ b/packages/frontity/package.json
@@ -73,6 +73,7 @@
     "@types/stack-trace": "0.0.29",
     "@types/tar": "^4.0.0",
     "jest": "^24.7.1",
+    "lodash.omit": "4.5.0",
     "ts-jest": "^24.0.2"
   }
 }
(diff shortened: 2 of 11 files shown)


DAreRodz (changes requested):
Hey Michal!

There is a problem with the `create` command. When you select a different theme than `mars-theme` the `frontity.settings.js` file won't include the theme you choose but `mars`. It's the [`createFrontitySettings`](https://github.com/frontity/frontity/blob/add-theme-picker/packages/frontity/src/steps/index.ts#L118L172) function (that is being called [here](https://github.com/frontity/frontity/blob/add-theme-picker/packages/frontity/src/commands/create.ts#L58)). Could you change that function and add some tests?

Code looks fine, by the way. 😊

luisherranz:
I've been thinking about this during the last few days and I think we can solve it using the `frontity.config.js` file: https://community.frontity.org/t/defining-default-settings-for-frontity-packages/1139

Please read that feature topic and let me know what you think (in the topic).

michalczaplinski:
> There is a problem with the `create` command. When you select a different theme than `mars-theme` the `frontity.settings.js` file won't include the theme you choose but `mars`. It's the [`createFrontitySettings`](https://github.com/frontity/frontity/blob/add-theme-picker/packages/frontity/src/steps/index.ts#L118L172) function (that is being called [here](https://github.com/frontity/frontity/blob/add-theme-picker/packages/frontity/src/commands/create.ts#L58)). 

oh, nice catch @DAreRodz, thanks! 🙂

luisherranz:
Yeah, that is actually the bug!
https://github.com/frontity/frontity/issues/132

Sorry we forgot to add it to the sprint. I've done so now and added a `Fixes` to this PR.

DAreRodz (approved):
Looks good to me! 👌

DAreRodz:
PS: I guess the default settings feature would be implemented in another PR. This is totally fine, isn't it?

luisherranz (approved):
Great work @michalczaplinski!

luisherranz:
> PS: I guess the default settings feature would be implemented in another PR. This is totally fine, isn't it?

Yeah, I guess this is fine for now. 

For reference, this should be the next step: https://community.frontity.org/t/defining-default-settings-for-frontity-packages/1139/3

PR #291, opened 2020-01-28
Add status codes and error messages when fetch responses are 4xx or 5xx

### LINKED ISSUE #286: Add status codes and error info to data objects in wp-source
**Feature:** [Add status code and error messages to data](https://community.frontity.org/t/add-status-code-and-error-messages-to-data/1125)
**User Story**: Add status codes and errors to data

- [x] Capture fetch errors.
- [x] Populate the data.
- [x] Add new properties to TypeScript.
- [x] Unit tests.
- [x] Update docs.

Branch: `s0-f3-status-errors-in-data`
[Sprint Card](https://github.com/orgs/frontity/projects/2#card-32117718)

### LINKED ISSUE #287: Throw when code errors are caught inside actions.source.fetch
**Feature:** [Add status code and error messages to data](https://community.frontity.org/t/add-status-code-and-error-messages-to-data/1125)
**User Story**: Add status codes and errors to data

- [x] Capture fetch errors.
- [x] Figure out if it's a JS error.
- [x] Unit tests.
- [x] Update docs.

Branch: `s0-f3-status-errors-in-data`
[Sprint Card](https://github.com/orgs/frontity/projects/2#card-32117718)

michalczaplinski:

Add status codes and error messages when fetch responses are 4xx or 5xx: #286 

#### My PR is a:

- 🔝 Improvement

- 🚧 Work in progress

#### Doc changes

https://github.com/frontity/gitbook-docs/pull/7
https://app.gitbook.com/@frontity-tech/s/frontity/~/diff/drafts/-M-KvDzW6IvP4834Kij_/revisions/-M-Kv753JbsCVEm3ULp-/api-reference-1/frontity-head-tags/@activity/-M-KvG-AetB6JBH26FTr


packages/wp-source/src/errors.ts
```diff
@@ -0,0 +1,13 @@
...
+  public status: number;
+  public statusText: string;
+
+  constructor(message = "", status: number, statusText: string = message) {
+    super(message);
+    this.message = message;
+    this.status = status;
+    this.statusText = statusText;
+  }
+}
```
luisherranz: Extending `Error` is a good idea :)

I think this can be moved to `@frontity/source` because it's a shared implementation between all the source packages.

We have to move the types of `@frontity/source` to `@frontity/source/types` (I'm sorry, this was meant to be done a long time ago) and use the `index.ts` to export things like this:

```js
import Source from "@frontity/source/types";
import { ServerError } from "@frontity/source";
```

packages/wp-source/src/libraries/api.ts
```diff
@@ -41,16 +42,25 @@ class Api {
     const query = stringify(params, { arrayFormat: "bracket", encode: false });
 
     // Send request
-    return fetch(`${requestUrl}${query && "?"}${query}`);
+    return fetch(`${requestUrl}${query && "?"}${query}`).then(
+      (response: Response) => {
+        if (!response.ok) {
+          const { status, statusText } = response;
+          throw { status, statusText, isServerError: true };
```
luisherranz: I guess this needs to be `ServerError` right?

packages/wp-source/src/actions.ts
```diff
@@ -53,13 +55,23 @@ const actions: WpSource["actions"]["source"] = {
...
     } catch (e) {
-      console.log(e);
-      // an error happened
-      source.data[route] = {
-        is404: true,
-        isFetching: false,
-        isReady: false
-      };
+      // It's a server error (4xx or 5xx)
+      if (e instanceof ServerError) {
```
luisherranz, line 59: Maybe I'm wrong but I think that for this check to work this way, you need to change the `api.get`:
``` js
if (!response.ok) {
  const { status, statusText } = response;
  throw { status, statusText, isServerError: true }; // <-- this needs to throw a ServerError
```

luisherranz (changes requested):
It works for me, except for a couple of things I commented on the code. Great work Michal! 👍 

But I didn't review the `Data` TypeScript part. I think that should be done by @DAreRodz so please wait for his comments/approval. I know we've battled before with `Data` booleans being true instead of boolean and there was a reason.

packages/wp-source/src/libraries/api.ts
michalczaplinski: siiii 🙂

packages/wp-source/src/actions.ts
michalczaplinski: yes, I had a different data structure before but refactored it to extend from `Error` and must have missed it here.

packages/wp-source/src/errors.ts
michalczaplinski: oh, okay, gotcha 🙂

michalczaplinski:
> It works for me, except for a couple of things I commented on the code. Great work Michal! 👍
> 
> But I didn't review the `Data` TypeScript part. I think that should be done by @DAreRodz so please wait for his comments/approval. I know we've battled before with `Data` booleans being true instead of boolean and there was a reason.

Yeah, I spoke with @DAreRodz already briefly about this and I'm not 100% sure if this is the right solution but it seems to work "for now"... I think that the [union type](http://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions) for `Data` might not be the right way to model our data. Perhaps we should have a "hierarchy" of interfaces like: 

```typescript
interface BaseData {
  id: string;
  ...
}

interface PostData extends BaseData { ... }
interface AuthorData extends BaseData { ... }

export interface Data extends PostData, AuthorData, Every, Other, Data, Interface;
```

@DAreRodz Do you have some thoughts on this?

DAreRodz:
Indeed, the `Data` type would be better modeled as a hierarchy (it is, actually). But, if I remeber well, we used interfaces in the first place but there were some problems trying to override properties. I think that's the reason we ended using types and that Merge trick.

DAreRodz:
I think I've fixed the problems with the `Data` types. Check it out, @michalczaplinski!

PS: The `TypeError: body used already for` is still there. I'll take a look at it as well.

michalczaplinski:
Had to rebase and force push because I accidentally put the 9d286142b74 commit in this branch.

DAreRodz:
I just saw that you found what was causing the problem, haha. Good. 👍

packages/wp-source/src/__tests__/__snapshots__/actions.tests.ts.snap
```diff
@@ -1,23 +1,22 @@
...
-    "isReady": true,
-    "type": "example",
+    "isFetching": true,
+    "isReady": false,
   },
 }
 `;
 
-exports[`fetch returns 404 is fetch fails 1`] = `
+exports[`fetch does nothing if data exists 1`] = `
```
michalczaplinski, line 12: This is actually a new test, but in the diff it looks like the old test was modified unfortunately.

packages/wp-source/src/actions.ts
```diff
@@ -26,7 +28,7 @@ const actions: WpSource["actions"]["source"] = {
         isReady: false,
         isFetching: false
       };
-    } else if (data.isReady || data.isFetching || data.is404) {
+    } else if (data.isReady || data.isFetching || data.isError) {
```
michalczaplinski, line 31: Now we have a `isError` property for all errors instead of just `is404`.

packages/wp-source/src/libraries/__tests__/api.tests.ts
```diff
@@ -1,12 +1,15 @@
+import * as frontity from "frontity";
 import Api from "../api";
-import { fetch } from "frontity";
+import { Response as NodeResponse } from "node-fetch";
 
 jest.mock("frontity");
-const mockedFetch = (fetch as any) as jest.Mock;
+const mockedFrontity = frontity as jest.Mocked<typeof frontity>;
```
michalczaplinski, line 6: I had to change how we mock the `frontity` module, because doing `fetch as any` in the line above makes it lose the type information.

packages/wp-source/src/libraries/handlers/__tests__/post-type.tests.ts
```diff
@@ -22,9 +22,14 @@ beforeEach(() => {
 describe("postType", () => {
   test("returns 404 if not found", async () => {
     // Mock Api responses
-    api.get = jest.fn().mockResolvedValue(mockResponse([]));
+    api.get = jest.fn(_ => Promise.resolve(mockResponse([])));
+
+    // TODO: The test fails if we use this form instead 🤷‍♂️:
+    // api.get = jest.fn().mockResolvedValue(mockResponse([]));
```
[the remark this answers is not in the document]
DAreRodz: The difference is that in the first case you instantiate a new response everytime `api.get` is executed. In the second case it returns the same instance, and it fails because `response.json()` can only be run once (which is weird).

When I implemented this part, my idea was to use the response object to be able to access its headers in addition to the body object, but it turns out not to be a good decission, I guess.  😅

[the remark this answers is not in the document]
michalczaplinski: right, thanks for the explanation 🙂 I ll update the comment then.

luisherranz:
I'm finishing the review and I'm taking the opportunity to add a couple of changes to the `source` types. I have to go now but I'll try to push the changes this night.

luisherranz (approved):
Ok, everything looks great. Great work @michalczaplinski! 👍

luisherranz (changes requested):
Hey, this needs a changeset.

luisherranz (approved):
Perfecto, thanks!

PR #296, opened 2020-01-31
Fix: Critical bug causes a rejected Promise to be swallowed.

michalczaplinski:

The fix is self-explanatory. We used to not reject the promise that failed which caused errors to be swallowed in actions. This was discovered In #291 

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes!

### DIFF
--- a/packages/connect/src/create-store.js
+++ b/packages/connect/src/create-store.js
@@ -21,12 +21,18 @@ const getSnapshot = obj => {
 
 const convertToAction = (fn, instance) => (...args) => {
   const first = fn(instance);
-  if (first instanceof Promise)
-    return new Promise(resolve => first.then(() => resolve()));
+  if (first instanceof Promise) {
+    return new Promise((resolve, reject) =>
+      first.then(resolve).catch(err => reject(err))
+    );
+  }
   if (typeof first === "function") {
     const second = first(...args);
-    if (second instanceof Promise)
-      return new Promise(resolve => second.then(() => resolve()));
+    if (second instanceof Promise) {
+      return new Promise((resolve, reject) =>
+        second.then(resolve).catch(err => reject(err))
+      );
+    }
   }
 };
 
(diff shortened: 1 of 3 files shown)


packages/connect/src/create-store.js
```diff
@@ -21,12 +21,18 @@ const getSnapshot = obj => {
 
 const convertToAction = (fn, instance) => (...args) => {
   const first = fn(instance);
-  if (first instanceof Promise)
-    return new Promise(resolve => first.then(() => resolve()));
+  if (first instanceof Promise) {
+    return new Promise((resolve, reject) =>
+      first.then(resolve).catch(err => reject(err))
```
luisherranz: The test is fine. Just don't return anything in the `resolve` here. Actions do not return anything to force people to use state and avoid an antipattern.

packages/connect/src/create-store.js
```diff
@@ -21,12 +21,18 @@ const getSnapshot = obj => {
...
+    );
+  }
   if (typeof first === "function") {
     const second = first(...args);
-    if (second instanceof Promise)
-      return new Promise(resolve => second.then(() => resolve()));
+    if (second instanceof Promise) {
+      return new Promise((resolve, reject) =>
+        second.then(resolve).catch(err => reject(err))
```
luisherranz: Same here, don't do `then(resolve)` but `then(() => resolve)` to avoid returning the value.

luisherranz (changes requested):
The test is fine!

Could you please also add tests for this new case (actions that cause exceptions)?
Thanks :)

michalczaplinski:
@luisherranz Can we merge ?

luisherranz:
I think so but I can't approve it because it's closed. Did you close it on purpose or it was a mistake?

luisherranz (approved):
Anyway, I've reopened to be able to approve it. Feel free to close it again if somehow this is not needed anymore.

michalczaplinski:
> I think so but I can't approve it because it's closed. Did you close it on purpose or it was a mistake?

oh yeah, sorry, I think I just clicked the wrong button! 🤦‍♂

PR #299, opened 2020-02-04
Fix duplicated array elements in the state

### LINKED ISSUE #237: Array elements from package state are duplicated on `csr`

DAreRodz:

This PR adds a `overwriteArrays` option to the `mergePackages` function, allowing it to overwrite arrays in the Frontity state when is set to `true`. Also, that option is used in the client's `createStore` function in order to replace state arrays by those generated during SSR.

Fixes #237

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes!

### DIFF
--- a/.changeset/calm-onions-call.md
+++ b/.changeset/calm-onions-call.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Prevent array elements stored in the state to be duplicated during CSR
--- a/packages/core/src/client/store.ts
+++ b/packages/core/src/client/store.ts
@@ -12,7 +12,7 @@ export default ({
   state: Package["state"];
 }) => {
   state.frontity.platform = "client";
-  const merged = mergePackages({ packages, state });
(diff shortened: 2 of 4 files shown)


luisherranz (approved):
I've been playing a bit with it and it looks like it works fine.

Thanks @DAreRodz! I guess this is the correct way to deal with the array merges during client hydration.

PR #303, opened 2020-02-06
Fix `populate` overwrites

### LINKED ISSUE #245: Second `source.populate` library call in taxonomy handler overrides previous content
**_Bug report_**

The second call to `source.populate` should add the new data without removing the previous one.

When the taxonomy handler is used, two calls are made.
The first one tries to get the category `id` from the `slug`:
https://github.com/frontity/frontity/blob/514b9fa393154c797dd70c7db1a373b61b06ba88/packages/wp-source/src/libraries/handlers/taxonomy.ts#L23-L24
Here is how it looks `state.source.category[8]` after those lines:
```ts
{
  id: 8,
  count: 10791,
  description: 'Category description',
  link: '/category-link/',
  name: 'Category name',
  slug: 'category-slug',
  taxonomy: 'category',
  parent: 0,
  meta: [],
  _links: {
    self: [ [Object] ],
    collection: [ [Object] ],
    about: [ [Object] ],
    'wp:post_type': [ [Object] ],
    curies: [ [Object] ]
  }
}
```

Later another call is made to request the posts of that category:
https://github.com/frontity/frontity/blob/514b9fa393154c797dd70c7db1a373b61b06ba88/packages/wp-source/src/libraries/handlers/taxonomy.ts#L33-L46

After those lines `state.source.category[8]` looks like this:
```ts
{
  id: 8,
  link: '/noticias-motor/',
  name: 'Noticias motor',
  slug: 'noticias-motor',
  taxonomy: 'category',
  _links: {
    self: [ [Object] ],
    collection: [ [Object] ],
    about: [ [Object] ],
    'wp:post_type': [ [Object] ],
    curies: [ [Object] ]
  }
}
```
As you can see some fields have been removed, like `description` or `parent`. Therefore I'm forced to call `categories/?include=8` again in order to access the `description` property.

1. Insert console logs after each `source.populate` call in the taxonomy handler to see the output difference.
2. Visit an archive url for the taxonomy handler to be used.

DAreRodz:

This PR adds a new `force` param to `actions.source.fetch` and `libraries.source.populate`.

The implementation follows what was proposed at https://github.com/frontity/frontity/issues/245#issuecomment-571676896.

📖 Docs changes: [here]( https://app.gitbook.com/@frontity-tech/s/frontity/~/diff/drafts/-M-UJGk-GwBkzPngAfuO/api-reference-1/wordpress-source/@drafts).

Fixes #245.

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes! 
<!-- - 🚧 Work in progress ->


packages/wp-source/src/actions.ts
```diff
@@ -11,19 +11,23 @@ import { ErrorData } from "@frontity/source/types/data";
 import { ServerError } from "@frontity/source";
 
 const actions: WpSource["actions"]["source"] = {
-  fetch: ({ state, libraries }) => async link => {
+  fetch: ({ state, libraries }) => async params => {
     const { source } = state;
 
     const { handlers, redirections } = libraries.source;
 
+    // Get params
```
michalczaplinski: I believe that writing explicit `if...else` is more readable than a ternary in most cases, even if it covers more lines. Something like:
```js
let link;
let force = false;
if (typeof params === "string") {
  link = params;
} else {
  link = params.link;
  force = params.force;
}
```
It also makes it more easily extensible in the future, because you only need to add another `else if {}` if there is another case to handle.

Your code is fine, of course, you **don't** need to change anything. I'm just using this as an opportunity to share an opinion 🙂

packages/wp-source/src/libraries/populate.ts
```diff
@@ -60,29 +61,42 @@ const populate: WpSource["libraries"]["source"]["populate"] = async ({
           isFetching: false
         });
 
+      let entityMap: any;
+      let entityKey: string | number;
+
```
michalczaplinski, line 66: I like this. More readable 👍

michalczaplinski (approved):
Looks good to me! Great job! 👍

michalczaplinski:
@DAreRodz I slightly updated the docs as well, but I think they're good to merge too.

luisherranz (changes requested):
Yeah, the code is fine 👍 

But before we merge, I think we need to make a decision about Frontity APIs in general. **What we decide here will be used from now on** so this is important.

## Option 1: "Required parameters separated or mixed with options object"

This API is what @DAreRodz implemented here. You can use the method by either using the required parameters only or switching to an obejct in the first parameter that contains everything. 
```ts
type fetch = (linkOrOptions: string | { link: string, force: boolean }) => void;
actions.source.fetch("/url");
actions.source.fetch({ link: "/url", force: true });

type set = (linkOrOptions: string | { link: string, method?: "push" | "replace" }) => void;
actions.router.set("/url");
actions.router.set({ link: "/url", method: "replace" });

type sendPageview = (linkOrOptions: string | { link: string, title: string, namespaces?: string[] }, title?: string) => void;
actions.analytics.sendPageview("/url", "My Page");
actions.analytics.sendPageview({ link: "/url", title: "My Page", namespaces: ["googleAnalytics", "mixpanel"] });

type sendEvent = (eventOrOptions: string | { event: string, payload?: object, namespaces: string[] }, payload?: object) => void;
actions.analytics.sendEvent("event name", { eventProp1: "Some payload for the event" });
actions.analytics.sendEvent({ event: "event name", payload: { eventProp1: "Some payload for the event" }, namespaces: ["..."] });
```

## Option 2: "Optional options object after required parameters"

You can use the method normally with the required parameters, and then you can add an options object as the final parameter:
```ts
type fetch = (link: string, options?: { force?: boolean }) => void;
actions.source.fetch("/url");
actions.source.fetch("/url", { force: true });

type set = (link: string, options?: { method?: "push" | "replace" }) => void;
actions.router.set("/url");
actions.router.set("/url", { method: "replace" });

type sendPageview = (link: string, title: string, options?: { namespaces?: string[] }) => void;
actions.analytics.sendPageview("/url", "My Page");
actions.analytics.sendPageview("/url", "My Page", { namespaces: ["googleAnalytics", "mixpanel"] });

type sendEvent = (event: string, payload?: object, options?: { namespaces?: string[]) => void;
actions.analytics.sendEvent("event name", { eventProp1: "Some payload for the event" });
actions.analytics.sendEvent("event name", { eventProp1: "Some payload for the event" }, { namespaces: ["..."] });
```

I've added a "change request" to this PR for the decision.

My choice is Option 2 but let's wait for opinions until next Monday. If nobody raises any red flag, we will continue with Option 2.

michalczaplinski:
I prefer Option 2 because it's a more clean and intuitive API.

I actually thought that it was going to be implemented like that, but I just assumed that there was some reason for *not* doing it.

That said, my only concern with Option 2 is that it's somewhat less flexible. If we define a function signature like:
```
const fetch = (url, {someOption } => {}
```
And then we want to add a new parameter like: 
```
const fetch = (url, newParam, { someOption }) => {}
```

Then, we cannot do it without breaking backwards compatibility, because the `2nd` param will always **have** to be the options object. But if we go with Option 1, we can add a new `2nd` param without a problem.

However, I think that this is **very** unlikely to happen in practice and if it does it will probably mean that we need a new API anyway 😅

DAreRodz:
Hey, I prefer option 2 as well! Also, the implementation would be simpler (there wouldn't be any need to check if the arguments are a string or an object or whatever). 🤔

The reason I chose option 1 to implement this was that the [typescript definition](https://github.com/frontity/frontity/blob/dev/packages/types/src/action.ts) for `Action` restricts it to have just one argument, and I thought It had to be that way for some reason. But now, checking the code where actions are "converted" berfore adding them in the store ([here](https://github.com/frontity/frontity/blob/dev/packages/connect/src/create-store.js#L22)) I saw that my assumption was not true. We need to fix that types.

My vote is for **Option 2**. ✋

luisherranz:
Ok, let's go ahead with Option 2 then :)

> That said, my only concern with Option 2 is that it's somewhat less flexible.

You're right.

Regarding this, in my opinion, when the intention of the API is not clear from the beginning we shall choose an object-based API instead of fixed parameters:

```js
const fetch = ({ url }, { ...options }) => {}
``` 

But for some APIs like `actions.source.fetch`, `actions.router.set` or `state.source.get` we decided to use a primitive string for the sake of simplicity.

> The reason I chose option 1 to implement this was that the typescript definition for Action restricts it to have just one argument

You're absolutely right. Could you please add support for multiple arguments? In this branch there is a working example but if you need help let me know:
https://github.com/frontity/frontity/blob/connect-middleware/packages/connect/types.ts#L178-L282

DAreRodz:
Yeah, haha, but I found an ugly problem now. With the actual definition of `Action` optional arguments cannot be added... 🙁

I hope to find a solution quikly.

DAreRodz:
This is the problem I'm facing right now. When you want to define an action with optional arguments, like `sendEvent`,

```typescript
sendEvent(
  name: string,
  payload?: object,
  options?: { namespaces?: string[] }
): void
```

the `Action` interface should differentiate between two types of arguments: required and optional arguments. How? I don't know really. There's not an easy way to do that, but I can think of two options:

**Option 1:** using arrays (this is not backward-compatible):

```typescript
type Action<Package, RequiredArgs = [], OptionalArgs = []> = {};
```

In the case of `sendEvent`, `RequiredArgs` and `OptionalArgs`
would be something like this:

```typescript
type SendEvent = Action<
  GoogleAnalytics,
  [string],
  [object, { namespaces: string[] }]
>;
```

**Option 2:** mark arguments as _optional_ using the Union operator (`|`) with `undefined` (which would be backward-compatible, but I don't know if it's even possible to implement).

```typescript
type SendEvent = Action<
  GoogleAnalytics,
  string,
  object | undefined,
  { namespaces: string[] } | undefined
>;
```

For both options, the resulting type should be like this (I didn't figure out how to do that):

```typescript
type SendEvent =
  // Required arguments
  ((...req: [string]) => void) &
  // Optional arguments
  ((...args: [string, object]) => void) &
  ((...args: [string, object, { namespaces?: string[] }]) => void);
```

Even with the returned type above, you will still need to define the types for the arguments when you are implementing the function (they won't be inferred automatically).

```typescript
const sendEvent: SendEvent = (
  name: string,
  payload?: object,
  options?: { namespaces?: string[] }
) => {
  name;
  payload;
  options;
};
```

Then, this would work.

```typescript
sendEvent("click");
sendEvent("click", { element: "menu" });
sendEvent("click", { element: "menu" }, { namespaces: ["googleAnalytics"] });
```

---

**Option 3 (extra):** pass a function to the `Action` type. Just as simple as

```typescript
type SendEvent = Action<
  GoogleAnalytics,
  (name: string, payload?: object, options?: { namespaces?: string[] }) => void
>;
```

In that case you wouldn't need to define again the type of the arguments:

```typescript
const sendEvent: SendEvent = (name, payload, options) => {
  name;
  payload;
  options;
};
```

and the implementation would be straightforward (it wouldn't be backward-compatible, though).

---

I wouldn't make `Action` type unnecessarily complex, by the way, so I would choose the **option 3**. 😅

@luisherranz, @michalczaplinski, I want to read your opinions on this problem!

PS: I'm going to mark the issue #245 as blocked while we decide what to do here.

michalczaplinski:
> **Option 1:** using arrays (this is not backward-compatible):

I don't quite understand. How is it not backwards-compatible? Not backwards compatible with what? 

> 
> ```ts
> type Action<Package, RequiredArgs = [], OptionalArgs = []> = {};
> ```
> 
> In the case of `sendEvent`, `RequiredArgs` and `OptionalArgs`
> would be something like this:
> 
> ```ts
> type SendEvent = Action<
>   GoogleAnalytics,
>   [string],
>   [object, { namespaces: string[] }]
> >;
> ```

Yeah, this makes sense.

> I wouldn't make `Action` type unnecessarily complex, by the way, so I would choose the **option 3**. 😅
> and the implementation would be straightforward (it wouldn't be backward-compatible, though).

I'm not sure what you mean with backwards compatibility again. We just have to adjust the types that we use internally, correct? This does not impact the API I think.

I think the Option 3. makes the most sense here, but let me try to clarify here. Ideally, you would like to be able to do something like:

```ts
type SomeActionParams = {
  requiredParam: string;
  optionalParam?: {
    x: number;
  };
}

const someAction: Action<
   WpSource,
   SomeActionParams
>;
```

But because of TS limitation, it is not possible to extract the types from `SomeActionParams` inside of the `Action`. Is that correct?

DAreRodz:
> I don't quite understand. How is it not backwards-compatible? Not backwards compatible with what?

By backward-compatibility I mean that `Action` type, and others that `frontity/types` exports, are not only for internal stuff, they're intended to be used by any package developer that are using TypeScript, so it would affetct the API in some way.

But actually you are right, that's not a problem, we can create a new type for this case instead of modifying `Action`.

> But because of TS limitation, it is not possible to extract the types from SomeActionParams inside of the Action. Is that correct?

Maybe it is possible, but I've not figured out yet how to do such thing.

I also tried to do something like:

```typescript
actions: {
  analytics: {
    sendEvent: 
      | Action<Analytics, string>
      | Action<Analytics, string, object>
      | Action<Analytics, string, object, { namespaces: string[] }>;
  };
};
```

Whe you defined the function implementation the types are correctly inferred, but not when you try to use the action, it complains about the number of arguments you passed. 😕

DAreRodz:
@luisherranz, @michalczaplinski, I think I solved it! 🎉

It was a problem in `ResolveActions` when inferring types for the arguments of an action. Now, the last approach I've wrote (using the union operator) is working. There's no need to change `Action` or write new types for this.

DAreRodz:
It's not fully working yet. I fixed `ResolveActions` in `types/src/utils.ts` but not `ResolveActions` in `connect/src/index.d.ts`.

luisherranz:
@DAreRodz let's jump to a call to discuss this in detail.

DAreRodz:
Good news, it's working!

Both `ResolveActions` types (one in `@frontity/connect` an the other in `@frontity/types`) are fixed. The way to define actions with optional arguments will be like this:

```typescript
interface WpSource extends Source {
  actions: {
    source: {
      fetch:
        | AsyncAction<WpSource, string>
        | AsyncAction<WpSource, string, { force?: boolean }>;
      init: Action<WpSource>;
    };
  };
}
```
Now, types are automatically inferred when implementing or calling actions. Great. 🙌

luisherranz (approved):
I like the syntax for the actions with optional parameters :clap:

The one thing I'm not sure about this is if `{ force: true }` should belong to the `source` definition, instead of only to the `wp-source` one but let's merge it as it is and we will see if it is better to move it in the future.

PS: I have added typescript tests for the new actions.

PR #302, opened 2020-02-06
Expose function for unescaping content from the WP REST API and use it in our themes

luisherranz:
#### Description of what you did:

- [x] Update the implementation of `decode`
- [x] Expose the function in the `frontity` package
- [x] Use it in the 2020 theme
- [ ] ~Use it in the mars theme~ not necessary as far as I can tell.
- [x] Add documentation
#### My PR is a:

- 🐞 Bug fix
- 🚀 New feature

#### Is the PR ready to be merged?

Ready

### Documentation: 
https://app.gitbook.com/@frontity-tech/s/frontity/~/diff/drafts/-M-QaS9tefj0Wl8oJB58/api-reference-1/frontity/@drafts

PR #305, opened 2020-02-06
Unescape the URLs in the router

### LINKED ISSUE #294: UTM links are modified in Frontity's blog: they show `&amp;` when ampersand is used
**_Bug report_**

If I use an UTM link in a Frontity's blog post which contains an ampersand (`&`), it should keep it as it is and not replace the ampersand with `&amp;`.

Frontity displays `&amp;` (instead of `&`) in the UTM links of our blog posts. Therefore, it's not possible to correctly track them on Google Analytics.

For example,

**In WordPress**, a link has these UTM parameters:

wordpress-tracking-link

**In Frontity** it changes and the same link shows the `&amp`; instead:

frontity-tracking-links

michalczaplinski:

Fixes: #294 

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes!


michalczaplinski:
blocked by #302 because it needs the `decode` function from `"frontity"`

luisherranz:
I've merged https://github.com/frontity/frontity/pull/302. Feel free to merge `dev` in this branch.

Moving this back to to-do 👍

luisherranz (commented):
I'm not sure if `decode` should be added in the `normalize` (`routeToParams`) instead of here:
https://github.com/frontity/frontity/blob/fix-ga-links/packages/wp-source/src/libraries/route-utils.ts#L40-L54

@DAreRodz what do you think?

By the way, why the links can contain an encoded query in the first place? Where does this come from?

luisherranz (changes requested):
Could you please add a couple of unit tests to cover these cases?

michalczaplinski:
I've just talked with Reyes, this does not actually fix the problem. 

The problematic content shows up inside of the `state.source.post[id].content.rendered`.

Screenshot 2020-02-10 at 12 23 39

I think that we would need a "anchor link processor" in order to solve this...? Could you confirm that this is the right approach @luisherranz ?

luisherranz:
I see. Let's discuss this via Skype 👍

luisherranz:
Our conclusions from the call:

- WordPress does encode the links in the REST API output so we have to solve this.
- Apparently, we are already decoding the content in the html2react parser:
https://github.com/frontity/frontity/blob/fix-ga-links/packages/html2react/src/libraries/parse/index.ts#L53

So next thing to investigate why that `decode` call is not fixing the encoded links.

michalczaplinski:
So, I can confirm that we have not been unescaping the URLs in the anchor links. 

Also, I noticed that our `html2react` is a bit rudimental. There are some edge cases that we're not handling, like `for` attribute on `<label>` should be translated to `htmlFor` in JSX. 

React-DOM does much more when mapping between the DOM and react: https://github.com/facebook/react/blob/master/packages/react-dom/src/shared/DOMProperty.js

I will fix whatever inconsistencies I see there as well.

@luisherranz

michalczaplinski:
Also, this will need a couple of new tests, of course.

luisherranz:
> So, I can confirm that we have not been unescaping the URLs in the anchor links.

Oh, that's true, we only `decode` the content, not the attributes (props). 

What is your proposed solution then? `decode` the attributes?

michalczaplinski:
yup, exactly. Right now, I'm only decoding the `href`, but I'm not sure if this should apply to all other attributes by default. Do you think this is sufficient for now @luisherranz ?

luisherranz (changes requested):
> but I'm not sure if this should apply to all other attributes by default

I don't think so because with the current method we are creating a new HTML node for each `decode` call and that is slow.

Maybe in the future we can add an option for a faster "non-full version" that implements something like https://github.com/humanmade/simple-entity-decode for these cases:
```js
decode("string", { full: false });
```

By the way, this is marked as ready to merge but two things are still missing:
- Unit tests
- Changeset

packages/html2react/src/libraries/parse/index.ts
```diff
@@ -27,6 +27,10 @@ const adaptNode: AdaptNode = (himalayaNode, parent) => {
           } else if (key === "style") {
             // Add inline styles to the component with `emotion`.
             props.css = css(value);
+          } else if (key === "href") {
+            props.href = decode(value);
+          } else if (key === "for") {
```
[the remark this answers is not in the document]
michalczaplinski: sorry, the following line should be : `props.htmlFor = value` because of https://reactjs.org/docs/dom-elements.html#htmlfor. I will fix that.

michalczaplinski:
> By the way, this is marked as ready to merge but two things are still missing:
> - Unit tests
> - Changeset

yup, I just wanted to confirm the solution first.

packages/html2react/src/libraries/parse/index.ts
```diff
@@ -22,11 +22,21 @@ const adaptNode: AdaptNode = (himalayaNode, parent) => {
...
+            props.htmlFor = value;
+          } else if (key === "accept-charset") {
+            props["acceptCharset"] = value;
+          } else if (key === "http-equiv") {
+            props["httpEquiv"] = value;
             // Add inline styles to the component with `emotion`.
+          } else if (key === "style") {
             props.css = css(value);
+            // Wordpress returns links with HTML escaped entitites so we have to decode them
+          } else if (key === "href") {
```
DAreRodz: I guess we need the same for `src` or `srcset` for images, iframes, etc, right? 🤔

DAreRodz (dismissed):
Code is great! 🙌

I just saw a couple of things that I think can be fixed before merging.

packages/html2react/src/libraries/parse/index.ts
michalczaplinski: good point! Will add it 👍

luisherranz:
I've been investigating this because I'm a bit worried about the performance of our `decode`.

1. We can do a simple regexp to detect and fix the most common entities. Something similar to what Ryan is doing here: https://github.com/humanmade/simple-entity-decode/blob/master/index.js (or even use his library).
2. We can check again if there is any entity left, using this regexp: `/&[a-zA-Z]+(;|\s)/`
    - This is the complete list of HTML entities: https://html.spec.whatwg.org/entities.json
    - You can see the regexp in action here: https://regex101.com/r/ufVAxt/2
3. If we detect a missing entity that was not fixed with the fast implementation, we apply the slower `he` (server) or the `DOMParser` (client).

Once we do that we can add `decode` to all attributes by default and avoid maintaining a list of attributes that need decoding.

@DAreRodz @michalczaplinski thoughts?

michalczaplinski:
@luisherranz I def like the idea! But I would actually benchmark the performance first. My expectation was that the `DOMParser` should be plenty fast, but this list of all HTML entities looks pretty long so might be wrong 😅

I ll make a quick benchmark.

luisherranz:
I did a benchmark and it is significantly faster: https://twitter.com/luisherranz/status/1165196400507981825 :)

michalczaplinski:
oh, can you share the code someplace? I'd like to compare it with using `DOMParser` as we do now on the client. 

Also, by default, does `decode` run both on the client and the server when "normally" rendering the pages?

michalczaplinski:
yeah, according to my quick and dirty test, `DOMParser` is super slow. At least an order of magnitude slower than `he`. The regex `/&[a-zA-Z]+(;|\s)/` is super fast, about 100x faster than `he` and 1000x faster than `DOMParser`.

Screenshot 2020-02-14 at 15 50 22

Time in miliseconds. 

This is more or less a representative run of: 

```js
i = 0;
time = performance.now();
while (i < 100) {
  // clientDecode(html) / he.decode(html) / regex(html)
  i++;
}
console.log("decode/he/regex", performance.now() - time);
```

michalczaplinski:
@luisherranz The method from https://github.com/humanmade/simple-entity-decode/blob/master/index.js is about 2x faster than `he`:

Screenshot 2020-02-14 at 16 18 51

I've added the optimization like you suggested with some tests.

- Opt out of decoding HTML if it doesn't contain special characters
- Use a basic version of `decode` by default that ony handles most common characters.
- As a last resort, use the full decode

luisherranz:
The test is just an html2react execution with x6 slowdown in the Chrome devtools.

I think the bottleneck of createElement/DomParser is the paint that the browser needs to execute (blue in the picture) but you can test it out if you want 👍

luisherranz:
Awesome!! 😁

luisherranz (changes requested):
The new optimization looks great! 😄 

These two things are missing:
- Remove `className`, `htmlFor`, `acceptCharset` and `httpEquiv` from parse/attributes/html.json.
- Either create a whitelist of attributes that need decoding (like `href`, `src`, `srcset` and so on...) or decode all the attributes by default.

michalczaplinski:
> Either create a whitelist of attributes that need decoding (like href, src, srcset and so on...) or decode all the attributes by default.

So, I decided to just `decode` all the attributes that are strings. I could not find a good reason not to.

luisherranz (approved):
I have improved the changesets because the `frontity` one was referring to the `html2react` change instead of to the performance improvement of `decode`. But other than that evertyhing is fine. Thanks @michalczaplinski!

PR #318, opened 2020-02-20
Implement Switch component

### LINKED ISSUE #317: Add switch component to route between pages
**Feature:** [Switch component](https://community.frontity.org/t/switch-component/881)
**User Story**: Add switch component

[Sprint Card](https://github.com/orgs/frontity/projects/2#card-33357730)

iamuchejude:

Create a `<Switch />` component to render components that match a condition/route.

#### My PR is a:

- 🚀 New feature

#### Feature Request

https://community.frontity.org/t/switch-component/881

#### Docs Changes

https://app.gitbook.com/@frontity-tech/s/frontity/~/drafts/-M11ZtY9S7JLZh9FzRMz/api-reference-1/frontity-components/@drafts


packages/components/switch.tsx
```diff
@@ -0,0 +1,99 @@
...
+const description = (
+  <>
+    Don&apos;t panic! Seems like you encountered an error. If this persists,
+    <a href="https://community.frontity.org"> let us know </a> or try refreshing
+    your browser.
+  </>
+);
+
+// Default 404 Error Component
+const DefaultErrorComponent = ({ state }) => {
```
DAreRodz: Hey, I would remove this `DefaultErrorComponent`, I guess it's better for the `Switch` component to render nothing (return `null`) if there is no default component among its children, right? Just my opinion.

packages/components/switch.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const lastComponentIsDefault =
+    React.isValidElement(lastComponent) && !lastComponent.props.when;
+
+  if (hasInvalidComponent) {
+    warn("Children of <Switch /> component should be a type of ReactNode");
+  }
+
+  // Filter components by the value of the 'when' props or path
+  const filteredComponents = components.filter(
+    component => isValidElement(component) && component.props.when
```
DAreRodz, line 25: I was checking the [Feature card](https://community.frontity.org/t/switch-component/881/8) and I didn't found it but I was sure that the `Switch` component would have to render just one of its children, not an array of them.

So, I would change this by
```tsx
const filteredComponent = components
  .filter(isValidElement)
  .find(({ props }) => props.when);
```

packages/components/switch.tsx
```diff
@@ -0,0 +1,99 @@
...
+  );
+
+  // Render filteredComponents
+  if (filteredComponents.length > 0) {
+    return (
```
DAreRodz: and then, return the `filteredComponent` or the default one if it exists.
```tsx
if (filteredComponent) return filteredComponent;
if (lastComponentIsDefault) return lastComponent;

// Return `null` if none of the children have matched.
return null;
```

Btw, you don't have to wrap the return value in a fragment (`<></>`), it's already a react element.

DAreRodz (dismissed):
Hey Jude, good work! 👍

There are some things that could be changed, I think. Take a look at the review and ask me anything if you need so.

packages/components/iframe.tsx
```diff
@@ -80,7 +80,7 @@ const Iframe: Component = ({
 }) => {
   const lazyAttributes: Attributes = {
     "data-src": src,
-    className: "frontity-lazy-iframe".concat(className ? ` ${className}` : ""),
+    className: "frontity-lazy-iframe ".concat(className ? ` ${className}` : ""),
```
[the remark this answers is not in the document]
luisherranz: Guys, quick question:
- Why this change?
- And why in this PR?

packages/components/__tests__/__snapshots__/switch.tests.tsx.snap
```diff
@@ -0,0 +1,19 @@
...
+<h1>
+  Test Archive Component
+</h1>
+`;
+
+exports[`Switch should return null if no component match 1`] = `
+<h1>
+  Test Blog Component
+</h1>
+`;
```
luisherranz: This is returning the `TestBlog` component instead of `null`.

packages/components/iframe.tsx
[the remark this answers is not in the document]
luisherranz: Why this change?
And why is it in this PR?

luisherranz (changes requested):
Excellent code Jude!

I think one of the tests is not correct, though. Could you please take a look?
Also, I don't quite get the modification of the className of Iframe and Image. May I know more about the reason?

packages/components/__tests__/__snapshots__/switch.tests.tsx.snap
iamuchejude: Fixed

iamuchejude:
> Excellent code Jude!
> 
> I think one of the tests is not correct, though. Could you please take a look?
> Also, I don't quite get the modification of the className of Iframe and Image. May I know more about the reason?

I didn't notice the spaces in both lines. I'm gonna fix them now.

luisherranz (approved):
Thanks @iamuchejude!

I have fixed a typo in the changeset and some typos in the docs but apart from that everything is great.

PR #319, opened 2020-02-20
Fix type declarations import

luisherranz:

A while ago we added a package to add type declarations for files that can be imported in Frontity like fonts or images.

This package was not working properly because it looks like that TypeScript needs at least a JavaScript file import to add the declarations. I have fixed it now.

I have also added `css` files to the list of declarations.

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes!


michalczaplinski (approved):
LGTM, but I don't exactly understand the problem that this is solving. 

Did importing css files not work at all before? Did it only fail in certain circumstances? Why are we now importing the index.js from type-declarations ? :)

Sorry, if this is too many questions 😅 

I think going forward you're welcome to let me know if I should spend some time trying to understand 100% of the PR when reviewing or if it's okay if I just do a sanity check and move on!

luisherranz:
Ask as many questions as you want :)

We are trying to solve a TS error when importing CSS/image/fonts files, but to be honest we don't know exactly how to solve it so we are trying things.

iamuchejude:
Does this PR fix the error yet?

PR #321, opened 2020-02-24
fixes `popstate` handler when state is null for `#element`-like links

orballo:

I added an `if` condition in the `popstate` handler in `tiny-router` for the case where `event.state` is `null`. This happens when a link with a url like `#element` is clicked. The solution prevents `frontity` from getting in the middle of the default browser behavior.

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes!


DAreRodz:
The code looks good to me! Do you think it would be possible to add some tests to check that it's working as expected now?

luisherranz:
I'm going to create the e2e project for tiny-router 👍

luisherranz:
@iamuchejude could you please help @orballo here and add an e2e project for `tiny-router` in this branch?

Maybe you can also add some basic tests to make sure `actions.router.set` works by using a button to change the routing. Something like this:
```js
const Root = connect(({ state, actions }) => (
  <>
    <button onClick={() =>{ actions.router.set("/"); } }>Home</button>
    <button onClick={() =>{ actions.router.set("/a-page/"); } }>Page with trailing slash</button>
    <button onClick={() =>{ actions.router.set("/a-page"); } }>Page without trailing slash</button>

    {state.router.link === "/" && <div data-test-id="content">The home</div>}
    {state.router.link === "/a-page/" && <div data-test-id="content">One page</div>}
 </>
);
```

What do you think?

iamuchejude:
Yea, I think that will do it.

luisherranz:
I have updated the package-lock because for some reason the install was failing.

@orballo, I think @iamuchejude has added all the necessary files and folder for e2e testing of the tiny-router package. Is that right @iamuchejude? Can you confirm it?

If so, you can go ahead and add:
- Your code here: https://github.com/frontity/frontity/blob/fix-tiny-router-hashtag-links/e2e/packages/tiny-router/src/index.tsx
- Your tests here: https://github.com/frontity/frontity/blob/fix-tiny-router-hashtag-links/e2e/integration/tiny-router.spec.js

Then, go to the e2e folder, run `npm i` and `npm run e2e:open` to start cypress (or `npm run e2e:run` to run it in headless/ci mode).

iamuchejude:
I’m done with the test @luisherranz

luisherranz:
Awesome, thanks Jude.

luisherranz (dismissed):
@orballo could you please add one or more e2e tests to check that this is working as expected?

orballo:
@luisherranz I'll try to do this after my trip, as I've never worked with e2e tests and it will take some time.

luisherranz:
No problem. If you need help let us know 👍

orballo:
@luisherranz Could you look into this issue? Should I finish it before making any further changes in `tiny-router` or just ignore it and implement the solution again?

luisherranz:
Let me take a look.

luisherranz:
Could you please explain what is the problem you are trying to solve with this change?

orballo:
Fixes a bug when trying to use `#` links. Because the `isPopState` is being set when there is not a new state, the router then fails to set any new links (or at least the next one, I don't remember right now).

luisherranz:
I have been doing some tests and this solution doesn't seem to work if you navigate between pages or use `Link` (`actions.router.set`).

It works if:
- If you are on `/page-a` and use `<a href="/page-a#some-id">`.

It doesn't work if:
- If you are on `/page-b` and use `<a href="/page-a#some-id">`.
- If you are on `/page-a` and use `<Link link="/page-a#some-id">`.
- If you are on `/page-b` and use `<Link link="/page-a#some-id">`.

Can you please confirm that or maybe I am doing something wrong?

luisherranz:
I'm going to think if this can be solved with my proposal of Router Converters: https://community.frontity.org/t/router-converters/1514

luisherranz:
Ok, it looks like Router Converters are not relevant for that.

To make it work with client-side rendering, we need Source converters, which are the same concept but applied to source: The idea is that `"/some-post"` and `/some-post/#some-id` point to the same `source.data` object.

I was already thinking about proposing this for source v2 but they can be added earlier.

What I didn't notice is that for client-side rendering we need to:
- Wait until `data.isReady` so the element with the #id is present on the screen.
- Do the scroll programmatically ourselves.

We could do it with a `useEffect` hook which could be added by `@frontity/tiny-router` by default. Something like this:

```js
const ScrollToHash = ({ state }) => {
  React.useEffect(() => {
    if (hasHash(state.router.link) && data.isReady) {
      const id = getHash(state.router.link);
      window.scrollTo(getElementById(id));
    }
  }, [data.isReady, state.router.link]);
};
```

It's going to be slightly more complex because we should avoid doing the scroll on the first render (the browser will do it because of the SSR) and be careful to wait until `data.isReady`, but I think it's doable. Also, I don't remember right now the exact API to scroll to an element but you get the idea.

----

I'm ok to merge this now to make it work with for now:
- If you are on `/page-a` and use `<a href="/page-a#some-id">`.

@orballo would you mind adding a small e2e test for that case before we merge this?

I will open the Feature Discussions for the Source Converters and the `ScrollToHash` component.

orballo:
Alright 👍

luisherranz:
@orballo I see you're still pushing changes. Let me know once this is ready for review again.

orballo:
@luisherranz it's ready. I was fixing the missing dependency.

luisherranz (approved):
Thanks @orballo! Looking great now.

I'll merge it once the tests pass.

PR #327, opened 2020-02-28
Add state to processors

### LINKED ISSUE #326: Access state in html2react processor
**Feature:** [Access state inside a processor](https://community.frontity.org/t/access-state-inside-a-processor/1294)
**User Story**: Access state in html2react processor

DAreRodz:
Change `Html2React` to allow using `state` and `libraries` inside processors.

Tasks:
- [x] Pass state to processors using the new API (make it backward-compatible).
- [x] Add e2e tests to check processors can use state.
- [x] Update our processors to match the new API (`iframe`, `image`, `script`).
- [x] Rename `process` to `processor`.
- [x] Update documentation with the new API.
- [x] Add changesets.

Extra:
- [x] Check if `Html2React` rerenders based on the state props accessed by processors.

#### Docs draft
https://app.gitbook.com/@frontity-tech/s/frontity/~/diff/drafts/-M1UvPS_1dSH4uORpyk4/api-reference-1/frontity-html2react

#### My PR is a:

- 🚀 New feature

- 🚧 Work in progress


DAreRodz:
The e2e tests for the **ES5** bundle are failing because the variable `process` (the one with `process.env`) is somehow hidden by the `process` function of the processor.

I seems to happen during compilation and minification with processors that use **`css`**.

For example, this processor

```tsx
export const testProcessor: Processor<
  React.HTMLProps<HTMLParagraphElement>,
  Html2ReactTests
> = {
  test: ({ node }) => node.type === "element" && node.component === "p",
  process: ({ node, state }) => {
    if (node.type === "element") {
      node.props.css = css`
        color: ${state.html2reactTests.color};
      `;
    }
    return node;
  }
```

is compiled as

```js
({
  test: function(e) {
    var t = e.node;
    return "element" === t.type && "p" === t.component;
  },
  process: function e(t) {
    var n = t.node,
      r = t.state;
    return (
      "element" === n.type &&
        (n.props.css = Object(u.css)(
          "color:",
          r.html2reactTests.color,
          ";" +
            ("production" === e.env.NODE_ENV
              ? ""
              : "/*# sourceMappingURL=data:application/json;charset=utf-8;base64,{...} */")
        )),
      n
    );
  }
});
```

As you can see in the code above, the `process` attribute ends up being a named function `e(t) { ... }` that is hidding the previous `e` declaration.  That `e` variable was probably `process` since the error occurrs when `"production" === e.env.NODE_ENV` is evaluated:

```
TypeError: Cannot read property 'NODE_ENV' of undefined
```

I've tested to change `process` by a named function...

```tsx
process: function p({ node, state }) { ... }
```

...and the error disappears, but that should not be the way to make it work.

I don't know if we can modify how this code is compiled... It looks like we will have to change `process` name by something different...

What a weird error. 😓

cc: @luisherranz

luisherranz:
Haha... oh my god this is the month of the renamings.

I guess we can rename it to `processor` and keep backward compatibility with `process`. At least for the ones that don't use the `css` prop and work 😆 

I'll take care of it tomorrow morning.

luisherranz:
The code is fixed. I have renamed `process` to `processor` (keeping backward compatibility).

@SantosGuillamot could you please take care of the other two remaining tasks? (docs and changeset)

luisherranz (approved):
Code is great. Thanks @DAreRodz!

SantosGuillamot:
Sure! I'll check them now and let you know.

SantosGuillamot:
I've created a new draft to the documentation. Could you take a look at the changes made @luisherranz ? I wasn't 100% sure about some things. I've added the link to the description of the PR.

luisherranz:
I have tweaked it a little but I think it's good enough 👍

PR #335, opened 2020-03-06
Fix isFetching not being true when data exists

### LINKED ISSUE #334: isFetching is not true for taxonomies
**_Bug report_**

`data.isFetching` should be `true` when you fetch a taxonomy.

`data.isFetching` doesn't change to `true` when you fetch a taxonomy.

Go to the home, click in any taxonomy link. `data` never gets `isFetching === true`.

luisherranz:

We removed that logic in the last version because we moved the `isFetching: true` to a place where it never gets if `data` exists. It was my suggestion actually 😅 

I have added a test to cover this case.

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes!

### DIFF
--- a/.changeset/metal-bats-lick.md
+++ b/.changeset/metal-bats-lick.md
@@ -1,0 +1,5 @@
+---
+"@frontity/wp-source": patch
+---
+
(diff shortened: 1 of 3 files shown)


michalczaplinski (approved):
LGTM 👍

luisherranz:
Thanks @michalczaplinski!

PR #336, opened 2020-03-09
Add deprecation notice to processors

luisherranz:

I have added a deprecation notice when people use `process` instead of `processors`. I didn't want to add a Proxy to `({ node })` just for the shake of the deprecation. As soon as we have a link with migration instructions we can add it to the deprecation notice and people should update their functions to `({ node })` as well.

I also removed the old types because TypeScript errors don't cause Frontity to fail so I guess it's a good way to inform about the new API.

#### My PR is a:

- 🔝 Improvement

- ✅ Yes!


packages/html2react/src/libraries/component.tsx
```diff
@@ -1,20 +1,23 @@
...
 } from "../../types";
 
 const applyProcessors: ApplyProcessors = ({ node, processors, ...payload }) => {
   for (const proc of processors) {
-    const processor =
-      (proc as NewProcessor).processor || (proc as OldProcessor).process;
+    // Add deprecation warning for process.
+    if ((proc as any).process)
+      warn(
+        "The property 'process' has been deprecated. Please rename it to 'processor' instead."
```
michalczaplinski: @luisherranz  Just a small nitpick: I think I would word it a bit differently. Can we add a link to the docs as well? Something like:

`The property 'process' has been deprecated. Please use the new 'processor' property instead. See the documentation https://docs.frontity.org/api-reference-1/frontity-html2react#create-your-own-processors for details about the new API`.

Just to make it clear that the API has changed and not ONLY the name.

luisherranz: I was waiting until we have a migration guide for this but I agree that linking to the new documentation is good enough. I'll do that, thanks @michalczaplinski.

michalczaplinski (approved):
👍

PR #338, opened 2020-03-10
Fix fixed dependencies and remove helmet types

luisherranz:
#### Description of what you did:

I have fixed a problem with e2e installation due to fixed dependencies.

I also take the opportunity to finally remove the types of `react-helmet` in favor of the types of `react-helmet-async`, the package we are currently using.

No changeset is needed because these changes don't affect the implementation of any of the Frontity packages.

#### My PR is a:

- 🐞 Bug fix
- 🔝 Improvement

#### Is the PR ready to be merged?

- ✅ Yes!

PR #339, opened 2020-03-11
Fix backward compatibility problems with last wp-source release

luisherranz:

- I have added a fallback to 0 to make sure they are always a number.
- I have fixed a wrong set of the total pages when there are items and the headers are missing.
- I have fixed a backward compatibility problem with our internal handlers.

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes!


luisherranz:
Ready for review but merge blocked by https://github.com/frontity/frontity/pull/338.

luisherranz:
Blocked until I review the logic when taxonomies exist but don't have any post.

luisherranz:
After reviewing the logic with @DAreRodz it looks like the logic was matching WordPress behavior, which is a bit weird because it returns 0 pages for taxonomies without items, instead of 1 page. So I guess it's fine to match what WordPress does... 🤷‍♂️

Ready for review!

michalczaplinski (approved):
LGTM 👍

.changeset/early-goats-appear.md
```diff
@@ -0,0 +1,7 @@
+---
+"@frontity/wp-source": patch
+---
+
+- Fix wrong default totals when headers are missing. It was set to 0, but it has to be 1.
```
michalczaplinski: Is this correct? Because in the code, I can see that the fallback is still `0`.

michalczaplinski (approved):
I just have the one comment. Not sure if it's legitimate or not. If not then approved 👍

.changeset/early-goats-appear.md
luisherranz: Good catch, thanks!

PR #349, opened 2020-03-23
Add bin property to `create-frontity`

luisherranz:

I forgot to add a `"bin"` property to the `package.json` of `create-frontity` so it doesn't work. This PR fixes that.

#### My PR is a:

- 🐞 Bug fix

- ✅ Yes!

### DIFF
--- a/.changeset/wise-dolls-judge.md
+++ b/.changeset/wise-dolls-judge.md
@@ -1,0 +1,5 @@
+---
+"create-frontity": patch
+---
+
+Add a `"bin"` property to the `package.json` of `create-frontity`.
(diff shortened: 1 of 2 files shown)


SantosGuillamot (approved):
Thanks Luis! I've just realised that the name of the package is `create-frontity` instead of `@frontity/create-frontity`. Is that correct?

luisherranz:
> I've just realised that the name of the package is create-frontity instead of @frontity/create-frontity. Is that correct?

Yes, that's the required name of the package to make it work with `npm init frontity`.

More info on the https://docs.npmjs.com/cli/init docs.

Thanks for the approval! :)

PR #351, opened 2020-03-24
Use an options object on the `error` function

luisherranz:

Switch to an options object for the second parameter of the `error` function to keep consistency with the rest of the Frontity APIs.

Now, if users don't want to throw, they have to use:

```js
import { error } from "frontity";

error("My message", { throw: false });
```

This function was not documented yet, so we can safely do the change.

#### My PR is a:

- 🔝 Improvement

- ✅ Yes!

### DIFF
--- a/packages/error/index.ts
+++ b/packages/error/index.ts
@@ -1,6 +1,16 @@
 const suffix = "\nVisit https://community.frontity.org for help! 🙂\n";
 
-export const error = (message: string, doThrow = true) => {
+interface FrontityError {
+  (
+    message: string,
+    options?: {
+      throw?: boolean;
+    }
+  ): void;
+}
+
+export const error: FrontityError = (message, options = {}) => {
+  const doThrow = typeof options.throw !== "undefined" ? options.throw : true;
   if (process.env.NODE_ENV !== "production") {
     if (doThrow) throw new Error(message + suffix);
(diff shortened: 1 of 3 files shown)


iamuchejude (approved):
Looks good.

luisherranz:
Thanks @iamuchejude 🙂

packages/error/index.ts
```diff
@@ -1,6 +1,16 @@
...
+      throw?: boolean;
+    }
+  ): void;
+}
+
+export const error: FrontityError = (message, options = {}) => {
+  const doThrow = typeof options.throw !== "undefined" ? options.throw : true;
   if (process.env.NODE_ENV !== "production") {
     if (doThrow) throw new Error(message + suffix);
     console.error(message + suffix);
```
michalczaplinski, line 16: @luisherranz Perhaps there is a reason I don't see but why are we throwing an error and also `console.error()`ing the message?

luisherranz: `throw` is like `return`, the function won't run further than that so no console.erro :)

michalczaplinski: sorry I forgot how to javascript for a moment... 🤦‍♂

luisherranz: haha no problem

PR #357, opened 2020-03-27
Update minimist to ^1.2.2 to solve vulnerability

luisherranz:
It looks like GitHub was not able to create a PR for this so I had to do it manually.

This is the vulnerability: https://snyk.io/test/npm/minimist/1.2.0

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -60,7 +60,7 @@
     "koa-route": "^3.2.0",
     "koa-static": "^5.0.0",
     "lodash-es": "^4.17.11",
-    "minimist": "^1.2.0",
+    "minimist": "^1.2.5",
     "open": "^6.3.0",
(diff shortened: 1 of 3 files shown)


packages/core/package.json
```diff
@@ -60,7 +60,7 @@
     "koa-route": "^3.2.0",
     "koa-static": "^5.0.0",
     "lodash-es": "^4.17.11",
-    "minimist": "^1.2.0",
+    "minimist": "^1.2.5",
```
DAreRodz, line 63: Shouldn't `package-lock.json` have been changed after updating this package?

luisherranz: You're absolutely right.

DAreRodz (approved):
Looks good to me. :+1:

PR #304, opened 2020-02-06
Add error component showing how to handle 404 and other errors

michalczaplinski:
#### Description of what you did:

fixes #298 

#### My PR is a:

- 🔝 Improvement

#### Is the PR ready to be merged?

- ✅ Yes!

PR #358, opened 2020-03-28
Fix twentytwenty-theme 404s

SantosGuillamot:
**What**:
This is a PR to solve 404 pages in the twentytwenty-theme.

**Why**:
It seems that the 404 pages in the twentytwenty-theme are returning an `Internal Server Error` right now -> https://twentytwenty.frontity.org/error

**How**:
I've seen that we are not using `connect` inside the `page-error.js`, and adding it seems to solve the problem -> https://github.com/frontity/frontity/blob/dev/packages/twentytwenty-theme/src/components/page-error.js#L37

### DIFF
--- a/.changeset/twenty-squids-flow.md
+++ b/.changeset/twenty-squids-flow.md
@@ -1,0 +1,5 @@
+---
+"@frontity/twentytwenty-theme": patch
+---
+
+Fix 404 pages of twentytwenty-theme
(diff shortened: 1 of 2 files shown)


luisherranz:
Hey, thanks for the fix @SantosGuillamot.

@darkchris86 also noticed this error and commented about it in https://github.com/frontity/frontity/pull/304#issuecomment-605671657.

luisherranz (approved):
Code looks great, feel free to merge 👍

PR #360, opened 2020-03-31
Get rid of getSnapshot in the store

luisherranz:
**What**:

Move `getSnapshot` to an import instead of a store prop.

**Why**:

When people do this:

```js
const MyComp = ({ state, ...props }) => (
  <OtherComp {...props} />
);
```

they get error warnings about `getSnapshot` being there.

**How**:

Now `getSnapshot` can be imported from `"@frontity/connect"`.

**Tasks**:

- [x] Code
- [ ] ~Documentation~: `getSnapshot` is internal and not documented.
- [x] Unit tests
- [x] End to end tests
- [x] TypeScript
- [ ] ~Starter Themes~: themes don't use `getSnapshot`.
- [ ] ~Community discussions~: `getSnapshot` doesn't appear in any discussion.
- [x] Changeset

PR #361, opened 2020-03-31
Instructions to strikethrough non relevant tasks in PRs

luisherranz:
Add instructions to strikethrough non-relevant tasks, instead of just deleting them. People should also add the reason why this is not relevant.

The reviewer needs to agree on why those tasks are not relevant. Leaving them in the description and adding a reason why they are not relevant gives him/her an opportunity to disagree.

There is an example in this PR: https://github.com/frontity/frontity/pull/360

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -27,8 +27,8 @@ merge of your pull request!
 
 <!-- Have you done all of these things?  -->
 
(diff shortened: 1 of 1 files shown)


iamuchejude (approved):

luisherranz:
Thanks @iamuchejude :)

PR #369, opened 2020-04-03
Make isHome present when isReady is true

### LINKED ISSUE #368: Set isHome at the same time than isReady
**_Bug report_**

**Expected behavior**

`data.isHome` information should be available when `data.isReady` turns true.

**Observed behavior**

Right now, we first set `data.isReady = true` and then, at a later moment we set `data.isHome = true`. This is causing a bug when things expect to have `data.isHome` defined and a double render.

luisherranz:
**What**:

`data.isHome` was not present when `data.isReady` was true. Now it is.

**Why**:

`data.isHome` information should be available when `data.isReady` turns true.

**How**:

I moved the `isHome` set to the moment when `isReady` turns true.

**Tasks**:

- [x] Code
- [ ] ~Documentation~
- [x] Unit tests
- [ ] ~End to end tests~
- [ ] ~TypeScript~
- [ ] ~Starter Themes~
- [ ] ~Community discussions~
- [x] Changeset

### DIFF
--- a/.changeset/chilled-trees-grow.md
+++ b/.changeset/chilled-trees-grow.md
@@ -1,0 +1,5 @@
+---
+"@frontity/wp-source": patch
+---
+
+`data.isHome` is now available when `data.isReady` turns true.
--- a/packages/wp-source/src/__tests__/actions.tests.ts
+++ b/packages/wp-source/src/__tests__/actions.tests.ts
@@ -79,6 +79,52 @@ describe("fetch", () => {
     expect(store.state.source.get("/some/route").isReady).toBe(true);
   });
 
+  test('should set isHome in "/"', (done) => {
+    observe(() => {
+      const data = store.state.source.get("/");
+      if (data.isReady) {
+        expect(data.isHome).toBe(true);
+        done();
+      }
+    });
+    store.actions.source.fetch("/");
+  });
+
+  test('should set isHome in "/page/x"', (done) => {
+    observe(() => {
+      const data = store.state.source.get("/page/123");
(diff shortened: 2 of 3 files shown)


luisherranz:
@DAreRodz, could this bug cause a problem with `actions.analytics.sendPageview` and send the wrong title if someone is using this code for their `<title>`:

```js
const Title = ({ state }) => {
  if (data.isReady) {
    if (data.isHome) return <title>I am the home</title>;
    return <title>I am not the home</title>;
  }
  return <title>Loading...</title>;
}
```

I guess so, right?

I am asking to know if we need to do a release for this or not.

DAreRodz:
~You are right, because the analytics package is sending pageviews using the title that is rendered inside `<Head>`, and the code you wrote would fail without the bugfix. So yeah, we have to do a release!~

Edit: I didn't understand the problem at first sight, sorry. :sweat_smile: 

The thing is, with the code you mentioned,`data.isHome` would be `false` the first time `data.isReady` was `true`, and just after that it would render again with `data.isReady` and `data.isHome` being both `true`.

If that's the case, the analytics package would send the wrong title, the one that was rendered the first time `data.isReady` was `true` (in this case `I am not the home`).

So yeah, I guess we will have to do a release.

luisherranz:
Ok, thanks @DAreRodz!

@SantosGuillamot we have to plan a release then :)

DAreRodz (approved):
Code and tests look great. :+1:

luisherranz:
Thanks @DAreRodz !

PR #374, opened 2020-04-14
Re-use an existing browser tab with with the same URL when launching `frontity dev`

michalczaplinski:
**What**:

When running `frontity dev`, a new browser tab is reopened every time the command is run. 

Now, `frontity dev` will attempt to re-use the already open browser tab and switch to it instead of opening a new tab, So, if you already have a tab with http://localhost:3000 open in chrome, we just switch to it. This stops the script from opening many tabs with the same location.

The implementation is a slightly modified version of create-react-app implementation https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/openBrowser.js therefore I have retained the original MIT license in the file.

~Additionally, I have pinned the `react-helmet-async` dependency to `1.0.4` because in the latest release they have broken the types in `index.d.ts` and it causes an error on `npm install` in our repository. https://github.com/staylor/react-helmet-async/pull/91~

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Test in macOS
- [x] Test in Ubuntu
- [x] Test in Windows

Unrelated:

- ~Update starter themes~
- ~Update other packages~
- ~Community discussions~
- ~Documentation~
- ~Unit tests~
- ~End to end tests~
- ~TypeScript tests~


michalczaplinski:
I think this is the kind of change that is a great candidate for e2e testing, but I'm happy to hear other ideas on how you'd test it.

luisherranz:
> Additionally, I have pinned the react-helmet-async dependency to 1.0.4 because in the latest release they have broken the types in index.d.ts and it causes an error on npm install in our repository. staylor/react-helmet-async#91

You should not have used a single PR for two different things because now I can't approve the `react-helmet-async` fix right away.

luisherranz:
> I think this is the kind of change that is a great candidate for e2e testing

Sure. What are your ideas?

luisherranz:
Is this ready for review or are you going to add tests first?

michalczaplinski:
> Sure. What are your ideas?

So, I don't think that you can test the actual functionality of this change (that it does not open a new tab but rather switch to the existing tab). Or, at least, it would be a bit absurd to try to test it this way...

Could we not try to run the cypress e2e tests against a `frontity dev` as well as a production bundle? This way, we can at least confirm that `frontity dev` is not broken

Right now, we're only building and running frontity serve: https://github.com/frontity/frontity/blob/4d5bdd42fc3c17c4602b616469b9344e9541b225/.github/workflows/e2e-tests.yml#L34

which is calling: 

https://github.com/frontity/frontity/blob/107d3543ce5463186809b7e6f50ca31ffbdc107d/package.json#L7-L9

luisherranz:
Sure, good idea. Maybe we can even add a write to a file to make sure HMR is working.

Should we add a `frontity dev` e2e test to the testing roadmap? It doesn't feel like it belongs here, right?

luisherranz:
Have you tested this in Ubuntu and Windows?

michalczaplinski:
> Have you tested this in Ubuntu and Windows?
I haven't because I only have MacOS and no space for a VM on my machine : (
Do you know a cloud VM service that I could use or are you able to test it
yourself?

> Should we add a frontity dev e2e test to the testing roadmap? It doesn't feel like it belongs here, right?

I'm not sure. I think we could just add an extra test to launch the e2e test with `frontity dev` just to make sure "it works". There is already an npm script for it: 

https://github.com/frontity/frontity/blob/1c68016be047f890c0ae48a7dd4a5b507e3609bc/e2e/package.json#L9

Can I just do it? :)

luisherranz:
> Do you know a cloud VM service that I could use or are you able to test it
yourself?

Maybe with https://crossbrowsertesting.com/ or https://www.browserstack.com/?

> Can I just do it? :)

Sure.

packages/core/src/scripts/utils/open-browser.ts
```diff
@@ -0,0 +1,150 @@
...
+ * Reads the BROWSER environment variable and decides what to do with it. Returns
+ * true if it opened a browser or ran a node.js script, otherwise false.
+ */
+export function openBrowserTab(url: string) {
+  const { action, value, args } = getBrowserEnv();
+  switch (action) {
+    case Actions.NONE:
+      // Special case: BROWSER="none" will prevent opening completely.
+      return false;
+    case Actions.SCRIPT:
```
luisherranz, line 142: What is this for?

luisherranz (changes requested):
To be consistent between args and ENV variables we need to change the browser arg of the `dev` command too: https://github.com/frontity/frontity/blob/dev/packages/core/src/scripts/dev.ts#L21

Also, to be consistent with the work of https://github.com/frontity/frontity/pull/262 you should rename `BROWSER` to `FRONTITY_BROWSER`.

packages/core/src/scripts/utils/open-browser.ts
michalczaplinski: @luisherranz  I don't exactly know as this was copied 1:1 from create-react-app. 

By just looking at the code, it seems that if the value of the `BROWSER` environment variable is a `.js` script, it executes this script instead of launching a browser. I'm not sure if this is a workaround for some edge case, etc. so I would just keep this code as is.

michalczaplinski:
> To be consistent between args and ENV variables we need to change the browser arg of the `dev` command too: https://github.com/frontity/frontity/blob/dev/packages/core/src/scripts/dev.ts#L21

I'm not sure what you mean, or perhaps it's a misunderstanding. That specific option (`openBrowser`) refers to whether a browser window should be opened or not. The `BROWSER` env variable holds the path to the default browser, so those are 2 different things if I understand you correctly.

> Also, to be consistent with the work of #262 you should rename `BROWSER` to `FRONTITY_BROWSER`.

The `BROWSER` variable is a system-wide setting across different operating systems (and especially on linux) and not a Frontity-specific option, so I think we should just keep it as `BROWSER`.

michalczaplinski:
Also, should I wait until I add the `frontity dev` e2e test or can we merge it beforehand?

luisherranz:
> The BROWSER env variable holds the path to the default browser, so those are 2 different things if I understand you correctly.

Looking at the code I see that you can use `BROWSER=none` which is basically `--openBrowser=false` so they overlap and it should be only one or it is going to be confusing.

If we go with `BROWSER` instead of `--openBrowser` we need to remove `--openBrowser` and create a `--browser` that matches `BROWSER`. 

> The BROWSER variable is a system-wide setting across different operating systems (and especially on linux) and not a Frontity-specific option, so I think we should just keep it as BROWSER.

Ok. For consistency with the rest of our ENV variables, what about accepting both `BROWSER` and `FRONTITY_BROWSER`? Does it make sense?

luisherranz:
> should I wait until I add the frontity dev e2e test

I don't think it is related so whatever you want.

packages/core/src/scripts/utils/open-browser.ts
```diff
@@ -0,0 +1,150 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file: https://github.com/facebook/create-react-app/blob/a4fa63fcc1fb97fa50778b7c1a73a01da3a3e022/LICENSE
+ */
+
+import chalk from "chalk";
+// const chalk = require("chalk");
```
luisherranz: I guess this comment can be deleted.

packages/core/src/scripts/utils/open-browser.ts
luisherranz: I wouldn't like to add code to the framework without understanding it so please check it out before we merge.

Also, if this covers some use cases, it needs to go to the documentation.

packages/core/src/scripts/utils/open-browser.ts
michalczaplinski: yup, done 🙂

packages/core/src/scripts/utils/open-browser.ts
michalczaplinski: Looks like this is just an advanced feature, in case you want to run a js script that eventually launches a browser in order to do some initialization, etc. There are some more details here: https://create-react-app.dev/docs/advanced-configuration/

Maybe this could be useful for automated testing with cypress? 🤔

michalczaplinski:
> Looking at the code I see that you can use BROWSER=none which is basically --openBrowser=false so they overlap and it should be only one or it is going to be confusing.

> If we go with BROWSER instead of --openBrowser we need to remove --openBrowser and create a --browser that matches BROWSER.

Right, I see what you mean. I had a look and the current CLI flag is actually `--dont-open-browser` 😄 

https://github.com/frontity/frontity/blob/107d3543ce5463186809b7e6f50ca31ffbdc107d/packages/frontity/src/cli/index.ts#L70

So, I think it's fine if we just keep the BROWSER env variable as is and then the matrix of options is:

1. `BROWSER=none`, whatever other flags passed - we dont open the browser
2. `BROWSER=something`, `--dont-open-browser` - also dont open browser
3. `BROWSER=something`, no CLI flags - open the browser
4. no env variables or CLI flags  - open the browser

> Ok. For consistency with the rest of our ENV variables, what about accepting both BROWSER and FRONTITY_BROWSER? Does it make sense?

With respect to that I don't really have a strong opinion, but one argument in favour of just using the BROWSER variable is that some users might be already familiar with it. I don't see how anybody would be getting confused even if this one variable is not consistent in naming with the rest of variables in frontity 🙂. I think accepting 2 environmental variables that serve the same purpose would be more confusing, in my opinion!

michalczaplinski:
@luisherranz Could you please check if this works on windows when running `frontity dev` ? 🙏 

I've signed up for browserstack but as far as I can see I only use a browser to look at websites and I cannot use a full windows environment, unfortunately. I've tested it on MacOS and works as expected. On windows it should fall back to just opening a new tab.

Let me know if there's anything else stopping this from merging 🙂

luisherranz:
I've finally managed to run our monorepo in Windows and try this PR, but it's not working: it opens a new tab each time you run `npx frontity dev`.

I don't know if I'm missing something. I'm off for today but we can take a look together next week.

luisherranz:
@DAreRodz can you test it in Ubuntu?

DAreRodz:
> @DAreRodz can you test it in Ubuntu?

I've just tested it and the `dev` command always opens a new tab. :confused:

Maybe it is because the `create-react-app` solution only works for OSX? Each time I run `npm start` in a `create-react-app` project also opens a new tab.

DAreRodz:
Well, it is. Looking at the code, there is only one specific case when the tab can be reused, and is this one:

https://github.com/frontity/frontity/blob/35966b4d6ad2cd5afb29ebad5521e03612f09c1f/packages/core/src/scripts/utils/open-browser.ts#L66-L69

The rest of the covered cases are these (and use [`open`](https://www.npmjs.com/package/open) as we currently do):

https://github.com/frontity/frontity/blob/35966b4d6ad2cd5afb29ebad5521e03612f09c1f/packages/core/src/scripts/utils/open-browser.ts#L108-L129

luisherranz:
I see, thanks @DAreRodz. @michalczaplinski can you confirm that? Is this supposed to only work on OSX?

michalczaplinski:
> I see, thanks @DAreRodz. @michalczaplinski can you confirm that? Is this supposed to only work on OSX?

That's correct! And I'm sorry -that should have been more clear from the beginning. 

By "works on windows & linux" I meant that it doesn't break anything, but the actual behaviour of preserving the tab only works on MacOS.

packages/core/package.json
```diff
@@ -49,6 +50,8 @@
     "babel-plugin-transform-inline-environment-variables": "^0.4.3",
     "babel-polyfill": "^6.26.0",
     "core-js": "^3.0.1",
+    "copyfiles": "^2.2.0",
```
luisherranz: This needs to be in `devDependencies`.

.changeset/rare-mayflies-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Try to re-use an open browser tab when running `frontity dev`
```
luisherranz: Could you please add that this is only for OSX?

luisherranz (changes requested):

luisherranz (approved):
Thanks @michalczaplinski. I'm really sorry it took me so long to review this. Let's merge it now 🙂

PR #380, opened 2020-04-22
Add `options` argument to `actions.router.set()`

orballo:
**What**:

Implements `options` argument for `actions.router.set()`, and adds `method` and `state` fields to `state.router` according to [this discussion](https://community.frontity.org/t/router-set-options-method-title-and-state/1508).

**Why**:

Because we need to rule the world with infinite scroll.

**How**:

With the help of tea, sushi, and a bit of wine.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] TypeScript tests
- [x] Community discussions
- [x] Unit tests
- [x] End to end tests
- [x] Documentation (https://github.com/frontity/gitbook-docs/pull/29)
- [x] Changeset 

Unrelated:

- [ ] ~Update starter themes~
- [ ] ~Update other packages~


orballo:
@luisherranz @DAreRodz please have a look at this before I finish wrapping everything up :)

packages/router/types.ts
```diff
@@ -5,13 +5,22 @@ interface Router<T = null> extends Package {
     frontity?: Package["state"]["frontity"];
     router: {
       link: string;
+      method: SetOptions["method"];
```
luisherranz: Let's remove `method` from the state.

packages/router/types.ts
```diff
@@ -5,13 +5,22 @@ interface Router<T = null> extends Package {
...
+        | Action<T extends null ? Router : T, string>
+        | Action<T extends null ? Router : T, string, SetOptions>;
     };
   };
 }
 
 export default Router;
+
+export interface SetOptions {
+  method?: "push" | "replace" | "pop";
```
luisherranz: I would prefer not to expose `"pop"` to the users because they will see it in TS autocompletion and it is meant to be used internally only.

I didn't like the `isPopState` flag either but... I don't know. Maybe doing manual casting of that type internally? I can't think of a nice way to do it right now, honestly.

packages/tiny-router/src/__tests__/index.test.ts
```diff
@@ -87,30 +172,31 @@ describe("actions", () => {
...
-      // check that first state is correct
-      expect(window.history.state).toMatchObject({ link: "/" });
+      const link = "/about-us/";
 
-      // check reactions to "popstate" events
-      let currentLink = store.state.router.link;
-      observe(() => {
-        currentLink = store.state.router.link;
+      Object.defineProperty(window, "location", {
+        value: { pathname: link, search: "", hash: "" },
```
luisherranz: I would add values to search and hash here, to have a broader test.

luisherranz (changes requested):
Awesome work @orballo 👏 

I think we can get rid of `state.router.method` and I did a couple of comments more but apart from that, the code is ready for merge.

A couple of things missing:
- Fix unit tests.
- An e2e test for replace.
- A changeset.

I have edited the OP tasks.

If you need help with those let us know.

----

We also need to update the docs but I'll ask the DevRel team about it.

luisherranz (approved):
Fantastic. Thanks @orballo!

luisherranz:
I have added a PR to change our documentation: https://github.com/frontity/gitbook-docs/pull/29

@orballo could you please review it as well? Thanks!

orballo (commented):
Nice. I tried to write something checking the length of `window.history` but for some reason it is always 50. I didn't think about this approach.

PR #381, opened 2020-04-23
docs: prefer to place README.md for a new project

hideokamoto:
**What**:

Put `README.md` files into starter template packages.

**Why**:

When we create a new project, we can easy to understand how to develop and build it.
But now, there is no README.md file in our created project.

**How**:

Add a simple `README.md` file into these starter templates.

**Tasks**:

- [ ] Code
- [ ] TypeScript

- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests

- [x] Update starter themes
- [ ] Update other packages

- [ ] Documentation
- [ ] Community discussions

- [ ] Changeset

### DIFF
--- a/packages/mars-theme/README.md
+++ b/packages/mars-theme/README.md
@@ -1,0 +1,27 @@
+# Marth Theme
+
+## Getting started
+
+```bash
+$ npm install
+$ npm run dev
+
+SERVER STARTED -- Listening @ http://localhost:3000
+  - mode: development
+  - target: module
+
+```
+
+## Build
+
+```bash
+$ npm run build
+```
+
+
+## Deployment (Vercel (now.sh))
+
+```bash
+$ npx now login
+$ npx now --prod
+```
--- a/packages/twentytwenty-theme/README.md
+++ b/packages/twentytwenty-theme/README.md
@@ -1,0 +1,27 @@
+# TwentyTwenty Theme
+
+## Getting started
+
+```bash
+$ npm install
+$ npm run dev
+
+SERVER STARTED -- Listening @ http://localhost:3000
+  - mode: development
+  - target: module
+
+```
+
+## Build
+
+```bash
+$ npm run build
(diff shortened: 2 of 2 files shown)


michalczaplinski:
Thanks for the PR @hideokamoto ! 

You're absolutely right that README file are useful for understanding the project 🙂 

However, in this case, the README files are going to be present in the `/packages/mars-theme/README.md` and `/packages/twentytwenty-theme/README.md` even before a user creates any projects. In this case, I feel that it might be a bit confusing because the READMEs would suggest that the user can for example run `npm run build` inside `packages/mars-theme` in the main repository which is something that they should not do! 

But, what we could do is to create the README.md files programatically when the `frontity create` command runs! 

It would require a little bit of coding, but if you're up for that challange @hideokamoto, you could start by having a look [this function](https://github.com/frontity/frontity/blob/107d3543ce5463186809b7e6f50ca31ffbdc107d/packages/frontity/src/steps/index.ts#L69) which is used to programatically create a `package.json` file when you run `frontity create`. You could use the same approach for creating the README files and start by adding a new function like `createReadmeFile`.

Let me know if you have more questions and I ll be happy to guide you 🙂

luisherranz:
Hmm.. yeah. Thanks @hideokamoto. I agree that it would be nice to provide a `README.md` with instructions but as @michalczaplinski said, it's better that it ends up at the root level:

```bash
/my-frontity-project
  /packages/...
  frontity.settings.js
  README.md
```

@hideokamoto if you are willing to help us with this let me know and I can create a small guide to show you where you have to add the logic to add this file.

hideokamoto:
Thanks for the feedback!
I've tried to create a new Pull Request to create the README.md by script.
https://github.com/frontity/frontity/pull/391

Thanks.

michalczaplinski:
Closing this in favour of #391 .

PR #387, opened 2020-04-24
Core: fix yarn symlink bug

### LINKED ISSUE #364: Yarn isn't working [8pt]
**_Bug report_**

**Expected behavior**
Users expect to be able to use `yarn` instead of `npm`.

**Observed behavior**
`yarn` is not working because it doesn’t install local dependencies using `symlinks`. It copies the folder instead.

**Possible solution**
We can use `symlink-dir` at `frontity dev`.

iamuchejude:
**What**:

Yarn install local dependencies using symlinks. It copies the folder instead. See #364 

**How**:

Apply the [`symlink`](https://www.npmjs.com/package/symlink-dir) to folders when you run `frontity dev`

**Tasks**:

- [x] Code
- ~[ ] TypeScript~

- [x] Unit tests
- ~[ ] End to end tests~
- ~[ ] TypeScript tests~

- ~[ ] Update starter themes~
- ~[ ] Update other packages~

- ~[ ] Documentation~
- ~[ ] Community discussions~

- [x] Changeset

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -61,6 +61,7 @@
     "koa-static": "^5.0.0",
     "lodash-es": "^4.17.11",
     "minimist": "^1.2.5",
+    "mock-fs": "^4.12.0",
     "open": "^6.3.0",
     "promise-monofill": "^1.0.1",
     "raw-loader": "^3.0.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -73,7 +73,8 @@
     "webpack": "^4.29.6",
     "webpack-bundle-analyzer": "^3.1.0",
     "webpack-dev-middleware": "^3.7.2",
-    "webpack-hot-middleware": "^2.24.3"
+    "webpack-hot-middleware": "^2.24.3",
+    "symlink-dir": "3.1.2"
   },
   "devDependencies": {
     "@frontity/types": "^1.2.0",
--- a/packages/core/src/scripts/utils/__tests__/create-symlinks.test.ts
+++ b/packages/core/src/scripts/utils/__tests__/create-symlinks.test.ts
@@ -70,12 +70,6 @@ describe("Create Symlink", () => {
       }`,
     });
 
-    try {
-      await createSymlinks();
-    } catch (e) {
-      expect(true).toBeTruthy();
-    }
-
     await expect(createSymlinks()).rejects.toThrow(
       `${projectDir}/packages/${packageName} for ${packageName} does not exist.`
     );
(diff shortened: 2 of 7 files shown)


packages/core/src/scripts/utils/create-symlinks.ts
```diff
@@ -26,7 +30,7 @@ export default async () => {
     dependencyNames.map(async (name) => {
       const packageDir = dependencies[name].replace(/^(file:)/, "");
 
-      const dir = await resolve(process.cwd(), packageDir);
+      const dir = resolve(process.env.CWD, packageDir);
```
michalczaplinski, line 33: @iamuchejude I've changed this ⬆️ 

Because I set the `process.env.CWD` in the test I can "mock" that we are running the command in a fake project directory (which is really just a temp dir). But make sure that it still works when you test manually 😀

iamuchejude: Oh cool!

packages/core/src/scripts/utils/__tests__/create-symlinks.test.ts
```diff
@@ -0,0 +1,107 @@
...
+          }
+        }`,
+      [`${projectDir}/packages/dep-with-file`]: {
+        "package.json": `{ name: "dep-with-file" }`,
+      },
+    });
+
+    await createSymlinks();
+
+    const stats = await lstatPromise("./node_modules/dep-with-file");
```
michalczaplinski: There is also [`lstatSync`](https://nodejs.org/api/fs.html#fs_fs_lstatsync_path_options) by the way, that I think would be better to use instead I realized, but it's not too important 🙂

packages/core/src/scripts/utils/__tests__/create-symlinks.test.ts
```diff
@@ -0,0 +1,107 @@
...
+          "${packageName}": "file:./packages/${packageName}"
+        }
+      }`,
+    });
+
+    try {
+      await createSymlinks();
+    } catch (e) {
+      expect(true).toBeTruthy();
+    }
```
michalczaplinski: Why do we need this? 🙂

packages/core/src/scripts/utils/__tests__/create-symlinks.test.ts
michalczaplinski: You don't have to change it, just a comment :)

packages/core/src/scripts/utils/__tests__/create-symlinks.test.ts
iamuchejude: Oh, I was supposed to get rid of this.

michalczaplinski (approved):

packages/core/package.json
```diff
@@ -61,6 +61,7 @@
     "koa-static": "^5.0.0",
     "lodash-es": "^4.17.11",
     "minimist": "^1.2.5",
+    "mock-fs": "^4.12.0",
```
luisherranz, line 64: I think this should be on `devDependencies`, right?

PR #389, opened 2020-04-24
Correct the behavior of actions.source.fetch with { force: true }

### LINKED ISSUE #340: actions.source.fetch with "force: true" doesn't work as expected [3pt]

If I use `actions.source.fetch("/some-link", { force: true })`:
- The data object property `isReady` is not reset to `false` because it is still ready even though we are fetching again to see if there are any updates in the content.
- The data object custom properties (like `items`, `isCategory`...) are not removed.
- The entities I get from the new fetch overwrite the old ones.

If I use `actions.source.fetch("/some-link", { force: true })`:
- The data object property `isReady` resets to `false`.
- The data object custom properties (like `items`, `isCategory`...) are removed.
- The entities I get from the new fetch don't overwrite the old ones.

**Possible solution**

Don't reset `data` when it exists, even if `force` is true (https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/actions.ts#L34-L41).

Modify all the handlers so they pass down `force` to `populate`.

### LINKED ISSUE #365: actions.source.fetch should populate link, route, page and query even if data exists [3pt]
**_Bug report_**

**Expected behavior**

When calling `actions.source.fetch`, it should populate data with `link`, `route`, `page` and `query`, even if data already exists.

**Observed behavior**

Right now, if data exists, none of those values are populated and a Frontity theme expecting them fails.

I have added a failing test in the  `populate-link-route-page-and-query-even-if-data-exists` branch: https://github.com/frontity/frontity/blob/populate-link-route-page-and-query-even-if-data-exists/packages/wp-source/src/__tests__/actions.tests.ts#L82-L98

**Possible solution**

Change the logic of `fetch` at this point: https://github.com/frontity/frontity/blob/populate-link-route-page-and-query-even-if-data-exists/packages/wp-source/src/actions.ts#L32 so it doesn't matter if data exists or not.

Use the https://github.com/frontity/frontity/tree/populate-link-route-page-and-query-even-if-data-exists branch.

michalczaplinski:
**What**:

Fix the bugs in wp-source:

- If I use `actions.source.fetch("/some-link", { force: true })`:
  - The data object property `isReady` should never be reset to `false`. This causes inconsistent UI glitches because the data that is already present will disappear.
  - The data object custom properties (like `items`, `isCategory`...) should not be removed from the data.
  - The entities I get from the new fetch should overwrite the old ones.
- When calling actions.source.fetch, it should populate data with link, route, page and query, even if data already exists. 

Details in #365, #340 

**How**:
1. Update the logic of `actions.ts` to handle the additional cases
2. Pass the `force` parameter that is passed to `fetch()` down to each handler type.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests
- ~~End to end tests~~
- ~~TypeScript tests~~
- ~~Update starter themes~~
- ~~Update other packages~~
- ~~Documentation~~
- ~~Community discussions~~


packages/wp-source/src/libraries/__tests__/populate.tests.ts
```diff
@@ -83,12 +84,11 @@ describe("populate", () => {
       state,
       response: mockResponse({
         id: 1,
-        count: 5,
         link: "https://test.frontity.io/category/cat-1/",
         slug: "cat-1",
         taxonomy: "category",
+        count: 5,
         description: "This is the Category 1",
-        parent: 0,
```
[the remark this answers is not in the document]
luisherranz: Be aware that actual responses from the REST API contain `parent` and `count`.

packages/wp-source/src/libraries/handlers/__tests__/author.tests.ts
```diff
@@ -69,6 +69,40 @@ describe("author", () => {
...
+        {
+          "X-WP-Total": "5",
+          "X-WP-TotalPages": "2",
+        }
+      )
+    );
+
+    expect((store.state.source.data["/author/author-1/"] as any).name).toEqual(
+      "Author 1"
+    );
```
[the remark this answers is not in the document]
DAreRodz: Uhm, I'll take a look.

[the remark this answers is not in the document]
DAreRodz: I found the problem. It's failing because `name` is not included in the `data` object but in the `author` object, so in order to get the name of the author you have to do

```ts
state.source.author[authorId].name
```

Remember that the author id is a number, not the link.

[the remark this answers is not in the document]
michalczaplinski: Why is the `name` not included in the `data` though? Is this by design?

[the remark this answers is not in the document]
DAreRodz: If I remember correctly, the idea was to have in `data` only the props needed to get the entities from the state, along with other properties that are not found in entities (like which items belong to that specific category page, the number of pages, and so on).

For getting posts, pages, etc. you would need `type` and `id`.

For getting authors you just only need `id` so, yeah. I would say this is by design. :smile:

packages/wp-source/src/libraries/__tests__/populate.tests.ts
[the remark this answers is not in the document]
michalczaplinski: It's irrelevant for this test case so I think this was just a copy-paste.

packages/wp-source/src/libraries/__tests__/populate.tests.ts
```diff
@@ -74,6 +74,7 @@ describe("populate", () => {
           link: "https://test.frontity.io/category/cat-1/",
           slug: "cat-1",
           taxonomy: "category",
+          count: 1,
```
michalczaplinski, line 77: I've added the `count` here so that we can test if this property was actually updated and not just added to the state.

michalczaplinski:
@DAreRodz When you're reviewing this, let me know if you have more (additional) ideas on how to test this 🙂

packages/wp-source/src/actions.ts
```diff
@@ -29,20 +29,34 @@ const actions: WpSource["actions"]["source"] = {
...
-        query,
-        page,
         isFetching: true,
         isReady: false,
       };
-    } else if (data.isReady || data.isFetching || data.isError) {
+    } else if (force) {
+      // If we fetch with `{ force: true }`, then wipe out whatever
+      // other data was present before and just set those flags:
+      source.data[link] = {
```
luisherranz: This is wrong, we shouldn't wipe out the data, only set `data.isFetching` to true. Please read the issue in detail: https://github.com/frontity/frontity/issues/340

luisherranz (changes requested):

packages/wp-source/src/actions.ts
michalczaplinski: This was my thought initially as well, but then I realized that if I implement this part like:

```
 source.data[link] = {
        ...source.data[link],
        isFetching: true,
        isReady: true,
      } as any;
```

Then we run into the following problem: If the previous `fetch()` has returned an error, the data for the error would not be removed. We would have to add additional logic to remove the `isError`, `errorStatusText` and `errorStatus`. 

This snapshot failure shows it:
```
  ● fetch › should run again when `force` is used

    expect(received).toMatchSnapshot()

    Snapshot name: `fetch should run again when \`force\` is used 1`

    - Snapshot
    + Received

    @@ -1,8 +1,11 @@
      Object {
        "/some/route/": Object {
    +     "errorStatus": 408,
    +     "errorStatusText": "Request Timeout",
          "id": 1,
    +     "isError": true,
          "isFetching": false,
          "isPostType": true,
          "isReady": true,
          "link": "/some/route/",
          "page": 1,

      138 |     await store.actions.source.fetch("/some/route/", { force: true });
      139 |     expect(handler.func).toHaveBeenCalled();
    > 140 |     expect(store.state.source.data).toMatchSnapshot();
          |                                     ^
      141 |   });
```

Why would it be incorrect to clear all the existing data if we run with `{force: true}`?

luisherranz: > Why would it be incorrect to clear all the existing data if we run with {force: true}?

Because this is just a refresh of the content. The page needs to remain as it is.

Imagine you are on your twitter feed and click a "Refresh" button. You don't want your current feed to disappear, you just want that when the "Refresh" fetching ends, you see the new tweets appear in your feed.

packages/wp-source/src/actions.ts
```diff
@@ -29,20 +29,37 @@ const actions: WpSource["actions"]["source"] = {
...
-    } else if (data.isReady || data.isFetching || data.isError) {
+    } else if (force) {
+      // If we fetch with `{ force: true }`, then only set `isFetching` to true again
+      data.isFetching = true;
+
+      // This is a workaround in case that `data` has previously included an error
+      if (data.isError) {
+        delete data.isError;
+        delete data.errorStatus;
+        delete data.errorStatusText;
```
luisherranz: This is going to break the themes, because data is going to be ready (`data.isReady === true`) but with no information about what it is. You should reset the data object completely. Also, could you please add a test of that case?

packages/wp-source/src/libraries/handlers/__tests__/author.tests.ts
```diff
@@ -69,6 +69,44 @@ describe("author", () => {
     ]);
   });
 
+  test("overwrites the data when fetched with { force: true }", async () => {
```
luisherranz, line 72: Shouldn't be a test like this in all the other handlers?

luisherranz (changes requested):

packages/wp-source/src/actions.ts
michalczaplinski: I'm sorry but I don't really understand. 

> You should reset the data object completely

My understanding was that we are not supposed to reset the data object but just remove the error as we discussed?

> This is going to break the themes, because data is going to be ready (data.isReady === true) but with no information about what it is. 

Could you elaborate?

> Also, could you please add a test of that case?

done.

packages/wp-source/src/libraries/handlers/__tests__/author.tests.ts
michalczaplinski: My thinking was that it's clear that the same behaviour copies over to the other handlers because we are using `Object.assign()` everywhere. 

But let me know if you still think we need them and I ll add them 🙂

luisherranz: That's not the way to think about this. If someone refactors this code in the future, they will see the failing test in `author`, they will fix it, they'll see all tests passing and they can assume everything is fine in the rest of the handlers, they will ask for a review and unless someone is really paying attention to that small detail in the review, thing that I really doubt it, we would have introduced 3 bugs. So please add a test for each place where we need to ensure that this behavior is not changed over time.

- https://github.com/frontity/frontity/blob/8f764ad5431a70d53ebded630fa9021b4b2b547e/packages/wp-source/src/libraries/handlers/date.ts#L48
- https://github.com/frontity/frontity/blob/8f764ad5431a70d53ebded630fa9021b4b2b547e/packages/wp-source/src/libraries/handlers/postType.ts#L31
- https://github.com/frontity/frontity/blob/8f764ad5431a70d53ebded630fa9021b4b2b547e/packages/wp-source/src/libraries/handlers/postTypeArchive.ts#L43

packages/wp-source/src/actions.ts
luisherranz: > My understanding was that we are not supposed to reset the data object but just remove the error as we discussed?

But in the case of the error, you can break themes and packages because the data is going to be this!

```js
{
  isReady: true;
  isFetching: true;
}
```

So please, in case of a previous error, reset the data completely back to `isReady: false, isFetching: true`.

luisherranz (dismissed):

packages/wp-source/src/libraries/handlers/__tests__/author.tests.ts
michalczaplinski: You're right about this 😓

packages/wp-source/src/libraries/handlers/__tests__/date.tests.ts
```diff
@@ -120,4 +120,34 @@ describe("date", () => {
...
+
+    // Fetch the data again (this time returning `date2019PostsPage2`)
+    await store.actions.source.fetch("/2019/", {
+      force: true,
+    });
+
+    // Make sure that api.get() was called for each `source.fetch()`
+    expect(api.get).toHaveBeenCalledTimes(2);
+
+    expect(store.state.source).toMatchSnapshot();
```
michalczaplinski, line 151: I'm not 100% if this is the correct approach to take.

After the first fetch the state will contain the first 3 posts with id's `1`, `2` & `3`. Then, after the second fetch I fetch the two posts with id `4` & `5`. 

I'm not sure if the `state.source.post` should only contain the posts `4` & `5` or all the posts from 1 to 5.

To make matters more confusing `state.source.post` contains posts from 1 to 5: https://github.com/frontity/frontity/blob/3770212f44f701f77f521c2ce98009f98d4abd1e/packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/date.tests.ts.snap#L1265-L1300

but `state.source.data` only contains posts 4 and 5:
https://github.com/frontity/frontity/blob/3770212f44f701f77f521c2ce98009f98d4abd1e/packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/date.tests.ts.snap#L1321-L1338

DAreRodz: It's better to not remove the content once it was fetched. You can overwrite it, but not remove it.

For example, imagine that you navigate to the 2019 archive (`/2019/`), click on the post 1 (`/post-1/attachment-1/`) and while you are reading the content then frontity _updates_ the 2019 archive (e.g. with a link which executes `actions.source.fetch` on hover). The content you are reading would disappear!

Also, even if you haven't visited that post yet, when you visit it Frontity will avoid a WP REST API request because that post is already in the state.

So yeah, it's completely fine to have that data in `state.source.post` and `state.source.data`. :smile:

michalczaplinski: > So yeah, it's completely fine to have that data in state.source.post and state.source.data. 😄

ok, so you're saying that even if I fetch with `force: true` this situation is okay? 

Namely, that after the second fetch (with `force:true`) I have only posts 4 & 5 in `state.source.data` BUT posts 1,2,3,4 & 5 in `state.source.post`

packages/wp-source/src/__tests__/actions.tests.ts
```diff
@@ -275,4 +275,104 @@ describe("init", () => {
...
+    expect((store.state.source.data["/some/route/"] as any).items).toEqual([]);
+
+    await store.actions.source.fetch("/some/route/", {
+      force: true,
+    });
+
+    const data = store.state.source.get("/some/route/");
+
+    // NOTE!!! This should fail in wp-source 2.0, because `isCategory` and `items` should be removed
+    expect(data).toMatchSnapshot();
```
DAreRodz, line 349: I guess this tests are prepared to fail when we develop the next version of wp-source, right?

DAreRodz (approved):
Everything is fine. Great job @michalczaplinski!

packages/wp-source/src/__tests__/actions.tests.ts
michalczaplinski: That's the idea 🙂

michalczaplinski:
@luisherranz Do you also want to have a look before we merge this?

luisherranz (commented):
> @luisherranz Do you also want to have a look before we merge this?

No problem, go ahead with the merge 🙂

michalczaplinski:
I had to update the snapshots for the new tests added in this PR with the new values for API endpoint `frontity.io` -> `frontity.org` due to 5e01b119165fe09636719ec2c99952131dee6bac

Those are the only 4 lines that were changed:
https://github.com/frontity/frontity/pull/389/commits/328e3e295728cd25cb6b34ca6b0cf077a7caf058

PR #404, opened 2020-04-28
Implement `useConnect` hook

orballo:
**What**:

- Implements the `useConnect` hook to consume `state`, `actions` and `libraries` in components without passing them as props.
- Adds an options parameter to `connect()` that looks like: `{ injectProps: bool; }`.
- If `{ injectProps: false }` option is passed to `connect()`, the `state`, `actions` and `libraries` won't be passed as props to the component.

**Why**:

To allow the use of the store in custom hooks to abstract and reuse complex logic.

**How**:

Typing.

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] Community discussions (https://community.frontity.org/t/useconnect-hook/1079)
- [x] TypeScript
- [x] Changeset
- [x] Documentation: https://github.com/frontity/gitbook-docs/pull/61

**Irrelevant**
- ~End to end tests~
- ~TypeScript tests~
- ~Update starter themes~
- ~Update other packages~

### DIFF
--- a/packages/connect/src/connect.js
+++ b/packages/connect/src/connect.js
@@ -30,10 +30,16 @@ const mapStateToStores = (state) => {
     .map(raw);
 };
 
-export function connect(Comp, options = { injectProps: true }) {
+export function connect(Comp, options) {
   const isStatelessComp = !(Comp.prototype && Comp.prototype.isReactComponent);
 
   let ReactiveComp;
+
+  const defaultOptions = {
+    injectProps: true,
+  };
+
+  options = options ? { ...defaultOptions, ...options } : defaultOptions;
 
   if (isStatelessComp && hasHooks) {
     // use a hook based reactive wrapper when we can
--- a/packages/connect/src/connect.js
+++ b/packages/connect/src/connect.js
@@ -163,3 +163,5 @@ export function connect(Comp) {
 
   return ReactiveComp;
 }
+
+export const useConnect = () => useContext(context);
--- a/packages/connect/src/connect.js
+++ b/packages/connect/src/connect.js
@@ -185,13 +185,14 @@ export function connect(Comp, options) {
 }
 
 export const useConnect = () => {
-  if (!isConnected) {
-    warn(
-      "Warning: useConnect() is being used in a non connected component, " +
-        "therefore the component won't update on state changes. " +
-        "Please wrap your component with connect().\n"
-    );
-  }
+  if (!isConnected)
+    useEffect(() => {
+      warn(
+        "Warning: useConnect() is being used in a non connected component, " +
+          "therefore the component won't update on state changes. " +
+          "Please wrap your component with connect().\n"
+      );
+    }, []);
 
   return useContext(context);
 };
--- a/packages/connect/src/connect.js
+++ b/packages/connect/src/connect.js
@@ -186,13 +186,11 @@ export function connect(Comp, options) {
 
 export const useConnect = () => {
   if (!isConnected)
-    useEffect(() => {
-      warn(
-        "Warning: useConnect() is being used in a non connected component, " +
-          "therefore the component won't update on state changes. " +
-          "Please wrap your component with connect().\n"
-      );
-    }, []);
+    warn(
+      "Warning: useConnect() is being used in a non connected component, " +
+        "therefore the component won't update on state changes. " +
+        "Please wrap your component with connect().\n"
+    );
 
   return useContext(context);
 };
--- a/packages/connect/src/connect.js
+++ b/packages/connect/src/connect.js
@@ -30,7 +30,7 @@ const mapStateToStores = (state) => {
     .map(raw);
 };
 
-export function connect(Comp) {
+export function connect(Comp, options = { injectProps: true }) {
   const isStatelessComp = !(Comp.prototype && Comp.prototype.isReactComponent);
 
   let ReactiveComp;
@@ -67,7 +67,7 @@ export function connect(Comp) {
       }, []);
 
       // run the reactive render instead of the original one
-      return render({ ...props, ...frontity });
+      return render({ ...props, ...(options.injectProps ? frontity : {}) });
     });
   } else {
     const BaseComp = isStatelessComp ? Component : Comp;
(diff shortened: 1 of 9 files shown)


orballo:
Please, have a look at this, and let me know of any changes before I do the changeset.

luisherranz (changes requested):
Can you add a TypeScript export in https://github.com/frontity/frontity/blob/dev/packages/connect/src/index.d.ts?

My idea was to do this:

```ts
import MyPackage from "../types";
const Comp = () => {
  const { state, actions, libraries } = useConnect<MyPackage>();
  return <div>...</div>;
};
```

luisherranz:
@DAreRodz, it looks like an unrelated GTM e2e test has failed in this PR. May you please take a look? Thanks!

```
Google Tag Manager

    1) should sent pageviews when going back or forward
```

DAreRodz:
I've tested it in my local environment and it has worked. Maybe the spec file is wrong and there is some kind of race condition?

This is the relevant code:

```ts
beforeEach(() => {
  cy.visit("http://localhost:3001?name=google-tag-manager");
});

it("should sent pageviews when going back or forward", () => {
  cy.get("button#change-link").click();
  cy.go("back");
  cy.window().its("dataLayer").its(2).should("deep.equal", pageviewHome);

  cy.go("forward");
  cy.window().its("dataLayer").its(3).should("deep.equal", pageviewSomePost);
});
```

luisherranz:
I'm rerunning the tests.

DAreRodz:
Tests passed this time... :thinking:

luisherranz:
> Tests passed this time... 🤔

Then we may have a race condition. We'll need to pay attention to this test 👍

luisherranz (approved):
Awesome work @orballo.

@DAreRodz can you please do the documentation PR? You can merge this after.

luisherranz (changes requested):
@orballo I just thought it'd be amazing to have a console warning when you use the hook but forget to add the `connect`. 

Would you mind adding it? You can add a flag in `connect` when the component starts rendering and remove it afterwards. It'd be great to know the name of the component in the warning so maybe that can be the value of the flag.

For the warning you can use `@frontity/error` warn.

orballo:
> @orballo I just thought it'd be amazing to have a console warning when you use the hook but forget to add the `connect`.
> 
> Would you mind adding it? You can add a flag in `connect` when the component starts rendering and remove it afterwards. It'd be great to know the name of the component in the warning so maybe that can be the value of the flag.
> 
> For the warning you can use `@frontity/error` warn.

I honestly have no idea of how to do this. How can the hook be aware of `connect()` wrapping the component or a flag set by it? If you could point me in the right direction, that would be great.

EDIT: Hum, ok. I think I know what you want me to do.
EDIT2: It seems impossible to know the name of the component that has no `connect()` wrapping it precisely because there is no connect.

orballo:
@luisherranz Please, have a look at the changes and let me know if you can think of a way to print the failing component name on the warning. I couldn't come up with anything.

luisherranz:
My bad, we only have access to the name when the component is wrapper in `connect`. It also seems that Function.caller is not available in strict mode: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/caller

Ok, no problem. Everything is fine. Let's wait for the documentation PR to merge this.

luisherranz (approved):

packages/connect/src/connect.js
```diff
@@ -185,13 +185,14 @@ export function connect(Comp, options) {
...
 export const useConnect = () => {
-  if (!isConnected) {
-    warn(
-      "Warning: useConnect() is being used in a non connected component, " +
-        "therefore the component won't update on state changes. " +
-        "Please wrap your component with connect().\n"
-    );
-  }
+  if (!isConnected)
+    useEffect(() => {
```
orballo: @luisherranz I'm not sure why is needed `useEffect` here. Is it a standard practice to use `useEffect` around any kind of side effect?

if we read `isConnected` inside the `useEffect`, it's value might have changed already, though implementing it this way goes against the `hook` rules and might cause bugs in the future when using this hook with other hooks in the same component, I think: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

luisherranz: I did it to avoid showing the warning multiple times for the same component.

I think this is one of those cases where nesting a hook inside an `if` won't be problematic, because the `if` result won't change over the whole execution of the app, but you're right that it's safer to avoid it.

To avoid it, the `if` should be inside the `useEffect`, and if we do that, the `isConnected` won't be true at that point, due to the async nature of `useEffect`.

I guess we could use an internal state to store if `isConnected` was there and then look at it inside the `useEffect`, but it seems overcomplicated just to avoid multiple warnings, which is not a big deal.

I'm going to revert it. Good catch.

orballo: > because the if result won't change over the whole execution of the app, but you're right that it's safer to avoid it.

True. Yeah, I didn't think about that.

DAreRodz: @luisherranz @orballo what about using `useLayoutEffect` instead? It runs synchronously.

orballo: @DAreRodz I just tried that, but it doesn't work :/ I think it is because even if it's run synchronously, it's run when the component mounts, and the flag has already changed.

DAreRodz: :man_shrugging:

luisherranz: Yeah, don't worry about this. It's not a bid deal.

luisherranz:
@DAreRodz are you having problems with the docs or did you forget this?

DAreRodz:
Woops, totally my fault. :see_no_evil:

I guess I forget about this because it's not included in the board. I'll try to keep these tasks somewhere so it doesn't happen again.

DAreRodz:
@michalczaplinski could you review the documentation? Thanks :smile:

michalczaplinski:
@DAreRodz Reviewed the docs and you can merge this as soon as the docs are merged (minor changes requested)

luisherranz:
Hey guys, as soon as there is a PR on the docs repo we can merge this. There's no need to wait for the merge on the other repo.

PR #406, opened 2020-04-28
Avoid embedding fonts in HTML as it impacts LightHouse performance score

luisherranz:
**What**:

It looks like embedding fonts in HTML is not a good technique and the performance score of LightHouse is hurt. I have discovered it after doing some tests because of a conversation that started here in this community post: https://community.frontity.org/t/psi-as-expected/1617

**Why**:

Even though I still don't like the flash that happens when loading custom fonts with `font-display: swap`, I think it's better to start recommending it because of the performance gain in LightHouse.

**How**:

I have lowered the limit for embedding fonts to only 8Kbs from the previous 25kbs. I have also refactored the font-face component of the TwentyTwenty theme.

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] Update starter themes
- [x] Changeset
- [x] Community discussions

Unrelated:

- [ ] ~End to end tests~
- [ ] ~TypeScript~
- [ ] ~TypeScript tests~
- [ ] ~Update other packages~
- [ ] ~Documentation~

### DIFF
--- a/.changeset/serious-brooms-exist.md
+++ b/.changeset/serious-brooms-exist.md
@@ -1,0 +1,5 @@
+---
+"@frontity/twentytwenty-theme": patch
+---
+
+Use `swap` in all `font-displays`. This was needed because we are not embedding fonts smaller than 25Kbs anymore.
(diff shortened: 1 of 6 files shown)


SantosGuillamot (approved):
Looks great Luis! Thanks :slightly_smiling_face: 

By the way, I've added an issue to do the same to the `frontity.org` repo - [#98](https://github.com/frontity/frontity.org/issues/98)

PR #408, opened 2020-04-28
Populate link route page and query even if data exists

michalczaplinski:
**What**:

When calling actions.source.fetch, it should populate data with link, route, page and query, even if data already exists.

**Why**:

Right now, if data exists, none of those values are populated and a Frontity theme expecting them fails.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests
- [ ] Changeset

** N/A **
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages
- [ ] Documentation
- [ ] Community discussions


michalczaplinski:
This will be fixed in #389 because the changes in `source.fetch()` logic belong together in one place.

PR #413, opened 2020-05-03
Reimplementation of `useInView` hook

orballo:
**What**:

Reimplementation of `useInView` hook as a wrapper for `react-intersection-observer` that checks if `IntersectionObserver` is supported by the browser.

**Why**:

Currently some browsers, like Safari 11, do not support `IntersectionObserver`.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests
- [x] Commented code (please use TSDoc if possible)
- [x] Documentation: https://github.com/frontity/gitbook-docs/pull/63 
- [x] Update other packages
- [x] Changeset

Not relevant:
- Community discussions
- Update starter themes
- TypeScript tests

This work has been done in a branch coming from the `use-connect` PR branch. My guess is that once that branch is merged in `dev`, those changes won't show up here, but I'm not sure if this is too dirty. I did it this way so I can work in the next branch on the infinite scroll using the previous work done in this branches. Is there a better way?


orballo:
I'm not sure what the default value for `supported` should be on the server. What do you think?

luisherranz:
> I'm not sure what the default value for supported should be on the server. What do you think?

That's a really good question.

Maybe assume that is going to be supported, which is the best-case scenario?

```js
{
  inView: false,
  ref,
  supported: true
}
```

It will trigger a hydration warning when IO doesn't exist but that shouldn't be a problem (and the warning will go away with React Concurrent anyway).

What do you think?

orballo:
I'm starting to believe that React Concurrent are our fathers 😂

But yeah, that's the scenario I'm getting right now. I decided to return `supported: false` only if it's client and `window.IntersectionObserver` doesn't exist, which as you said prints a warning, but the support for IntersectionObserver is 90% so this should be the way to go.

luisherranz (changes requested):
Code looks great, thanks 🙂 

Some tasks/issues:
- We didn't really have a discussion about this API in the community, so can you at least confirm that this is the best approach for things like the `useInfiniteScroll` hook?
- Can you please rebase the branch onto dev? It seems like it was created out of the `useConnect` hook branch, which is not merged yet.
- Can you add a changeset for this, with a major bump?
- Can you add a TSDoc comment for the hook?

luisherranz:
@iamuchejude would you mind adding a documentation PR for this?

We need to create a new page for `@frontity/hooks`, similar to what we have for [`@frontity/components`](https://docs.frontity.org/api-reference-1/frontity-components). 

You can copy the hook options from here: https://github.com/thebuilder/react-intersection-observer#options

Thanks!

iamuchejude:
Okay @luisherranz

orballo:
@luisherranz I'm still working on `useInfiniteScroll` in another branch upon these changes and the changes in `useConnect` so I'd say we wait until that work is done in case more changes are needed before merging this PR.

luisherranz:
> I'd say we wait until that work is done in case more changes are needed before merging this PR.

Perfect!

SantosGuillamot:
>I'd say we wait until that work is done in case more changes are needed before merging this PR.

We'd like to use this hook to do some improvements in frontity.org. Do we still need to wait until the `useInfiniteScroll` hook is ready?

orballo:
@SantosGuillamot Not anymore. I'll work on this today and finish any pending work.

orballo:
@SantosGuillamot this should be ready to go, but I think https://github.com/frontity/frontity/pull/404 should be merged first, because this branch contains work done there, and once `use-connect` is merged, those changes shouldn't appear here.

SantosGuillamot:
Great, thanks @orballo !

DAreRodz:
Hey, there isn't any topic on the community about this. Is it supposed to be one? Do we have to create one before merging the PR? I say it because there is a task for that and it's unchecked.

cc: @luisherranz, @SantosGuillamot

DAreRodz:
By the way, why is not relevant to update other packages? Components in the `@frontity/components` package are using `useInView` from `react-intersection-observer` in their implementations. Shouldn't they use our `useInView` hook?

orballo:
Hummm... I wasn't aware of that. I'll work on that today.

Also, I generated a changeset for a minor version update, but I'm not sure that's right. Are these breaking changes?

.changeset/four-gorillas-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontity/hooks": minor
```
luisherranz: This should be a major change. We're changing the previous implementation. It used to return an array and now it returns an object.

luisherranz (changes requested):
Great use of TSDocs :)

> Are these breaking changes?

Yes, this needs to be a major bump.

I have also updated the tasks.

.changeset/four-gorillas-protect.md
orballo: If I change that manually to `major` works? or do I need to remove and do the changeset again? just curious :P

Also, when does the version get updated? Is the version updated in all the dependent packages?

luisherranz: Yes it does. Magic, right? 😆 

> Also, when does the version get updated? Is the version updated in all the dependent packages?

Not sure, it's our first major bump I think. We'll see.

orballo:
Packages updated. You can proceed however you deem pertinent.

luisherranz (approved):
Perfect. Thanks @orballo 🎉

PR #415, opened 2020-05-04
Migrate Frontity Connect to react easy state [21pt]

### LINKED ISSUE #236: Cursor jumps to end of input fields
This bug was first reported here: https://community.frontity.org/t/cursor-jumping-on-controlled-input/789

**_Bug report_**

While tipying in an `<input>` component, the cursor must remain in place.

The cursor jumps to the end every time the value changes.

Just add an `<input>` to your theme and test it.

luisherranz:
**What**:

Stop using our own fork and migrate back to `react-easy-state`.

**Why**:

Because they are adding hooks to their library, so we can build on top of it. This means that Frontity Connect will become an opinionated version of `react-easy-state` instead of a fork.

**How**:

I removed all our code, added `react-easy-state` back and used the hooks to replicate our changes.

**Tasks**:

- [x] Migrate `createStore`.
- [x] `createStore` unit tests.
- [x] Migrate `connect`.
- [x] `connect` unit tests.
- [x] Add `useConnect` and its tests (from https://github.com/frontity/frontity/pull/404).
- [x] Add `options.inject` to `connect` and it's tests (from https://github.com/frontity/frontity/pull/404).
- [x] Fix bug of `react-easy-state`: https://github.com/RisingStack/react-easy-state/issues/196
- [x] Fork `react-easy-state` and `@nx-js/observer-utils`
- [x] Add TSDocs
- [x] Review TypeScript tests.
- [x] Changeset.

Unrelated:

- [ ] Migrate to TypeScript.
- [ ] Open new FD for the refactor of the TypeScript.
- [ ] Use `batch()` to batch invalidations of `observe` as well.
- [ ] Refactor general TypeScript (can check this https://github.com/frontity/frontity/blob/connect-middleware/packages/connect/types.ts).
- [ ] Run performance check in https://github.com/krausest/js-framework-benchmark
- [ ] ~Update starter themes~
- [ ] ~Update other packages~
- [ ] ~Documentation~
- [ ] ~Community discussions~


luisherranz:
I have updated the code to use the last alpha versions from `react-easy-state` and `@nx-js/observer-util`. I've also merged it with the new `useConnect` hook.

luisherranz:
I've added an extra task to use `batch` to batch invalidations of `observe()` as well, although that can be done in a separate PR.

luisherranz:
Ok, this is not ready for review just yet, but I think the work is pretty much finished. I will take a good look next Monday to make sure I haven't missed anything and I will mark it as ready 🙂

EDIT: I will also record a video for the reviewer.

luisherranz:
This is finally ready for review 🙂

I've made a video to go over the changes: https://www.loom.com/share/e48c9accaabd4b69b8f4cc2349ed997e

packages/connect/src/connect.js
```diff
@@ -143,47 +196,43 @@ export function connect(Comp, options) {
...
-          const stores = mapStateToStores(state);
-          super.deriveStoresFromProps(props, ...stores);
-        }
-        // respect user defined getDerivedStateFromProps
-        if (super.getDerivedStateFromProps) {
-          return super.getDerivedStateFromProps(props, state);
-        }
-        return null;
-      }
-
```
michalczaplinski, line 159: What was the purpose of this custom `getDerivedStateFromProps` and why was it removed now?

I see that enables us to also remove the `mapStateToStores` (at the top of the file) but I don't understand the purpose.

packages/connect/src/connect.js
```diff
@@ -1,140 +1,193 @@
...
-  let ReactiveComp;
-
-  const defaultOptions = {
-    injectProps: true,
-  };
-
   options = options ? { ...defaultOptions, ...options } : defaultOptions;
 
-  if (isStatelessComp && hasHooks) {
```
michalczaplinski, line 47: Why did we remove the `hasHooks` check?

packages/connect/src/connect.js
```diff
@@ -1,140 +1,193 @@
...
-  useContext,
+  useMemo,
+  memo,
   useRef,
+  createContext,
+  useContext,
 } from "react";
 import { warn } from "@frontity/error";
-import { observe, unobserve, raw, isObservable } from ".";
+import { observe, unobserve } from "@frontity/observer-util";
```
michalczaplinski, line 11: I wasn't sure where to comment this but I see that we're not using the `view` function from `@frontity/react-easy-state` and instead we're creating our own wrapper here. 

Would it make sense to delete it altogether from our fork to shave some bytes (let's make @cristianbote happy 😅 )? I don't think that the function gets tree-shaken out of the bundle, right?

michalczaplinski (commented):
I haven't finished reviewing it all but I have some questions in the meantime 🙂

packages/connect/package.json
```diff
@@ -24,7 +24,10 @@
     "test:inspect": "node --inspect ../../node_modules/.bin/jest --watch --no-cache --runInBand"
   },
   "dependencies": {
-    "@frontity/error": "^0.1.3"
+    "@frontity/react-easy-state": "7.0.0",
+    "@frontity/observer-util": "4.3.0",
```
cristianbote: Do you feel strongly about having the versions locked? Since these are gonna be maintained by us, we might be able to use the `^` flag for the versions. Just as we do with the current packages.

.changeset/forty-mayflies-joke.md
```diff
@@ -0,0 +1,7 @@
+---
+"@frontity/connect": minor
+---
+
+Migrate the internals of `@frontity/connect` to use `react-easy-state` instead of a fork.
+
+_This means that in the future, merging bugs from `react-easy-state` will be much easier._
```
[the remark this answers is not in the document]
luisherranz: Hahaha, that's true. Done 👍

packages/connect/src/__tests__/types.tests.tsx
```diff
@@ -1,107 +1,113 @@
 /* eslint-disable @typescript-eslint/no-unused-vars, jest/expect-expect */
```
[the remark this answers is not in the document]
luisherranz: Good catch. Done 👍

packages/connect/src/connect.js
```diff
@@ -1,140 +1,193 @@
...
+ * reactive to its changes.
+ *
+ * @param Comp - The React component to be connected.
+ * @param options - The connect options, defined in {@link defaultOptions}.
+ *
+ * @returns The same Component, but reactive and with the store included in the
+ * props.
+ */
 export function connect(Comp, options) {
   const isStatelessComp = !(Comp.prototype && Comp.prototype.isReactComponent);
```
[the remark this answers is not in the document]
luisherranz: Good idea. Done 👍

packages/connect/src/connect.js
luisherranz: It is a new lifecycle that `react-easy-state` introduced to be able to manipulate stores that are local to a class component. I don't think it is relevant to our use case.

packages/connect/src/connect.js
luisherranz: Because Frontity ships with React 17, so it always has hooks.

In `react-easy-state` that's not always the case because it supports earlier versions of React.

packages/connect/package.json
luisherranz: Ups, good catch. Yeah, you are right. I had them fixed because they were pointing to a beta before, but I forgot to unfix them 👍

packages/connect/src/connect.js
luisherranz: No, I don't think it will get tree shaken because due to the scheduler, `react-easy-state` contains side-effects. 

I'll remove it from our fork 👍

luisherranz: The thing is that if I remove `view` I'll kind of break the library because tests won't pass. I'd have to remove the tests, but I think it'll be easier in the future if we need to do a change to make sure that everything works in `react-easy-state`.

Maybe we could mark the library as tree-shakable, except for the scheduler. 

@cristianbote, I've never dealt with this type of side-effect configuration before. Could you please help me out with that? Is it possible to do it, right?

luisherranz: We can make that in a different PR to avoid blocking this one (it's actually a different repo).

So I'll resolve this for now 🙂

packages/connect/package.json
luisherranz: Done!

luisherranz:
Thanks for the reviews guys. I've made the changes and should be up for a review again 🙂

packages/connect/src/connect.js
michalczaplinski: okay, I see. thanks!

packages/connect/src/connect.js
michalczaplinski: right, makes sense 🙂

packages/connect/src/connect.js
michalczaplinski: yeah, I didn't mean this to be a blocker of course. We can do it in a future PR :)

packages/connect/src/connect.js
```diff
@@ -1,140 +1,193 @@
...
+        () => {
+          reaction.current = observe(Comp, {
+            scheduler: () => {
+              // Trigger a new rerender if the component has already been
+              // mounted.
+              if (reaction.current.mounted) setState({});
+              // Annotate it as "changed" if the component has not been mounted
+              // yet.
+              else reaction.current.changedBeforeMounted = true;
+            },
```
michalczaplinski, line 80: Note to self: The custom scheduler is used here in order to fix the bug: https://github.com/RisingStack/react-easy-state/issues/196#issuecomment-766367636

michalczaplinski (approved):
OK, looks great @luisherranz 🙂 Let's merge this!

luisherranz:
Merged! Thanks a lot guys! 🙂

PR #420, opened 2020-05-05
Add publicPath option to CLI

### LINKED ISSUE #393: Add publicPath option to CLI and pass it down to Webpack in both the build and dev commands [3pt]
**Feature:** [Change publicPath](https://community.frontity.org/t/change-publicpath/1461).
**User Story**: Change the publicPath at build time.

[**Sprint Card**](https://github.com/orgs/frontity/projects/4#card-36476088)

DAreRodz:
**What**:

Add a `--publicPath` option to `build` and `dev` commands that is passed to webpack builds.

Closes #393 

**Why**:

To allow frontity users to change the public path option used by Webpack: https://webpack.js.org/guides/public-path/

**How**:

Following the implementation proposal: https://community.frontity.org/t/change-publicpath/1461/5

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] Documentation: https://github.com/frontity/gitbook-docs/pull/60
- [x] Community discussions: https://community.frontity.org/t/change-publicpath/1461/7?u=david
- [x] Changeset

**Unrelated tasks**:
- ~TypeScript~ No new types created.
- ~End to end tests~ There are no e2e tests for the CLI yet.
- ~TypeScript tests~ No new types created.
- ~Update starter themes~ Not needed.
- ~Update other packages~ Not needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
     "test": "lerna run test:ci --stream",
     "test:ci": "jest --clearCache && lerna run test:ci --stream",
     "e2e:prod:all": "cd e2e/project/ && npx frontity build && npx frontity serve --port 3001",
-    "e2e:prod:module": "cd e2e/project/ && npx frontity build --target module && npx frontity serve --port 3001",
-    "e2e:prod:es5": "cd e2e/project/ && npx frontity build --target es5 && npx frontity serve --port 3001",
+    "e2e:prod:module": "cd e2e/project/ && npx frontity build --publicPath /custom/path --target module && npx frontity serve --port 3001",
+    "e2e:prod:es5": "cd e2e/project/ && npx frontity build --publicPath /custom/path/ --target es5 && npx frontity serve --port 3001",
     "prepare": "lerna bootstrap --hoist",
     "reinstall": "lerna clean --yes && npm run prepare"
   },
--- a/packages/core/src/server/index.tsx
+++ b/packages/core/src/server/index.tsx
@@ -27,7 +27,12 @@ export default ({ packages }): ReturnType<Koa["callback"]> => {
   const app = new Koa();
 
   // Serve static files.
-  app.use(mount("/static", serve("./build/static")));
+  app.use(async (ctx, next) => {
+    const moduleStats = await getStats({ target: "module" });
+    const es5Stats = await getStats({ target: "es5" });
+    const stats = moduleStats || es5Stats;
+    if (stats) mount(stats.publicPath, serve("build/static"))(ctx, next);
+  });
 
   // Default robots.txt.
   app.use(
--- a/packages/core/src/server/index.tsx
+++ b/packages/core/src/server/index.tsx
@@ -31,7 +31,8 @@ export default ({ packages }): ReturnType<Koa["callback"]> => {
     const moduleStats = await getStats({ target: "module" });
     const es5Stats = await getStats({ target: "es5" });
     const stats = moduleStats || es5Stats;
-    if (stats) mount(stats.publicPath, serve("build/static"))(ctx, next);
+    const publicPath = stats ? stats.publicPath : "/static";
+    return mount(publicPath, serve("build/static"))(ctx, next);
   });
 
   // Default robots.txt.
--- a/.changeset/odd-students-look.md
+++ b/.changeset/odd-students-look.md
@@ -1,0 +1,6 @@
+---
+"@frontity/core": minor
+"frontity": minor
+---
+
+Add a `--publicPath` option to `build` and `dev` commands and pass it to webpack builds.
--- a/packages/core/src/config/index.ts
+++ b/packages/core/src/config/index.ts
@@ -8,13 +8,21 @@ import getFrontity from "./frontity";
 export default ({
   mode,
   entryPoints,
+  publicPath,
 }: {
   mode: Mode;
   entryPoints: EntryPoints[];
+  publicPath: string;
 }): Config => {
   const frontity = getFrontity();
   const babel = getBabel({ mode });
-  const webpack = getWebpack({ mode, babel, frontity, entryPoints });
+  const webpack = getWebpack({
+    mode,
+    babel,
+    frontity,
+    entryPoints,
+    publicPath,
+  });
   return {
     babel,
     webpack,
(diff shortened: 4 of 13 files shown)


DAreRodz:
Uhm, with the current changes the `serve` command will fail after doing `npx frontity build --publicPath="/custom/path/"`.

I'm not sure but I think the problem is in this line:

https://github.com/frontity/frontity/blob/661ac08316f44172166e79b05b47f0c15a837a9a/packages/core/src/server/index.tsx#L30

Should we have to add the `--publicPath` option to the `serve` command as well?

**cc:** @luisherranz

luisherranz:
No, the public path is in the `stats.publicPath`: https://github.com/frontity/frontity/blob/661ac08316f44172166e79b05b47f0c15a837a9a/packages/core/src/server/index.tsx#L58

DAreRodz:
Do you mean that I have to call `getStats` (which is async) and then get `stats.publicPath` and use it in this line?

https://github.com/frontity/frontity/blob/661ac08316f44172166e79b05b47f0c15a837a9a/packages/core/src/server/index.tsx#L30

DAreRodz:
I found a problem while trying to implement that idea about adding e2e testing to this feature using `--publicPath` argument for [`e2e:prod` scripts](https://github.com/frontity/frontity/blob/f952184dfbeeea1cd8f7a4c151998ba1c3f86149/package.json#L7-L9).

The next line is only executed when `--target` is not specified:

https://github.com/frontity/frontity/blob/13e596be378f1f7e730672885ed94b4d491a4dae/packages/core/src/server/utils/stats.ts#L55

So it would need `e2e:prod:all` script to be tested against all `--publicPath` cases.

I think we have to find another way to test this. :shrug:

luisherranz:
> So it would need e2e:prod:all script to be tested against all --publicPath cases.
> I think we have to find another way to test this. 🤷

What is the problem of adding `--publicPath` to `e2e:prod:all`?

DAreRodz:
> What is the problem of adding --publicPath to e2e:prod:all?

No, is not a problem, what I wanted to say is that the `e2e:prod:all` command should be tested using all the possibilities for the `--publicPath` argument.

---

EDIT: I don't know why I didn't think of that but we can add more `package.json` scripts to cover those cases, hehe. Would it be a good idea?

Of course, the new cases would need to be added to the GitHub action configuration.

package.json
```diff
@@ -6,8 +6,8 @@
     "test:ci": "jest --clearCache && lerna run test:ci --stream",
     "codecov": "curl -s https://codecov.io/bash | bash -s",
     "e2e:prod:all": "cd e2e/project/ && npx frontity build && npx frontity serve --port 3001",
-    "e2e:prod:module": "cd e2e/project/ && npx frontity build --target module && npx frontity serve --port 3001",
-    "e2e:prod:es5": "cd e2e/project/ && npx frontity build --target es5 && npx frontity serve --port 3001",
+    "e2e:prod:module": "cd e2e/project/ && npx frontity build --publicPath /custom/path --target module && npx frontity serve --port 3001",
+    "e2e:prod:es5": "cd e2e/project/ && npx frontity build --publicPath /custom/path/ --target es5 && npx frontity serve --port 3001",
```
michalczaplinski, line 10: I think this is fine as a workaround for now, but in the future, I suggest that we create a `e2e.test.js` node script that we can call from the github e2e-tests action [here](https://github.com/frontity/frontity/blob/dev/.github/workflows/e2e-tests.yml#L34)

And that script could take any number of parameters for anything that we want to test:
- target: `es5`
- target: `module`
- publicPath: `/custom/path`
- no public path specified

and whatever other params we have in the future.

packages/core/src/server/index.tsx
```diff
@@ -27,7 +27,13 @@ export default ({ packages }): ReturnType<Koa["callback"]> => {
   const app = new Koa();
 
   // Serve static files.
-  app.use(mount("/static", serve("./build/static")));
+  app.use(async (ctx, next) => {
+    const moduleStats = await getStats({ target: "module" });
+    const es5Stats = await getStats({ target: "es5" });
+    const stats = moduleStats || es5Stats;
+    const publicPath = stats ? stats.publicPath : "/static";
+    return mount(publicPath, serve("build/static"))(ctx, next);
```
michalczaplinski, line 35: is the `build/static` always going to refer to the same path as the `./build/static` from before the change?

michalczaplinski (changes requested):
Looks good to me, nice work @DAreRodz ! 🙂 

1. I'm not 100% sure about some minor details. 
2. Also, are you going to add the support for the env variables in this PR ?

I'm talking about:
target: FRONTITY_TARGET.
port: FRONTITY_PORT.
https: FRONTITY_HTTPS.
production: FRONTITY_PRODUCTION.
development: FRONTITY_DEVELOPMENT.
publicPath: FRONTITY_PUBLIC_PATH.

packages/core/src/server/index.tsx
DAreRodz: With `--publicPath` you change the path where those files are served (the URL) but they are always stored in the same folder (`./build/static`).

I don't know if this answer your question. 😅

DAreRodz:
> 1. I'm not 100% sure about some minor details.

I'll try to answer in your comments.

> 2. Also, are you going to add the support for the env variables in this PR ?
 
No, I'm not. The idea is to implement that in another PR. ☺

packages/core/src/server/index.tsx
luisherranz: Yes, I used an internal variable named `outDir` for the `/build` folder, but right now there's no way to change it from the outside. It's there waiting until we release the `frontity.config.js` file or we add a `--outDir` option.

luisherranz:
@DAreRodz if you are not planning to do any code changes I think you have to re-request a review, don't you?

packages/core/src/server/index.tsx
michalczaplinski: oh, I meant are there any circumstances where `build/static` can point to a different path than `./build/static`, but I've realized that they're exactly the same: https://unix.stackexchange.com/a/397526

michalczaplinski (approved):
LGTM @DAreRodz 👍

PR #429, opened 2020-05-08
Implement `useInfiniteScroll` hook

orballo:
**What**:

Implements the `useArchiveInfiniteScroll`, `usePostsInfiniteScroll` and `useInfiniteScroll` hooks.

**Why**:

To facilitate the implementation of infinite scroll to Frontity users.

**How**:

This are the types of the current implementations.
```ts
type UseInfiniteScroll = (options: {
  currentLink: string;
  nextLink?: string;
  fetchInViewOptions?: IntersectionOptions;
  routeInViewOptions?: IntersectionOptions;
}) =>
  | {
      supported: false;
      routeRef?: undefined;
      fetchRef?: undefined;
      routeInView?: undefined;
      fetchInView?: undefined;
    }
  | {
      supported: true;
      routeRef: (node?: Element) => void;
      fetchRef: (node?: Element) => void;
      routeInView: boolean;
      fetchInView: boolean;
    };
```
```ts
type UseArchiveInfiniteScroll = (options?: {
  limit?: number;
  active?: boolean;
}) => {
  pages: {
    key: string;
    link: string;
    isLast: boolean;
    Wrapper: React.FC<Connect<Source & Router>>;
  }[];
  isLimit: boolean;
  isFetching: boolean;
  fetchNext: () => Promise<void>;
};
```
```ts
type UsePostTypeInfiniteScroll = (options?: {
  limit?: number;
  active?: boolean;
  archive?: string;
  fallback?: string;
}) => {
  posts: {
    key: string;
    link: string;
    isLast: boolean;
    Wrapper: React.FC<Connect<Source & Router>>;
  }[];
  isLimit: boolean;
  isFetching: boolean;
  fetchNext: () => Promise<void>;
};
```
```ts
type Wrapper = (link: string) => React.FC<Connect<Source & Router>>;
```

Examples of how they could be used:
```tsx
import useArchiveInfiniteScroll from "@frontity/hooks/use-archive-infinite-scroll";

const Archive = () => {
  const { pages, isLimit, isFetching, fetchNext } = useArchiveInfiniteScroll();

  return (
    <div>
      {/* <ArchivePage /> would be a custom component created by the developer. */}
      {pages.map(({ key, link, isLast, Wrapper }) => (
        <Wrapper key={key}>
          <ArchivePage link={link} />
          {!isLast && <hr />}
        </Wrapper>
      ))}
      {/* <Loading /> would also be a custom component created by the developer. */}
      {isFetching && <Loading />}
      {isLimit && <button onClick={fetchNext}>Load Next Page</button>}
    </div>
  );
};
```
```tsx
import usePostTypeInfiniteScroll from "@frontity/hooks/use-post-type-infinite-scroll";

const PostTypeList = () => {
  const { state } = useConnect();
  
  const { posts, isLimit, isFetching, fetchNext } = usePostTypeInfiniteScroll({
    active: !!data.isPost
  });

  return (
    <div>
      {/* <PostType /> would be a custom component created by the developer. */}
      {posts.map(({ key, link, isLast, Wrapper }) => (
        <Wrapper key={key}>
          <PostType link={link} />
          {!isLast && <hr />}
        </Wrapper>
      ))}
      {/* <Loading /> would also be a custom component created by the developer. */}
      {isFetching && <Loading />}
      {isLimit && <button onClick={fetchNext}>Load Next Page</button>}
    </div>
  );
};
```

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests
- [x] TypeScript tests
- [x] Changesets (one for the new `updateState` action in router,  one for the infinite scroll itself and one for the fix in `@frontity/types`)

**To be done in a later PR**:
- End to end tests
- Documentation

**Irrelevant**:
- ~Update starter themes~
- ~Update other packages~
- ~Community discussions~


orballo:
@luisherranz Could you please have a look at this? Let me know of any changes or doubts on the API.

It still needs some adjustments and cleaning but everything seems to be working as expected. I'll be working on that and implementing tests and e2e.

luisherranz:
Sure, I'll take a look 🙂

luisherranz:
Sorry I haven't been able to check this today. I'll try to do it this weekend.

orballo:
@luisherranz I was thinking that maybe `usePostsInfiniteScroll` should take `context` option and `forceContext` option. So, what I have in mind are these three scenarios:

1. I want to use infinite scroll only if the context is set  in `state.router.state.context`.
2. I want to use any context set in `state.router.state.context` and use `options.context` as default (i.e. SSR).
3. I want to use my `options.context` no matter if the `state.router.state.context` is already set, so I can use a custom route as context like `@related-articles/:id`.

Other option would be not to use `options.context` as default but always forcing it, so that `options.context` is used only to override any previous context set, and the default context would be always `/`. If somebody wants the first scenario can force it programaticaly in the component where they are calling `usePostsInfiniteScroll`.

Let me know what do you think.

luisherranz:
I will have to go over `usePostsInfiniteScroll` to answer you on that. I only had time to check `useInfiniteScroll` and `useArchiveInfiniteScroll`. I'll try to do it this morning.

---

I really like it so far. And only have one suggestion: Why don't you create the `Wrapper` on demand for each page item with `link` embedded?

```ts
const Wrapper = (link: string): Wrapper =>
  connect(({ state, children }) => {
    // ...
  });
```

And then

```ts
// Map every link to its DIY object.
const pages = links.map((link) => ({
  key: link,
  link: link,
  isLastPage:
    link === last.link || (link === links[links.length - 2] && !last.isReady),
  Wrapper(link),
}));
```

Apart from that, I don't like the idea of using `actions.router.set` to replace `state.router.state`. Once we have hooks, people may subscribe to that action to do something when the URL has changed, and if we start doing this, people would have to check if the URL was the same or not.

So could you please do a PR to create `actions.router.updateState` and switch to that? Internally it does a `window.history.replaceState` with the current link and new `state`.

As soon as we have hooks, I'll create one to silently abort an `actions.router.set` if it has the same `link` to avoid causing problems to people subscribing to its executions.

packages/connect/src/index.js
```diff
@@ -1,5 +1,5 @@
 export { observe, unobserve } from "./observer";
-export { observable, isObservable, raw } from "./observable";
+export { observable, isObservable, raw, clone } from "./observable";
```
luisherranz: Could you please move `clone` to `tiny-router`? Sorry I think I probably told you to put it here but I don't want this to be a public API of `connect` yet. I want to discuss it with Miklos and see if how this can be avoided.

packages/hooks/use-archive-infinite-scroll.tsx
```diff
@@ -0,0 +1,139 @@
+import React, { useEffect } from "react";
+import { useConnect, connect, css } from "frontity";
+import { Connect } from "frontity/types";
+import useInfiniteScroll from "./use-infinite-scroll";
+import WpSource from "@frontity/wp-source/types";
+import TinyRouter from "@frontity/tiny-router/types";
+
+type Wrapper = React.FC<Connect<WpSource & TinyRouter, { link: string }>>;
```
luisherranz: Are you using the types of `WpSource` and `TinyRouter` instead of `Source` and `Router` for something in particular?

packages/hooks/use-posts-infinite-scroll.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+  return (
+    <div css={container} ref={routeRef}>
+      {children}
+      {!hasReachedLimit && <div css={fetcher} ref={fetchRef} />}
+    </div>
+  );
+});
+
+const usePostsInfiniteScroll: UsePostsInfiniteScroll = (options) => {
```
luisherranz: Could you please rename this to `usePostTypeInfiniteScroll` (singular)? I want to start using the same terms everywhere, and we are going to start using `postType` for custom post types, including posts and pages. Thanks!

packages/tiny-router/src/actions.ts
```diff
@@ -1,5 +1,5 @@
 import TinyRouter from "../types";
-import { warn } from "frontity";
+import { warn, clone } from "frontity";
```
luisherranz: Just import `clone` from Ramda here.

packages/hooks/use-archive-infinite-scroll.tsx
orballo: Just because I tend to use the types of the packages I'm working with. But in this case I guess we could use the generics packages, so different implementations are compatible?

luisherranz: Yes, it should work with any theme or source so it's better if we use the general ones.

orballo:
> I really like it so far. And only have one suggestion: Why don't you create the `Wrapper` on demand for each page item with `link` embedded?
> 
> ```ts
> const Wrapper = (link: string): Wrapper =>
>   connect(({ state, children }) => {
>     // ...
>   });
> ```
> 
> And then
> 
> ```ts
> // Map every link to its DIY object.
> const pages = links.map((link) => ({
>   key: link,
>   link: link,
>   isLastPage:
>     link === last.link || (link === links[links.length - 2] && !last.isReady),
>   Wrapper(link),
> }));
> ```

This forces rerenders of the whole list because I need to recreate every `Wrapper` when a new link is added to the list.

I might have miss something here. So if you have an idea on how to do that implementation, that's something worth trying again.

luisherranz:
You can use a `memoize` function to cache the result.

luisherranz:
After a call we have decided to go with this logic to decide which `archive` is going to be used for the postType infinite scroll:

image

Drawing: https://excalidraw.com/#json=6498078429282304,3HeDJQH8qiEJG4xFZjUfNg

orballo:
There are a couple of things that need to be changed regarding `usePostTypeInfiniteScroll`:

1. Rename the option/state `context` to `archive`.
2. Move the infinite scroll state from `state.router.state` to a namespace like `state.router.state.infiniteScroll`.
3. Create a `state.router.previous` so that the archive can be guessed from there.

orballo:
@luisherranz I was going to implement the changes on the namespace, and I realized that to prevent overriding the state I need to do, in my case, something like this:

```ts
actions.router.set(currentLink, {
  method: "replace",
  state: {
    ...state.router.state,
    infiniteScroll: {
      ...state.router.state.infiniteScroll,
      links,
    },
  },
});
```

This looks to me very convoluted and prone to errors and overriding necessary state by other packages or developers that might not be thinking about preventing that. Maybe we should go with a different API for `actions.router.set` like:

```ts
actions.router.set("/", {
  method: "replace",
  state: {
    namespace: "infiniteScroll",
    data: {
      ...state.router.state.infiniteScroll,
      links,
   },
  },
});
``` 
or
```ts
actions.router.set("/", {
  method: "replace",
  namespace: "infiniteScroll",
  state: {
      ...state.router.state.infiniteScroll,
      links,
   },
});
```

What do you think? Maybe you can suggest another approach.

luisherranz:
That is going to be managed with `actions.router.updateState` so what about doing a merge?

Similar to the React' `setState` in classes.

```js
actions.router.updateState({
  infiniteScroll: {
    links: [ updated ]
  }
});
```

I'm not super fan to be honest, because this may not be what the developer wants or expects. The first example, even when it looks convoluted, gives more control to the caller to delete everything or part of the current state.

orballo:
Yep, I thought about a merge too and it looks like a good idea, but it can be misleading. If you think the first case is not that awful, we can just let it be that way, at least for the moment?

One question, just curious, I'm thinking also about when we need to change the state with a `set` and some other package needs to do it's changes too. That other package would be listening to a hook in `actions.router.set` and then fire it's own `actioins.router.updateState` I guess, right? Or, I just remembered, you mentioned something like a `filter` that maps the params before executing the action?

orballo:
I'm going to list all the changes so they are just in one post:
- [x] Move state to `infiniteScroll` namespace.
- [x] Rename `usePostsInfiniteScroll` to `usePostTypeInfiniteScroll`.
- [x] Create `Wrapper` on demand using a memoize function to prevent rerenders.
- [x] Remove `clone` from `@frontity/connect`.
- [x] Use `@frontity/source` and `@frontity/router` types instead of `@frontity/wp-source` and `@frontity/tiny-router`.
- [x] Rename `context` to `archive` in `usePostTypeInfiniteScroll`.
- [x] Create a `state.router.previous` field holding the previous `state.router.link`.
- [x] Create the action `actions.router.updateState` to update the browser state without changing the route.
- [x] Use the action `actions.router.updateState` in the infinite scroll hooks.
- [x] Implement the new logic for `usePostTypeInfiniteScroll`.

orballo:
@luisherranz I'm working on the new logic for `usePostTypeInfiniteScroll` and I don't think it's possible to check if the current post type has a related archive. Is there a way to do that? Am I missing something?

luisherranz:
> That other package would be listening to a hook in actions.router.set and then fire it's own actions.router.updateState I guess, right? Or, I just remembered, you mentioned something like a filter that maps the params before executing the action?

Yes, they could do that, but in that case it would make sense to just modify the args. `hooks` are going to be something like this:

```js
export default {
  state: {
    hooks: {
      actions: {
        myPackage: {
          addStuffToRouterState: {
            path: "actions.router.set",
            library: "myPackage.addStuffToRouterState",
            priority: 5,
          },
        },
      },
    },
  },
  libraries: {
    hooks: {
      actions: {
        myPackage: {
          addStuffToRouterState: ({ args }) => {
            const [link, options] = args;
            options.state = {
              ...options.state,
              myStuff: {
                // ...
              },
            };
          },
        },
      },
    },
  },
};
```

> I don't think it's possible to check if the current post type has a related archive. Is there a way to do that? Am I missing something?

You're right. Right now you could check the settings of `wp-source`, but in the future, with other source packages, that information is not going to be there because postTypes will work without settings. I guess we would need a field called `archive` in each postType.

We can solve it manually, but using `options.archive` on the hook of each postType, but that would break the nice `state.router.previous` check with the current behavior. So I guess we finally need another option to make `options.archive` overwrite `state.router.previous`.

What about having:

```ts
interface Options {
  /**
   * If infinite scroll should be activated or not. Useful for sharing
   * the same component between posts and pages.
   */
  active: boolean;

  /**
   * The link of an archive to be used. It used in all circumstances,
   * even with `actions.router.set` contains an archive link.
   */
  archive: string;

  /**
   * The link of an archive to be used when other archive links are not
   * present, like the link specified on `actions.router.set` or
   * the link in `state.router.previous`.
   */
  fallback: string;
}
```

What do you think?

image


Drawing: https://excalidraw.com/#json=5687473816469504,KhkqMA8iJli8lYzK-pceDA

orballo:
It looks good to me. I'll implement it that way. I'll let you know if I run into something else.

orballo:
@luisherranz do you think we should implement `options.active` in `useArchiveInfiniteScroll` too? I can't think of any scenario, but it might be useful for somebody to be able to stop the infinite scroll behavior at any point 🤔

luisherranz:
Yes, let's make both consistent 🙂

orballo:
@luisherranz one more thing, I'm not sure how to call the fields `posts` (for `usePostTypeInfiniteScroll`) and `pages` (for `useArchiveInfiniteScroll`) in the return values. I was thinking maybe `list`? `entities`?

luisherranz:
To be honest, I think that semantically `pages` in `useArchiveInfiniteScroll` and `posts` in `usePostTypeInfinteScroll` make the more sense to me.

I don't like `entities` because they don't relate to pages, only to posts or taxonomies.

I guess we could use `list` and let the user decide what to to use:

```js
// useArchiveInfiniteScroll
list.map((page) => {
  // ...
});

// usePostTypeInfinteScroll
list.map((post) => {
  // ...
});
```

versus

```js
// useArchiveInfiniteScroll
pages.map((page) => {
  // ...
});

// usePostTypeInfinteScroll
posts.map((post) => {
  // ...
});
```

Up to you, but I wouldn't mind going with `pages` and `posts` either because they belong to two different hook, one to do infinite scroll of pages and the other to do infinite scroll of posts.

orballo:
Oh, I forgot to reply. I decided to keep `posts` and `pages` as you suggested.

luisherranz:
Let me know once this is ready for a beta release.

orballo:
It's ready.

luisherranz (approved):
I haven't checked the tests, but the code looks great to me.

I have only one question/doubt, I have added it below. I guess we'll see if that is a problem once we implement this in real apps.

We're going to do a beta release 🙂

packages/hooks/use-infinite-scroll.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      links.push(nextLink);
+
+      actions.router.updateState({
+        ...state.router.state,
+        infiniteScroll: {
+          ...state.router.state.infiniteScroll,
+          links,
+        },
+      });
```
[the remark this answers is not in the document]
orballo: If it is not ready, there is no ref catching the scroll to update the link. (At least in the current implementations) So maybe this is a valid concern if we create some other hook that uses `useInfiiniteScroll` and works in a different way.
However, even if the link is set before ready, the theme should be prepared for that.

luisherranz:
Types were failing so I have added some "as any" to be able to release the beta version in https://github.com/frontity/frontity/pull/429/commits/058f65db1f85bc59becccedc890719ee1c0753bc. Please take a look whenever you continue with this.

luisherranz:
I have published the packages in beta. You can install them using these tags:

- `frontity@1.8.1-infinite-scroll-beta.0`
- `@frontity/core@1.7.1-infinite-scroll-beta.0`
- `@frontity/hooks@2.1.0-infinite-scroll-beta.0`
- `@frontity/tiny-router@1.3.0-infinite-scroll-beta.0`

The branch for this beta releases is [`infinite-scroll-beta`](https://github.com/frontity/frontity/tree/infinite-scroll-beta). If you need to add additional changes, open a PR on that branch.

luisherranz:
@SantosGuillamot @orballo @javiersefer: you can now test it out in your themes. There is an example of how to use it in the opening post of this PR.here is an example of how to use it in the opening post of this PR.

PR #430, opened 2020-05-09
Implement `useFills` hook

### LINKED ISSUE #431: Create the useFills hook [5pt]
**Feature:** [Slot and Fill](https://community.frontity.org/t/slot-and-fill/895).
**User Stories**:
  - Add additional components to the theme.
  - Substitute certain components of the theme.

[**Sprint Card**](https://github.com/orgs/frontity/projects/4#card-36476149)

michalczaplinski:
**What**:

Implement the `useFills` React hook which is a prerequisite for creating `<Slot/>` components for a Slot and Fill pattern. This will allow theme creators to create "slots" where the theme users can insert their "fill" components. The "fill" can be any React component.

**Why**:

To allow users to create "slots" in themes that can then be filled by "fills". More rationale[ in the Feature Discussion](https://community.frontity.org/t/slot-and-fill/895)

**How**:

Following the Implementation Proposal: https://community.frontity.org/t/slot-and-fill/895/8

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Changeset
- [x] Community discussions
- [x] TSDocs
- [x] Unit tests
- [x] Move to "frontity" package
- [x] Add namespaces to `state.fills` and `libraries.fills`
- [x] End to end tests
- [x] TypeDocs
- [x] Documentation https://github.com/frontity/gitbook-docs/pull/75 (still WIP)

Unrelated:
- ~~TypeScript tests~~
- ~~Update starter themes~~
- ~~Update other packages~~


packages/hooks/use-fills.tsx
```diff
@@ -0,0 +1,39 @@
...
+      };
+    };
+  };
+  libraries: {
+    [key: string]: any;
+  };
+}
+
+const useFills = (name) => {
+  const { state, libraries } = useConnect<Fills>();
```
michalczaplinski: @orballo Could you tell me if this is the right approach to typing the output of `useConnect` ?

Or should I import the types from somewhere ?

orballo: That's the right way

packages/hooks/package.json
```diff
@@ -22,5 +22,9 @@
   },
   "dependencies": {
     "frontity": "^1.5.0"
+  },
+  "devDependencies": {
+    "react-test-renderer": "^16.13.1",
+    "@frontity/connect": "1.0.5"
```
luisherranz: Please don't add fixed dependencies.

packages/hooks/use-fills.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  Object.values(state.fills)
+    // Match only the fills for this name.
+    .find((fill) => fill.name === name)
```
luisherranz: This is `fill.slot`. `fill.name` doesn't exist. I have fixed it in the IP.

packages/hooks/use-fills.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+        FillComponent,
+      },
+    },
+  });
+});
+
+describe("test", () => {
+  it("should work", () => {
+    const Comp = () => {
+      const fills = useFills("test fill");
```
luisherranz: This is "slot 1". We want all fills for "slot 1" beucase this component is "Slot 1".

packages/hooks/use-fills.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+      return fills.map(({ Fill, props }) => <Fill key={props.id} {...props} />);
```
luisherranz: `props.id` is not guaranteed to exist. Let's add the fill's name as key: https://community.frontity.org/t/slot-and-fill/895/10?u=luisherranz

packages/hooks/use-fills.tsx
```diff
@@ -1,39 +1,73 @@
...
+const useFills = (name: string) => {
+  const { state, libraries } = useConnect<FillsPackage>();
+
+  if (!name) {
+    warn("You should pass the name of the slot that you would like to fill!");
+    return [];
+  }
+
+  if (!Object.values(state.fills).find((slot) => slot.slot === name)) {
+    warn(`Could not find any slot with name ${name}. Is this what you want?`);
```
luisherranz: This warning is not useful because the theme will declare many slots and most of them will be empty.

michalczaplinski:
As this is nearing finish, there are still some remaining questions:

1. Do we want to implement the [data assertions](https://community.frontity.org/t/slot-and-fill/895/3?u=mmczaplinski) here as well?
2. Is the current behaviour or returning the `[]` as a "default fill" okay?
3. Are we deprecating exporting the `roots` property of the state and does that mean refactoring e.g. the GTM package?

packages/hooks/package.json
```diff
@@ -18,9 +18,14 @@
   "scripts": {
     "test:ci": "../../node_modules/.bin/jest --passWithNoTests",
     "test": "../../node_modules/.bin/jest --watch",
-    "test:inspect": "node --inspect ./node_modules/jest/bin/jest.js --watch --no-cache --runInBand"
+    "test:inspect": "node --inspect ../../node_modules/jest/bin/jest.js --watch --no-cache --runInBand"
```
michalczaplinski, line 22: This was incorrect before for some reason.

packages/types/src/package.ts
```diff
@@ -15,6 +24,7 @@ export interface Package {
       packages?: string[];
       platform?: "client" | "server";
       rendering?: "ssr" | "csr";
+      fills: { [key: string]: Fill };
```
michalczaplinski: I'm not entirely sure if this is how I'm supposed to add the types here. @DAreRodz Do you have some experience with this?

luisherranz: You put `fills` inside `state.frontity`. This is wrong, it's not `state.frontity.fills`, it is `state.fills`.

luisherranz:
> 1. Do we want to implement the (data assertions)[https://community.frontity.org/t/slot-and-fill/895/3?u=mmczaplinski] here as well?

No, that's not included in the Implementation Proposal: https://community.frontity.org/t/slot-and-fill/895/6

> 2. Is the current behaviour or returning the `[]` as a "default fill" okay?

Yes, if there are no fills for that slot name an empty array is fine so people can still use `fills.map`.

> 3. Are we deprecating exporting the `roots` property of the state and does that mean refactoring e.g. the GTM package?

I suggested to rename it (while keeping backward compatibility), but that's outside of the scope of the Implementation Proposal.

packages/types/src/package.ts
michalczaplinski: Right, my bad, that's true. 

But assuming that I put them in `state.fills` my question was more about the general idea. Is this the right place to put the types for `fills`? I think so, but would just like to confirm.

michalczaplinski:
> No, that's not included in the Implementation Proposal: https://community.frontity.org/t/slot-and-fill/895/6

Right, I understand, but there were several updates to the implementation proposal and as the data assertions were mentioned at some point it was not 100% clear that we are not implementing them at this point 🙂

packages/types/src/package.ts
luisherranz: Yes, I think the placement and the types are right 👍

michalczaplinski:
Really not sure why pushing with `--force`  closed the PR 🤷

michalczaplinski:
The diff is showing the changes from the base branch (`use-connect`) relative to `dev` even though I'm trying to merge `use-connect` it. I'm not sure how to unfuck this.

Let's wait until the `use-connect` branch is merged into `dev`, so then we can change the base branch to `dev` and the diff should be clean.

michalczaplinski:
FYI, The [docs](https://github.com/frontity/gitbook-docs/pull/69) are not 100% finished for this but the code can be reviewed.

packages/hooks/use-fills.tsx
```diff
@@ -0,0 +1,71 @@
...
+      // 1. Nest the `key` of the fill
+      // 2. Add default priority of 10 if not present
+      .map(([key, value]) => ({
+        key,
+        priority: value.priority || 10,
+        ...value,
+      }))
+
+      // Match only the fills for this name.
+      .filter(({ slot }) => slot === name)
```
luisherranz: You can do this first.

packages/hooks/use-fills.tsx
```diff
@@ -0,0 +1,71 @@
...
+    fills: { [key: string]: Fill };
+  };
+  libraries: {
+    fills: {
+      [key: string]: React.ComponentType;
+    };
+  };
+}
+
+const useFills = (name: string) => {
```
luisherranz: This is missing TSDocs.

Maybe explicitly define the return type as well, so we cannot break it in a refactoring.

packages/hooks/use-fills.tsx
```diff
@@ -0,0 +1,71 @@
+import { useConnect, warn } from "frontity";
+import { Fill } from "@frontity/types";
+
+interface FillsPackage {
```
luisherranz: Instead of redeclaring the type, you can import `Package`. That way if we change the origin type, it changes here as well.

packages/types/src/package.ts
```diff
@@ -1,3 +1,12 @@
+export interface Fill {
+  slot: string;
```
luisherranz: Can you add TSDocs to this file as well?

packages/hooks/use-fills.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+    // We should only render 1 component with the prop `{number: 1}`
+    expect(app.toJSON().props).toEqual({
+      id: "test-fill",
+      "data-number": 1,
+    });
+
+    expect(app.toJSON()).toMatchSnapshot();
+  });
+});
```
luisherranz: Great tests.

The only one I miss is to check if it doesn't include a fill that doesn't match the slot name.

luisherranz (changes requested):
Awesome work.

I've added some details.

And as it seems that we are going to start using this package a lot, I would like to take this opportunity to mark this as side effect free.

https://webpack.js.org/guides/tree-shaking/

Could you please do so (`sideEffects: false`) and create and index file with all the exports?
Thanks!

.changeset/lucky-rocks-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontity/hooks": minor
```
luisherranz: Totally my fault I didn't mention this earlier, but could you please move it to the main `"frontity"` package?

luisherranz (changes requested):

packages/hooks/use-fills.tsx
michalczaplinski: 👍   on the TSDoc.

The return type is:
Screenshot 2020-05-22 at 14 42 44

because I assign a type here: https://github.com/frontity/frontity/blob/e2056a31a608e05d58f60c7ec98ee4ec5934209e/packages/hooks/use-fills.tsx#L47-L49

I need to specify the type of that accumulator so that I get type checks inside of the `.reduce` callback.

So the type inference figures out the correct return type of `useFills` for me. Is there a problem with this approach that I'm not aware of?

packages/hooks/use-fills.tsx
michalczaplinski: 👍

packages/types/src/package.ts
michalczaplinski: I've tried to do this, but don't really see what benefit it brings to add the TSDoc here, to be honest 🤔 

TSDoc does not really have a way to "document" the interfaces. I also don't really know what I should document here? It's just a type for the `Fill`, it should be self-evident what it is.

Could you elaborate on what you had in mind?

luisherranz: Let's not go over this, please. We need to add TSDocs to all our interfaces, functions, classes, types...

Interfaces are documented in the same way than classes.

packages/hooks/use-fills.tsx
luisherranz: Explicit return types make impossible to break the function (type-wise) in a subsequent refactoring.

luisherranz (dismissed):
I've done the TSDocs myself so you can get the idea of how they work. I've also added the explicit return type so you can see what I mean.

I think the only thing missing is to move this to the main `"frontity"` package and the unit test.

luisherranz:
I'm going to take over and finish this because I also want to add namespaces to both `state.fills` and `libraries.fills` as I have explained in the [feature discussion](https://community.frontity.org/t/slot-and-fill/895/17?u=luisherranz).

DAreRodz:
@luisherranz, I've assigned you the issue related to this PR (#431) and moved it to **in progress**.

luisherranz:
Thanks David! I totally have to get used to the board 😅

luisherranz:
@michalczaplinski, this is ready for code review. I cannot add you as reviewer but could you please check it out?

I saw that TS 3.9 included some breaking changes so I did a small refactor of the types. Then I saw that it also broke our `source` data types, so I removed the upgrade and fixed the version to TS 3.8 to open that in another PR.

Other than that I moved the hook to `"frontity"`, refactored to use namespaces inside `state.fills` and `libraries.fills`, added more TSDocs and some simple e2e tests.

I still want to:

- Explore how to export the markdown from TSDocs.
- Take a look at the Doc PR: https://github.com/frontity/gitbook-docs/pull/75

But the code should be fine now.

michalczaplinski:
Just FYI:

> I still want to:
> - Take a look at the Doc PR: frontity/gitbook-docs#75

The docs are mostly there, but I think that they still miss some more details. Are you planning to you finish those?

packages/frontity/src/utils/slot-and-fill/use-fills.tsx
```diff
@@ -0,0 +1,75 @@
...
+  if (!name) {
+    warn("You should pass the name of the slot that you would like to fill!");
+    return [];
+  }
+
+  return (
+    // Flat all the fills and turn them into entries.
+    Object.values(state.fills || {})
+      .reduce<[string, StateFill][]>(
+        (fills, fill) => [...fills, ...Object.entries(fill)],
```
michalczaplinski: Just a nit but wouldn't this be clearer as:
```
 (fills, namespace) => [...fills, ...Object.entries(namespace)],
```

packages/core/src/app/index.tsx
```diff
@@ -15,9 +15,6 @@ const App: React.FunctionComponent<Props> = ({ store, helmetContext = {} }) => {
         {Object.entries(store.roots).map(([namespace, Root]) => (
           <Root key={namespace} />
         ))}
-        {Object.entries(store.fills).map(([namespace, Fill]) => (
-          <Fill key={namespace} />
-        ))}
```
michalczaplinski, line 20: So, the `Fill` components were not supposed to be placed there in the first place?

michalczaplinski (commented):
I can't "approve" or request changes  because I'm the author, but I have a couple of comments.

lerna.json
```diff
@@ -1,12 +1,4 @@
...
   "packages": ["packages/*", "examples/*", "e2e/project", "e2e/packages/*"],
-  "version": "independent",
-  "command": {
-    "publish": {
-      "allowBranch": "master",
-      "conventionalCommits": true,
-      "githubRelease": true,
-      "message": "chore(release): publish [skip ci]"
-    }
-  }
```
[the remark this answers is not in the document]
luisherranz: We are using `changesets` instead of `lerna` to publish our packages so that is not necessary anymore.

Sorry to have done that here but something failed in lerna, I don't remember exactly what, so I removed that code to fix it and because we don't use it anymore.

packages/connect/src/index.d.ts
```diff
@@ -9,9 +9,15 @@ interface ObserveOptions {
   lazy?: boolean;
 }
 
+type Function = (...args: any) => any;
```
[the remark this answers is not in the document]
luisherranz: Oh, thanks. I'll remove that.

packages/core/src/app/index.tsx
luisherranz: They were. It was the first approach to integrate Slot and Fill and the way we used it in the old framework.

But then I decided to use `state.fills` and `libraries.fills` instead, which makes much more sense for Frontity.

It is explained in https://community.frontity.org/t/slot-and-fill/895

packages/frontity/src/utils/slot-and-fill/use-fills.tsx
luisherranz: Oh, absolutely. I'll change it.

luisherranz:
> Are you planning to you finish those?

Yes. But I want to do this first to export the markdown of the TSDocs and see how we can use it: https://community.frontity.org/t/code-documentation-workflow/1633/9

lerna.json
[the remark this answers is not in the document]
michalczaplinski: oh I see, thanks :)

michalczaplinski:
By the way, just to clarify, there were just a couple of comments but it LGTM 👍

michalczaplinski:
Oh, sorry, I've just realized that I think you forgot to include the changesets for all the other packages that you updated 🙂

luisherranz:
> Oh, sorry, I've just realized that I think you forgot to include the changesets for all the other packages that you updated 🙂

Thanks for the heads up! I have updated/added proper changesets 👍

luisherranz:
As the [Typedocs approach hasn't worked as we thought](https://community.frontity.org/t/adopting-a-standard-way-to-document-frontity-apis/1662/11?u=luisherranz), I'm going to continue with this PR.

luisherranz:
I have reused [your styles](https://github.com/wp-pwa/saturn-theme/blob/36852a9d8d39302e1603b15b879e16b0f38f80a4/src/shared/components/SlotInjector/index.js#L89-L93) @darerodz 🙂

I have added:

- Support for debug mode using `state.frontity.debug = true`.
- Unique `key`s by using the namespace as well as the fill name to avoid problems in case two packages using different namespaces use the same fill name.

This should be ready to merge. @michalczaplinski feel free to review these last two improvements again if you want.

I'll finish the docs tomorrow.

PR #435, opened 2020-05-19
Update lighthouse workflow to only deploy when the PR is opened by frontity

michalczaplinski:
Add a tiny change in the `lighthouse.yml` so that we only deploy to Vercel when the PR has been opened by a frontity team member. 

This should prevent the PRs from external contributors from trying to (and failing) to deploy to Vercel because the external PRs do not have access to our token.

- [x] Test new tasks action

- [ ] Test ignored tasks


luisherranz:
I thought that required tests couldn't be skipped because they will fail, but it doesn't seem the case: `2 skipped and 10 successful checks`.

What's missing then here Michal?

michalczaplinski:
The problem here is that those checks should not be skipped for this PR because I'm a member of the frontity organization on github (actually, I'm one of the "OWNERS", whatever that means).

So, the way that I added this check to the workflow file is not correct it seems, but I wasn't able to find the "right" way in the docs so far...

michalczaplinski:
I have just thought that maybe the workflows will run (almost) the way that we want them if we change the 

```yaml
# workflow.yml

on: [pull_request]
```

to 

```yaml
on:
  pull_request:
    types: [opened]
  push:
```

So, then, if a frontity team member pushes a commit to an external PR it should use the token, I think...

And I could just check out the PR locally, run `git commit --allow-empty -m "Trigger deploy" && git push` to trigger the deployment

luisherranz:
I saw this line when looking for examples of eslint actions:

https://github.com/frontity/frontity/blob/156bf2f8372e8aa5bd1f9d71e659bf302ed44f0e/.github/workflows/eslint.yml#L12

Maybe it would solve this problem?

luisherranz:
It looks like it works here. I'm going to try an external PR.

luisherranz:
It looks like it works here. I am going to test an external PR.

luisherranz:
Done in https://github.com/frontity/frontity/pull/505. It seems like the LightHouse action was skipped there. Let's see what happens with the merge, if GH allows it or not.

luisherranz:
[The test went great](https://github.com/frontity/frontity/pull/505#issuecomment-648742903) so I have merged it in this branch to test its changes.

luisherranz (approved):

luisherranz:
It looks like everything works here as well:

Screenshot 2020-06-24 at 13 32 47

So I'm going to merge this 🙂

PR #448, opened 2020-05-26
Fix processor return type

DAreRodz:
**What**:

Fix `Processor` return type.

**Why**:

Right now, TypeScript complains when a processor returns `null`, which is how nodes are removed.

There is an example in the [`@frontity/html2react` docs](https://github.com/frontity/frontity/pull/448), but I think we could explain this better.

**How**:

Just replacing `boolean` by `null` in the `Process` type here:

https://github.com/frontity/frontity/blob/f5fc7e42e1208dedf97ffec40170bfcc9fdffc71/packages/html2react/types.ts#L69L71

**Tasks**:

- [ ] TSDocs
- [x] TypeScript
- [x] End to end tests
- [x] TypeScript tests
- [ ] Link to PR in [Documentation](https://github.com/frontity/gitbook-docs/)

**Unrelated Tasks**

- Code
- Unit tests
- Update starter themes
- Update other packages
- Community discussions

### DIFF
--- a/.changeset/quiet-emus-joke.md
+++ b/.changeset/quiet-emus-joke.md
@@ -1,0 +1,5 @@
+---
+"@frontity/html2react": patch
+---
+
+Fix `Processor` type so TypeScript doesn't complain when a processor returns `null`, which is how nodes are removed.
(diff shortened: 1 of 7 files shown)


luisherranz:
Hey @darerodz, what's the status of this PR? Is it still needed?

DAreRodz:
Yes, it is! I guess we forgot about it because I started it but it wasn't added to the sprint.

DAreRodz:
I would have to take a look at this, it seems that almost all the work is already done. :thinking:

PR #450, opened 2020-05-28
E2E Tests that create custom "test" WordPress instances

### LINKED ISSUE #424: Research - E2E testing that requires WordPress instances [21pt]
**Feature:** [E2E testing that requires WordPress instances](https://community.frontity.org/t/e2e-testing-that-requires-wordpress-instances/1756).

michalczaplinski:
**What**:

This is a PoC of the E2E tests that will launch their own WordPress instances inside docker containers.

This PR follows the Feature Discussion: https://community.frontity.org/t/e2e-testing-that-requires-wordpress-instances/1756/

**Why**:

We want to be able to test features that rely on changes in a WordPress backend. E.g. we can test frontity's compatibility with certain plugins or compatibility with different WP versions.

**How**:

See the Feature Discussion mentioned above for details of design.

**Tasks**:

- [x] Code
- [x] End to end tests

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] TypeScript
- [ ] TSDocs
- [ ] Update starter themes
- [ ] Update other packages
- [ ] Unit tests

**Some of the outstanding tasks (in no particular order)**
- [x] Rename the `start.js` to `e2e.js`
- [x] Add the command line arguments to the `e2e.js` script (using `minimist`)
- [x] in dev run `npx frontity dev` but in prod run `build & serve`
- [x] remove all the unnecessary files (the ones in the `e2e-wp` folder)
- [x] at least locally run the new tests separately from the "old" test so that we don't have to wait for the docker container to launch for the tests that do not need the database. This can be accomplished with a CLI flag
- [x] do not activate the plugins automatically when installing them via a task. This should be done in a separate step or via the database dump
- [x] create tasks / scripts to:
  - [x] load the database
  - [x] dump the database
- [x] add a `publicPath` parameter to the CLI so that we can control the `publicPath` of the `build` 
- [x] simplify the npm scripts in the `e2e` folder
- [x] check if the cypress tasks can work with async functions directly or need that the user explicitly returns a `new Promise()`


e2e/plugins/index.js
```diff
@@ -1,6 +1,35 @@
...
+      //
+      // Asynchronous tasks HAVE TO return a promise.
+      // https://docs.cypress.io/api/commands/task.html#Return-a-Promise-from-an-asynchronous-task
+      //
+      //
+      return new Promise((resolve, reject) => {
+        (async () => {
+          try {
+            await execa.command(
+              `docker-compose run --rm wpcli wp plugin install ${name} --activate`,
```
luisherranz: I would leave the activation (or not) of the plugin to the test creator (so, to the database).

e2e/plugins/index.js
```diff
@@ -1,6 +1,35 @@
...
+    installPlugin({ name }) {
+      // Because cypress commands are async but are not promises,
+      // we have to wrap all task functions with a new Promise constructor.
+      // https://stackoverflow.com/a/49980928/2638310
+      //
+      // Asynchronous tasks HAVE TO return a promise.
+      // https://docs.cypress.io/api/commands/task.html#Return-a-Promise-from-an-asynchronous-task
+      //
+      //
+      return new Promise((resolve, reject) => {
```
luisherranz: Async functions do return a Promise themselves so I think it should be enough if you'd return the async function directly 🙂

luisherranz (commented):
Awesome 👏

.github/workflows/eslint.yml
```diff
@@ -1,4 +1,4 @@
-name: Eslint
+name: eslint
```
michalczaplinski, line 1: I took a moment to standardize the naming of all the workflows so that they're all small caps.

packages/core/package.json
```diff
@@ -63,7 +63,6 @@
     "koa-route": "^3.2.0",
     "koa-static": "^5.0.0",
     "lodash": "^4.17.15",
-    "minimist": "^1.2.5",
```
michalczaplinski, line 66: Removed because we are not using minimist in `@frontity/core`

luisherranz:
> There are some github actions that seem to be stuck forever:

That's because they are not run anymore but they are still required by our repository configuration. We can merge it without them and then change the configuration.

> It looks like the e2e tests only fail for google analytics

@darerodz is looking into that in this PR: https://github.com/frontity/frontity/pull/532

> I have not added the complete test for wp-comments because I realized that I would need to make some changes in the @frontity/wp-comments package itself to make tests pass and I want to keep this PR clean of changes to other packages. So, at the moment the test is not very useful, it just checks that the cypress tasks are running successfully.

Ok.

---

I've made a few changes before merging this:

- Add `process.exit(0)` after Cypress has finished to finish the Node process.
- Reactivate Eslint and added TSDocs to all the files.
- Change `--publicPath` to `--public-path` when running the Frontity project.
- Fix `--wp-version` (it was not working).
- Add `--target` and `--public-path` args when running the Frontity project in development.
- Fix suite names so they can be exclusive from each other (they were not working properly) and prepare them to run different wordpress/non-wordpress suites in parallel in the future.
- Remove activation of the WordPress plugins after installation.
- Add the `version` option to the `installPlugin` task.
- Fix export database command (it was not working) and add option to specify the path.
- Fix the `removePlugins` task (it was not working properly).
- Add `reset-db` command.
- Remove all the `wp-comments` test/package files.
- Add a couple of basic e2e tests to test that WordPress works, the REST API can be used and it can install plugins and replace the database.
- Update `README.md` with new sections explaining installation, usage and added the CLI argument options.
- Switch to `matrix.include` as we talked in [the FD](https://community.frontity.org/t/e2e-testing-that-requires-wordpress-instances/1756/30).

Everything is finished but I'm getting a `no such image: wordpress:: invalid reference format` error in the "Absolute Public Path" test. I'm going to check it out.

luisherranz (approved):
It seems to be working great finally. Let's merge this! 💪 💪

Issue #452, opened 2020-05-28
TypeScript 3.9 breaks `@frontity/source` types [8pt]


luisherranz:
**_Bug report_**

- I've updated both my project and my packages to the latest Frontity versions: https://docs.frontity.org/guides/keep-frontity-updated
- I've searched for existing issues.
- I've checked the documentation: https://docs.frontity.org
- I've checked the community forum: https://community.frontity.org

**Expected behavior**

The [`Data` types](https://github.com/frontity/frontity/blob/dev/packages/source/types/data.ts) should work.

**Observed behavior**

They work on TS 3.8 but they are broken on TS 3.9.

We think that they were working probably because of a bug.

**Steps involved to reproduce the problem**

Updating to `typescript` 3.9 breaks the data tests.

The update also broke other `connect` types but I have already fixed those on the PR.

**Possible solution**

We already wanted to get rid of that approach because it was not extensible. We did the research and we already have a solution. It's not written yet as an implementation proposal, but we could do so quickly.

https://community.frontity.org/t/extend-sources-data-interface-with-types-from-new-handlers/1187/28

I have already started a PR for this: https://github.com/frontity/frontity/pull/449

That approach not only fixes this problem but also makes the data types extensible.

**Temporal workaround**

Fix the `typescript` version to 3.8.

luisherranz:
It seems like we have the same problem with `Entities` in the `state.source.entity(link)` API.

More info in the FD that introduced the API: https://community.frontity.org/t/support-for-yoast-plugin-rest-api-fields/2626/23?u=luisherranz

PR #453, opened 2020-05-29
Fix lerna bootstrap in Windows and fix TypeScript

luisherranz:
**What**:

To fix TypeScript I fixed the version to 3.8 because 3.9 is causing issues with our current types. More info on https://github.com/frontity/frontity/issues/452.

To fix lerna bootstrap in Windows I've updated to lerna 3.22 and fixed a build command that didn't work on Windows.

**Why**:

It's important that people can contribute to Frontity even if they use Windows.

**Tasks**:

- [x] Code

**Unrelated Tasks**

- TSDocs
- TypeScript
- Unit tests
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages
- Community discussions

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -23,7 +23,7 @@
     "test:inspect": "node --inspect ../../node_modules/.bin/jest --watch --no-cache --runInBand",
     "dev:inspect": "node --inspect-brk -r ts-node/register src/scripts/dev.ts",
     "serve:inspect": "node --inspect-brk -r ts-node/register src/scripts/serve.ts",
-    "build": "../../node_modules/.bin/tsc --project ./tsconfig.build.json",
+    "build": "tsc --project ./tsconfig.build.json",
     "build:watch": "../../node_modules/.bin/tsc --project ./tsconfig.build.json --watch",
     "prepublish": "npm run build"
   },
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -24,7 +24,7 @@
     "dev:inspect": "node --inspect-brk -r ts-node/register src/scripts/dev.ts",
     "serve:inspect": "node --inspect-brk -r ts-node/register src/scripts/serve.ts",
     "build": "tsc --project ./tsconfig.build.json",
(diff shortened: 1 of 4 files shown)


packages/core/package.json
```diff
@@ -23,8 +23,8 @@
     "test:inspect": "node --inspect ../../node_modules/.bin/jest --watch --no-cache --runInBand",
     "dev:inspect": "node --inspect-brk -r ts-node/register src/scripts/dev.ts",
     "serve:inspect": "node --inspect-brk -r ts-node/register src/scripts/serve.ts",
-    "build": "../../node_modules/.bin/tsc --project ./tsconfig.build.json",
-    "build:watch": "../../node_modules/.bin/tsc --project ./tsconfig.build.json --watch",
+    "build": "tsc --project ./tsconfig.build.json",
+    "build:watch": "tsc --project ./tsconfig.build.json --watch",
```
michalczaplinski: Just updated this line as well to use `tsc` directly.

michalczaplinski (approved):
LGTM 👍

packages/core/package.json
luisherranz: Uh... sorry I didn't explain this better. Using `tsc` only works if you call it from the root, which is what lerna does. But it doesn't work when you are in the folder (`/packages/core`). When you are there, you need to do `../../node_modules/tsc`. But that doesn't work on Windows. So I changed only the one that is used by lerna.

For that reason, I used `tsc` on the script that lerna uses and the other one on the one that you use when developing `core`.

It's a mess, I know. I was trying to avoid installing `typescript` on all packages because it's already on the root, but maybe that's the best solution.

luisherranz: We have to either leave `build:watch` pointing to the root `tsc` or install `typescript` in the core.

michalczaplinski: oh, I see. I ll just revert that commit in that case.

michalczaplinski (approved):
👍 all good.

luisherranz:
Thanks @michalczaplinski !

PR #455, opened 2020-06-01
Add initial typedoc markdowns

luisherranz:
**What**:

I have added a new command `npx typedoc` to generate markdown files from our TSDocs and store them in the folder `/typedoc`.

**Why**:

So DevRels can see what has changed on each PR and update the documentation.

**How**:

Mainly with the `typedoc` package, the configuration in the `typedoc.json` file and two typedoc plugins, one for markdown (instead of html) and another for monorepos.

**Tasks**:

- [x] Community discussions (https://community.frontity.org/t/adopting-a-standard-way-to-document-frontity-apis/1662/10)
- [x] `typedoc` script to generate the files.
- [ ] GitHub action to check that typedocs are up to date (and fails if they are not).
- [ ] GitHub action to generate the typedocs and do a commit.

**Unrelated Tasks**

- TSDocs
- TypeScript
- Unit tests
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages

**Additional Comments**


luisherranz:
The things that are wrong right now:

- It doesn't work with nested properties.
- We have no way to mark which is public and which is not.
- Each time it builds, it changes all the files because it's adding the current commit hash in all the links. That makes it unusable for PR, because it will show unrelated changes for +50 files.

luisherranz:
I'm going to stop pursuing this approach for now. We'll revisit later.

NOTE: Please don't delete the branch.

PR #456, opened 2020-06-01
Replace `lodash-es` by `lodash`

### LINKED ISSUE #5: babel-plugin-lodash not working in node
`babel-plugin-lodash` works great in the client but for some reason, I can't make it work in node.

The import comes from this line: https://github.com/smooth-code/loadable-components/blob/master/packages/server/src/ChunkExtractor.js#L4

But it ends up with the whole lodash bundle inside `main.node.js`:
screen shot 2019-03-06 at 13 35 59

### LINKED ISSUE #405: Frontity doesn't work with Semantic UI [8pt]
**_Bug report_**

**Expected behavior**

Frontity should work with Semantic UI.

**Observed behavior**

It returns an error:

```
ModuleNotFoundError: Module not found: Error: Can't resolve 'lodash/fp/compact' in '/home/mpaccione/Projects/mpaccione-react/packages/mars-theme/node_modules/semantic-ui-react/dist/es/lib'
```

More info at https://community.frontity.org/t/how-to-add-semantic-ui-to-a-frontity-projectx/1634

- Start a new project.
- Add the [Semantic UI React](https://www.npmjs.com/package/semantic-ui-react) library.

**Possible solution**

I think we may be breaking the dependency of `lodash/fp` because, in order to do tree shaking of lodash, we have an alias of `lodash` to `lodash-es` in Webpack: https://github.com/frontity/frontity/blob/dev/packages/core/src/config/webpack/resolve.ts#L14

My feeling is that `lodash-es` doesn't have `loadsh-es/fp` and therefore we are breaking the dependency. To solve it we should find a way to do tree shaking of `lodash/fp`, or avoid doing the alias for `lodash/fp`, or maybe the main library (`lodash`) now supports tree shaking and the alias is not needed anymore.

DAreRodz:
**What**:

This PR removes `lodash-es` library and adds `lodash` and `babel-plugin-lodash`. As we are using Webpack ≥ 4, nothing more needs to be added in order to tree-shake the `lodash` library.

Fixes #405 

**Why**:

Right now, libraries that depend on `lodash` may not work if the use `lodash/fp` module (see #405).

**How**:

Just replace `lodash-es` by `lodash` (also the `@types` packages), and add `babel-plugin-lodash`.

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] Community discussions (update the following threads just before merge):
- https://community.frontity.org/t/how-to-add-semantic-ui-to-a-frontity-projectx/1634
- https://community.frontity.org/t/install-semantic-ui-react/1455
- https://community.frontity.org/t/external-component-libraries-using-emotion-not-rendering-styles-on-first-load/1837/25

**Unrelated Tasks**

- TSDocs
- TypeScript
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages


DAreRodz:
I tested it and `lodash` tree-shaking worked as expected. Using `semantic-ui-react` worked also, but apparently it's adding both the whole `semantic-ui-react` and `lodash` libraries in that case, I don't know why. :thinking:

.changeset/big-pugs-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Replace `lodash-es` by `lodash` library
```
luisherranz: Could you please be more descriptive here? Something like:

> Remove `lodash` alias to `lodash-es` to avoid problems with third-party libraries. Replace `lodash-es` with `babel-plugin-lodash` to still get tree shaking.

luisherranz (approved):
Perfect, thanks David.

I've added a suggestion for a more descriptive changeset, but the code is great.

I've checked, and it works now on the server as well, so we can close this issue too: https://github.com/frontity/frontity/issues/5

luisherranz:
> but apparently it's adding both the whole semantic-ui-react and lodash libraries in that case

Could you please elaborate on this?

DAreRodz:
> Could you please elaborate on this?

Oh, sorry. It is actually working. :+1: 

I thought it wasn't because the whole library (I mean `semantic-ui-react`) is added in development mode, but in production webpack only adds the imported libraries.

Everything OK.

luisherranz:
Awesome. Thanks @darerodz 🙂

PR #463, opened 2020-06-04
core package README.md

juanmaguitar:
**What**:

Adding README.md's of core package

**Why**:

So it has the proper info in NPM

**How**:

Creating the README.md's using on the proper folder

```
npx readme-md-generator -p ../../.github/README_TEMPLATE/package-readme.md -y
```

And then adding custom info on each README.md

**Tasks**:

**Unrelated Tasks**

- Code
- TSDocs
- TypeScript
- Unit tests
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages
- Community discussions

**Additional Comments**

### DIFF
--- a/.changeset/nasty-garlics-eat.md
+++ b/.changeset/nasty-garlics-eat.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": minor
+---
+
+README.md with full info of the package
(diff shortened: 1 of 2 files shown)


luisherranz (changes requested):
I think that the installation instructions are also confusing here because this package requires more things. I guess it's better to recommend people to use `npx frontity create` instead. What do you think?

luisherranz (approved):
Perfect now. Thanks @juanmaguitar!

PR #467, opened 2020-06-05
Dont debug in production

luisherranz:
**What**:

It makes sure that `state.frontity.debug` is not activated in production.

**Why**:

People may forget to turn it false again before deploying their site in production.

**How**:

It doesn't overwrite the `state.frontity.debug` value from the default (which is `false`) if we are in production.

They can still debug production mode using the console:

```bash
> frontity.state.frontity.debug = true;
```

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] Changeset 

**Unrelated Tasks**

- TSDocs
- TypeScript
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages
- Community discussions

### DIFF
--- a/.changeset/wet-pumpkins-sing.md
+++ b/.changeset/wet-pumpkins-sing.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Do not allow the new `state.frontity.debug` flag to be `true` in production.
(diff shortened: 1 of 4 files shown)


michalczaplinski (approved):
yup, LGTM 👍

PR #471, opened 2020-06-05
Fix analytics component

### LINKED ISSUE #460: Pageviews not sent when navigating to a link with the same title
**_Bug report_**

Pageview are not sent when you navigate from one page to another with the same title if the second one has the property `isReady` equals `true` (i.e. it was already fetched).

This happen, for example, if you use the `@frontity/head-tags` package, as it set the same title for different archive pages.

**Expected behavior**

To send a pageview for the page you are navigating to.

**Observed behavior**

No pageview is sent.

- Go to frontity.org/blog
- Click on the next page link (frontity.org/blog/page/2)
- Go back to the previous page
- A pageview shoud have been sent when going back, but it hasn't.

**Possible solution**

I was looking at the code and it looks like we can remove the title check that we are doing in the `roots.analytics` component as it's kind of redundant code.

Simply remove the next lines

https://github.com/frontity/frontity/blob/bb26a7ce8075b623ba6daf9af68306844193f9da/packages/analytics/src/components/index.tsx#L14-L15

and the comparison between `prevTitle` and `title` here, along with the `setPrevTitle(title)` code.

https://github.com/frontity/frontity/blob/bb26a7ce8075b623ba6daf9af68306844193f9da/packages/analytics/src/components/index.tsx#L34-L36

DAreRodz:
**What**:

It fixes the `@frontity/analytics` root component that wasn't sending pageviews when navigating to a new link with the same title.

I also take the opportunity to refactor the type definitions for the `@frontity/analytics` package and match what was discussed here in the community: [TypeScript: use two different types](https://community.frontity.org/t/typescript-use-two-different-types/1681)

Fixes #460 

**Why**:

To fix a bug that's causing some pageviews not to be sent.

**How**:

Just removing some innecessary checks on a page title that we were doing in the `@frontity/analytics` root component.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] End to end tests

**Unrelated Tasks**

- TSDocs
- Unit tests
- TypeScript tests
- Update starter themes
- Update other packages
- Community discussions

### DIFF
--- a/.changeset/great-lobsters-bow.md
+++ b/.changeset/great-lobsters-bow.md
@@ -1,0 +1,5 @@
+---
+"@frontity/analytics": patch
+---
+
+Fix a bug that was causing pageviews not to be sent if the previous page has the same title.
--- a/e2e/integration/analytics.spec.js
+++ b/e2e/integration/analytics.spec.js
@@ -9,6 +9,11 @@ describe("Analytics package", () => {
     title: "Some Post Title",
   };
 
+  const pageviewSomeOtherPost = {
+    page: "/some-other-post/",
+    title: "Some Post Title",
+  };
+
   const someEvent = {
     event: "some event",
     payload: { content: "some content" },
@@ -35,7 +40,9 @@ describe("Analytics package", () => {
     cy.get("button#change-link-post-2").click();
     cy.go("back");
     getAnalytics("pageviews").its(1).should("deep.equal", pageviewSomePost);
-    getAnalytics("pageviews").its(2).should("deep.equal", pageviewSomePost);
+    getAnalytics("pageviews")
+      .its(2)
+      .should("deep.equal", pageviewSomeOtherPost);
     getAnalytics("pageviews").its(3).should("deep.equal", pageviewSomePost);
   });
 
--- a/e2e/integration/analytics.spec.js
+++ b/e2e/integration/analytics.spec.js
@@ -1,0 +1,59 @@
+describe("Analytics package", () => {
+  const pageviewHome = {
+    page: "/?name=analytics",
+    title: "Homepage Title",
+  };
+
+  const pageviewSomePost = {
+    page: "/some-post/",
+    title: "Some Post Title",
+  };
+
+  const someEvent = {
+    event: "some event",
+    payload: { content: "some content" },
+  };
+
+  const getAnalytics = (data) =>
+    cy.window().its("frontity").its("state").its("testAnalytics").its(data);
+
+  beforeEach(() => {
+    cy.visit("http://localhost:3001?name=analytics");
+  });
+
+  it("should have sent the first pageview", () => {
+    getAnalytics("pageviews").its(0).should("deep.equal", pageviewHome);
+  });
+
+  it("should sent a pageview if the page changes", () => {
+    cy.get("button#change-link").click();
+    getAnalytics("pageviews").its(1).should("deep.equal", pageviewSomePost);
+  });
+
+  it("should sent a pageview if the page changes and title is the same", () => {
+    cy.get("button#change-link").click();
+    cy.get("button#change-link-post-2").click();
+    cy.go("back");
+    getAnalytics("pageviews").its(1).should("deep.equal", pageviewSomePost);
+    getAnalytics("pageviews").its(2).should("deep.equal", pageviewSomePost);
+    getAnalytics("pageviews").its(3).should("deep.equal", pageviewSomePost);
+  });
+
+  it("should sent pageviews when going back or forward", () => {
+    cy.get("button#change-link").click();
+    cy.go("back");
+    getAnalytics("pageviews").its(2).should("deep.equal", pageviewHome);
+
+    cy.go("forward");
+    getAnalytics("pageviews").its(3).should("deep.equal", pageviewSomePost);
+  });
+
+  it("should send events", () => {
+    cy.get("button#send-event").click();
+    cy.get("button#send-event").click();
+    cy.get("button#send-event").click();
+    getAnalytics("events").its(0).should("deep.equal", someEvent);
+    getAnalytics("events").its(1).should("deep.equal", someEvent);
+    getAnalytics("events").its(2).should("deep.equal", someEvent);
+  });
+});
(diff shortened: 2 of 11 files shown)


luisherranz (changes requested):
While trying to simplify the types, we discovered a bug with the types of `@frontity/google-analytics`.

This was the conversation: https://www.youtube.com/watch?v=WKTXO9ziZ6g

DAreRodz:
@luisherranz, I've changed my mind. Could you push your changes?

I think I've made it work but I don't know what I did differently.  :sweat_smile:

luisherranz:
Done 👍

DAreRodz:
Thanks, I'll take a look.

DAreRodz:
I didn't want to write such a long comment. I hope everything is understandable and clear. Sorry if it's not, @luisherranz, @michalczaplinski. 😅

**TL;DR**

- Do not use generics for exported code that is mean to be imported by other packages
- Avoid exporting types that are only used internaly (the `[key: string]` part)
- When an action could be reimplemented, try to pass the less possible amount of packages to that action type so it's more difficult to have a type conflict.

---

Ok, I guess the problem we had with TypeScript here was an edge case related to the way we implemented the analytics packages. Also, I don't clearly understand why TypeScript doesn't complain when you define the types this way but it does when you try to implement code using those types.

Well, let me try to explain it. 😄

The current implementation of the analytics package types has defined two optional `sendPageview` and `sendEvent` actions for any namespace (because we don't know the name of all analytics packages).

https://github.com/frontity/frontity/blob/5baa5094124f27aa711fc667227b66bfd0be5599/packages/analytics/types.ts#L24-L27

These types are exported in the `Analytics` interface, which is a generic type that accepts one argument for all the packages that is passed down to actions as the `Pkgs` argument you see above. By default, `Pkgs` is:

https://github.com/frontity/frontity/blob/5baa5094124f27aa711fc667227b66bfd0be5599/packages/analytics/types.ts#L36

The problem start when we define the `GoogleAnalytics` types.

Looking at the code at it is right now, we are using `Analytics` without passing any argument when defining `actions` types

https://github.com/frontity/frontity/blob/5baa5094124f27aa711fc667227b66bfd0be5599/packages/google-analytics/types.ts#L34-L39

The `Packages` type we are passing to actions is:

https://github.com/frontity/frontity/blob/5baa5094124f27aa711fc667227b66bfd0be5599/packages/google-analytics/types.ts#L42

At this point, I'm going to write some snippets to clarify what happen. These are the final types

```ts
type AnalyticsPkgs = Analytics & Router & Source; 
type GoogleAnalyticsPkgs = GoogleAnalytics & Analytics<GoogleAnalytics>;

interface GoogleAnalytis {
  actions: {
    analytics: {
      sendPageview: Action<AnalyticsPkgs, Pageview>; 
      sendEvent: Action<AnalyticsPkgs, Event>; 
    };
    googleAnalytics: { 
      sendPageview: Action<GoogleAnalyticsPkgs, Pageview>; 
      sendEvent: Action<GoogleAnalyticsPkgs, GoogleAnalyticsEvent>; 
    };
    [key: string]: {
      sendPageview: Action<AnalyticsPkgs, Pageview>; 
      sendEvent: Action<AnalyticsPkgs, Event>; 
    };
  }; 
}
```

As it can be seen in the code above, `sendPageview` and `sendEvent` actions inside `googleAnalytics` namespace would receive a state derived from `GoogleAnalyticsPkgs`, which conflicts with the state received by actions in `[key: string]`, which is derived from `AnalyticsPkgs`. They conflict because one has types from `GoogleAnalytics` and the other has types from `Router` and `Source`.

The error that TypeScript gives (and I don't understand why it's given when you implement the code and not when you define the types) is

```
Property 'googleAnalytics' is incompatible with index signature.
    Types of property 'sendPageview' are incompatible.
        Types of parameters '__0' and '__0' are incompatible.
            Types of property 'state' are incompatible.
                Property 'googleAnalytics' is missing in type...
```

If we would instead have define the actions passing `GoogleAnalytics` packages as argument to the generic `Analytics` type like this:

```ts
interface GoogleAnalytics extends Package {
  actions: Analytics<Packages>["actions"] & {
    googleAnalytics: {
      sendPageview: Action<Packages, Pageview>;
      sendEvent: Action<Packages, GoogleAnalyticsEvent>;
    };
  };
}
```

The resultant types would have been

```ts
type AnalyticsPkgs = Analytics & Router & Source; 
type GoogleAnalyticsPkgs = GoogleAnalytics & Analytics<GoogleAnalytics>;

interface GoogleAnalytis {
  actions: {
    analytics: {
      sendPageview: Action<GoogleAnalyticsPkgs, Pageview>; 
      sendEvent: Action<GoogleAnalyticsPkgs, Event>; 
    };
    googleAnalytics: { 
      sendPageview: Action<GoogleAnalyticsPkgs, Pageview>; 
      sendEvent: Action<GoogleAnalyticsPkgs, GoogleAnalyticsEvent>; 
    };
    [key: string]: {
      sendPageview: Action<GoogleAnalyticsPkgs, Pageview>; 
      sendEvent: Action<GoogleAnalyticsPkgs, Event>; 
    };
  }; 
}
```

Types are fine now, but after trying this it's still failing! Why?

Because the `analytics` code that we are importing here

https://github.com/frontity/frontity/blob/1fc5c4efe1ed737d75ae9c126300514e92674422/packages/google-analytics/src/index.ts#L1

and its actions that we are adding here

https://github.com/frontity/frontity/blob/1fc5c4efe1ed737d75ae9c126300514e92674422/packages/google-analytics/src/index.ts#L14

have the default types for `Packages` (i.e. `Analytics & Router & Source`), and so it receives a different `state`.

This is the error now:

```
The types of 'analytics.sendPageview' are incompatible between these types.
    Types of parameters '__0' and '__0' are incompatible.
        Types of property 'state' are incompatible.
            Type 'ResolveState<...>' is missing the following properties from type 'ResolveState<...>': router, source
```

---

These are the conclusions I've extracted:

- Do not use generics for exported code that is mean to be imported by other packages.
- Avoid exporting types that are only used internaly (the `[key: string]` part).
- When an action could be reimplemented, try to pass the less possible amount of packages to that action type so it's more difficult to have a type conflict.

michalczaplinski:
@DAreRodz  Just testing integromat because the scenario stopped working. Let me know if you get it please 🙁

luisherranz (approved):
Hey, sorry it took me so long to do this review. Being a critical bug I should've given it much more priority.

@SantosGuillamot feel free to merge and do a release.

PR #458, opened 2020-06-03
Add CLI env variables for `dev`, `build` and `serve` commands

DAreRodz:
**What**:

Add these env variables to the `dev`, `build` and `serve` commands:

**`dev`**

- target: `FRONTITY_DEV_TARGET`
- port: `FRONTITY_DEV_PORT`
- https: `FRONTITY_DEV_HTTPS`
- production: `FRONTITY_DEV_PRODUCTION`
- publicPath: `FRONTITY_DEV_PUBLIC_PATH`
- dont-open-browser: `FRONTITY_DEV_DONT_OPEN_BROWSER`

**`build`**

- target: `FRONTITY_BUILD_TARGET`
- development: `FRONTITY_BUILD_DEVELOPMENT`
- publicPath: `FRONTITY_BUILD_PUBLIC_PATH`

**`serve`**

- port: `FRONTITY_SERVE_PORT`
- https: `FRONTITY_SERVE_HTTPS`

Closes #394 

**Why**:

To allow users to pass CLI options through env variables.

**How**:

Following the implementation proposal in [Change publicPath](https://community.frontity.org/t/change-publicpath/1461/5).

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] Community discussions: https://community.frontity.org/t/change-publicpath/1461/8
- [x] Changeset (with link to its [Feature Discussion](https://community.frontity.org/c/33) if it exists)
- [X] Link to PR in [Documentation](https://github.com/frontity/gitbook-docs/) → https://github.com/frontity/gitbook-docs/pull/93

**Unrelated Tasks**

- TSDocs
- TypeScript
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages

PR #473, opened 2020-06-05
Update the test coverage collection 

michalczaplinski:
At the moment, the test coverage is only collected from the files that are required / used by our tests. The files that are not tested at all do no count towards the total number of lines covered.

This results in 2 problems:
- Our coverage is artificially inflated (the ~90% coverage was not accurate, it's actually closer to 70%, which is still not terrible actually)
- We run into a problem ([reported by @DAreRodz](https://github.com/frontity/frontity/pull/458#issuecomment-638822172)) where if we add a test case for a file that we previously untested, the code coverage actually goes down, even though logically _more_ of our code is being tested 🙂 

---

**This PR** fixes this by collecting the coverage from all JS / TS files in packages with some exceptions like jest configs. Collecting the coverage only takes place when jest is run with `--coverage` so updating the global config does not affect tests that are run locally

I've additionally updated the "patch" coverage to 100% - this means that all new code should have 100% coverage. We can adjust that in the future if it ends up not being practically feasible sometimes.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,16 @@
+# Round the coverage up, up to the 2nd decimal point. E.g
+# 89.0199 % => 89.02%
 coverage:
   precision: 2
   round: up
 
+  # The `patch` target of 100% means that all new code must be fully tested (100%)
+  status:
+    patch:
+      default:
+        target: 100%
+
+# Make the layout of the comment posted on the PR by the codecov bot a bit nicer
 comment:
   layout: "files"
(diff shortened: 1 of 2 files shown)


luisherranz (approved):
It looks good to me Michal.

Is it ready for merge?

michalczaplinski:
> It looks good to me Michal.
> 
> Is it ready for merge?

yes it was 😊 

Will merge this.

PR #475, opened 2020-06-08
The Slot component

### LINKED ISSUE #395: Create the Slot component [3pt]
**Feature:** [Slot and Fill](https://community.frontity.org/t/slot-and-fill/895).
**User Stories**:
  - Add additional components to the theme.
  - Substitute certain components of the theme.

[**Sprint Card**](https://github.com/orgs/frontity/projects/4#card-36476149)

michalczaplinski:
**What**:

Add the `Slot` component for the "Slot and Fill" feature.

Feature Discussion: https://community.frontity.org/t/slot-and-fill/895/

**Tasks**:
- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] Link to PR in Documentation: https://github.com/frontity/gitbook-docs/pull/90
- [x] Community discussions: https://community.frontity.org/t/slot-and-fill/895/19

**Unrelated Tasks**
- End to end tests
- TypeScript tests
- Update other packages
- Update starter themes


packages/types/src/package.ts
```diff
@@ -141,7 +141,7 @@ export interface Package {
      */
     fills?: {
       [namespace: string]: {
-        [name: string]: React.ReactType;
+        [name: string]: React.ElementType;
```
michalczaplinski, line 144: `ReactType` is deprecated, we should rename it in other packages and other places, too but that's out of the scope for this PR :)

packages/components/slot.tsx
```diff
@@ -0,0 +1,33 @@
...
+ */
+const Slot: SlotType = ({ name, state, children, ...slotProps }) => {
+  // Get the data, either from props or the current link.
+  const data = slotProps.data || state.source.get(state.router.link);
+
+  // Get the fills for this name.
+  const fills = useFills(name);
+
+  return (
+    // Returning a <Fragment/> because returning an array of components results in a typescript error
```
michalczaplinski: This is the issue that seems to appears otherwise:

Screenshot 2020-06-08 at 18 06 53

packages/frontity/src/utils/slot-and-fill/__tests__/slot.tests.tsx
```diff
@@ -0,0 +1,147 @@
...
+      <Provider value={store}>
+        <Slot name="slot 3" whateverProp={"Data passed as prop"} />
+      </Provider>
+    );
+
+    expect(app.toJSON().children[0]).toEqual("Data passed as prop");
+    expect(app.toJSON()).toMatchSnapshot();
+  });
+
+  it("should override the prop provided by the theme creator with the one given by the user", () => {
```
michalczaplinski, line 126: To the reviewer: I believe this is the correct behaviour, but just pointing it out here because it's not specified in the proposal.

michalczaplinski:
~~@luisherranz just a minor detail when you review it - I've just realised now that I forgot export the component from the index file!~~

~~Will fix this tomorrow ofc~~

done.

packages/frontity/src/utils/slot-and-fill/__tests__/slot.tests.tsx
DAreRodz: Yup, this totally makes sense. :+1:

packages/frontity/src/utils/slot-and-fill/slot.tsx
```diff
@@ -0,0 +1,51 @@
...
+   */
+  [key: string]: any;
+}
+
+type SlotType = React.FC<Connect<Package, SlotProps>>;
+
+/**
+ * A React component that renders a Fill to fulfill a Slot and Fill pattern.
+ */
+const Slot: SlotType = ({ name, state, children, ...slotProps }) => {
```
DAreRodz: The properties `actions` and `libraries` are being included in `slotProps` and passed to `Fill` as props later.

This is wrong, isn't it? :thinking:

I guess the way of passing `state`, `actions` and `libraries` would be to connect the `Fill` component before exporting it.

DAreRodz (changes requested):
Everything is great, @michalczaplinski!

I saw a problem tho with the props that are passed to the `Fill` component, check it out.

packages/frontity/src/utils/slot-and-fill/slot.tsx
michalczaplinski: > I guess the way of passing state, actions and libraries would be to connect the Fill component before exporting it.

yeah, you're right that this is probably not what we want. 

Those `...slotProps` are there so that the creator of the `Slot` could pass some props by default to the Fill, not in order to pass the `state`, `actions` or `libraries` down.

I guess we can "filter out" the props like: 

```
const Slot: SlotType = ({ name, state, children, libraries, actions, ...slotProps }) => {
	// We destructure the `libraries` and `actions` but not use them
	...
}
```

DAreRodz: I guess so. Yeah, think it's a good idea to filter out the props like that. :+1:

michalczaplinski:
Looks like the vercel deployment is failing due to permissions on this and on other PRs.

packages/frontity/src/utils/slot-and-fill/slot.tsx
```diff
@@ -28,7 +28,15 @@ type SlotType = React.FC<Connect<Package, SlotProps>>;
...
  * A React component that renders a Fill to fulfill a Slot and Fill pattern.
  */
-const Slot: SlotType = ({ name, state, children, ...slotProps }) => {
+const Slot: SlotType = ({
+  name,
+  state,
+  children,
+  libraries, //  These props are unused on purpose.
+  actions, ////  We're only destructuring them so that we can filter them out from `...slotProps`
+  roots, //////  which are passed to the Fill component
```
DAreRodz: Wow, this line makes me think that maybe `connect` is wrong:

Here, the types for `Connect`. The `name`, `roots` and `fills` props are removed:

https://github.com/frontity/frontity/blob/84cd8aeaa33bb8a34a8de499e7476eac80bb544f/packages/types/src/connect.ts#L5-L11

But here in `connect` all frontity props are passed, including `roots` and `fills`:

https://github.com/frontity/frontity/blob/2e9d6137c5d9c5e6cd5aac6b3b06bd600fba5b80/packages/connect/src/connect.js#L84-L85

https://github.com/frontity/frontity/blob/2e9d6137c5d9c5e6cd5aac6b3b06bd600fba5b80/packages/connect/src/connect.js#L116-L117

What should be the correct behavior for connect? Not to pass those props? Or add them to the `Connect` type and update all docs?

cc: @michalczaplinski @luisherranz

michalczaplinski: To me it seems that we should pass all those props: `state`, `actions` , `name`, `roots`. So, I'm not sure why the types are erased in `Connect`

Also I think that `fills` was removed, in favour or `state.fills`, correct?

CC: @luisherranz

michalczaplinski:
I think we should still add a usage example to our themes.

This should make writing the documentation easier for @mburridge as well.

Let's not merge this yet, I've converted it back to draft

packages/frontity/src/utils/slot-and-fill/slot.tsx
luisherranz: Hey guys, just curious: why don't you use `useConnect` with `injectProps: false`? It was created specifically to avoid this problem:

```typescript
const Slot: React.FC<SlotProps> = ({ name, children, ...slotProps }) => {
  const { state } = useConnect<Packages>();

  // Get the data, either from props or the current link.
  const data = slotProps.data || state.source.get(state.router.link);

  // Get the fills for this name.
  const fills = useFills(name);

  return (
    <>
      {fills.length > 0
        ? fills.map(({ Fill, props, key }) => (
            <Fill key={key} name={key} data={data} {...slotProps} {...props} />
          ))
        : children}
    </>
  );
};

export default connect(Slot, { injectProps: false });
```

packages/frontity/src/utils/slot-and-fill/slot.tsx
luisherranz: You're right that roots should not be passed down to connected components and fills is deprecated. I want to review the types once we migrate to RES.

But for this case, please use `useConnect`: https://github.com/frontity/frontity/pull/475#discussion_r440256221

packages/frontity/src/utils/slot-and-fill/slot.tsx
michalczaplinski: You're right, this is the cleanest way :)

michalczaplinski:
I'm not sure how to add the example for this to our themes because lerna does not play nice with `packages` inside of the `examples/<theme-name>` folder.

I think we can do this in a future PR. I ll mark this as ready again.

DAreRodz (approved):
It's great, Michal. Approved!

DAreRodz:
> I think we can do this in a future PR. I ll mark this as ready again.

Should we move the "Update starter themes" task to Unrelated, then?

michalczaplinski:
> > I think we can do this in a future PR. I ll mark this as ready again.
> 
> Should we move the "Update starter themes" task to Unrelated, then?

indeed 🙂

PR #476, opened 2020-06-09
Fixes archive not being fetched when post type is loaded from SSR

orballo:
**What**:

Fixes a bug where the archive is not requested if a post type is loaded from SSR.

**Why**:

It was a bug.

**How**:

Removed the condition where the link must belong to an archive to be requested. This is wrong because before requesting it that value of `data.isArchive` will always evaluate as false.

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] Changeset

**Unrelated Tasks**
-  TSDocs
-  TypeScript
-  End to end tests
-  TypeScript tests
-  Update starter themes
-  Update other packages
-  Link to PR in [Documentation](https://github.com/frontity/gitbook-docs/)
-  Community discussions

**Additional Comments**

### DIFF
--- a/.changeset/moody-windows-love.md
+++ b/.changeset/moody-windows-love.md
@@ -2,4 +2,4 @@
 "@frontity/hooks": patch
 ---
 
-Fixes a bug where archive is not requested if loading post type from SSR
+Fixes a bug on the `usePostTypeInfiniteScroll` where the archive is not requested if we are loading the post type from SSR.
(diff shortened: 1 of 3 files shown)


michalczaplinski:
Assigned it to me because I'm the supporter this week.

michalczaplinski (commented):
I think I don't have all the context to review this efficiently - I think @luisherranz should have a look 🙂

luisherranz (approved):
Thanks @orballo and sorry for the delay.

luisherranz:
I have published a new beta version with this change: `@frontity/hooks@2.1.0-infinite-scroll-beta.1`.

PR #479, opened 2020-06-10
Proper dependencies relationship frontity

juanmaguitar:
**What**:

Updated dependencies relationship

**Why**:

So it avoid possible issues on build or deployment
So the proper relationship between packages can be tracked in places like this: https://www.npmjs.com/browse/depended/@frontity/connect

**How**:

Changing packges from "devDependencies" to "dependencies"

**Tasks**:

- [X] Changeset 

**Unrelated Tasks**

- Code
- TSDocs
- TypeScript
- Unit tests
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages
- Community discussions

### DIFF
--- a/.changeset/twenty-mails-hammer.md
+++ b/.changeset/twenty-mails-hammer.md
@@ -1,0 +1,5 @@
+---
+"frontity": patch
+---
+
+Proper dependencies relationship.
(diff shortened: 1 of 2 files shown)


luisherranz:
I am sorry Juanma but neither `@frontity/core` nor `@frontity/connect` can be `dependencies` of `frontity`, they must be `devDependencies`. The first one because we want to avoid the download of all the core dependencies when people do `npx frontity create` and the second one to avoid a bug when people update Frontity, having duplicate versions of `@frontity/connect` installed. `@frontity/core` might be a `peerDependency` of `frontity` but npm would throw a warning when doing `npx frontity create` so I didn't like that either.

PR #495, opened 2020-06-12
Init wp-comments package and add action to create new comments

### LINKED ISSUE #439: Add action to create new comments [13pt]

DAreRodz:
**What**:

Implements the `comments.submit` action described here: https://community.frontity.org/t/wordpress-comments-package/1267

Closes #439 

**Why**:

To allow Frontity users to publish new comments to their WP sites from Frontity.

**How**:

Following the implementation proposal: https://community.frontity.org/t/wordpress-comments-package/1267/34

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] Link to issue in [Documentation](https://github.com/frontity/gitbook-docs/wiki/Code-Releases): https://github.com/frontity/docs/issues/174
- [x] Community discussions (https://community.frontity.org/t/wordpress-comments-package/1267/50)

**Unrelated Tasks**

- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages


DAreRodz:
@mburridge I assign you this PR to open the PR for the documentation (you are supporter this week :grin:).

Once the documentation PR is open, please add the link to this PR description and check the task.

packages/wp-comments/src/__tests__/actions.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      url: "https://frontity.org",
+      postId: 60,
+    });
+
+    // Check the comment was not sent.
+    expect(fetch).not.toHaveBeenCalled();
+
+    // Check that a warning message was shown instead.
+    expect(warn.mock.calls).toMatchSnapshot();
+  });
```
michalczaplinski: Could we also add at least one test case that is negative? E.g.: Posting a comment for a postId that does not exist?

michalczaplinski (changes requested):
Looks great 👍 Nice and clean code 🙂 

Just a couple of comments in addition to what I mentioned:
- Do you want to include an example for one of our themes or will do that in another PR?
- It was not documented in the FD, but I'm assuming that we plan to update the implementation so that it will work with wp.com and with the theme bridge afterwards, right?
- I guess the e2e tests are not going to work until we have the full e2e testing with WP instances. Note to self: Let's remember to test this

packages/wp-comments/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@frontity/wp-comments",
+  "version": "0.1.0",
```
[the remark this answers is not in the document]
DAreRodz: I don't think so, when you want to release the first version, you set `0.X.X` to bump it to the `1.0.0`.

packages/wp-comments/src/__tests__/actions.test.ts
DAreRodz: Oh, yeah, you are right.

I was thinking about this during the weekend. As actions don't return anything, I'm not sure what would be the correct way to inform that the request was succesful/unsuccesful.

Changing some prop in the state, maybe? Or throwing an error? :shrug:

packages/wp-comments/package.json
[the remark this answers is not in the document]
michalczaplinski: oh, I see!

DAreRodz:
I'm going to change the PR status to draft, and block the issue until we decided how to implement the result for the `create` action. https://github.com/frontity/frontity/pull/495#discussion_r440126118

packages/wp-comments/src/__tests__/actions.test.ts
luisherranz: I have added my opinion in https://community.frontity.org/t/wordpress-comments-package/1267/36

luisherranz:
I have added my opinion in https://community.frontity.org/t/wordpress-comments-package/1267/36

DAreRodz:
I just saw it, thanks! :+1:

michalczaplinski (changes requested):
Great work @DAreRodz 👍 👍 👍  

I have just pushed some fixes for very minor typos.

The only comment that I have is that I think we could avoid having both `isApproved` and `isUnapproved` booleans in the `forms[id].submitted`.

As far as I can tell, the form state is a state machine that looks like this: 

image

so, if we assume that the starting state **always** has `isPending: true`, then we can describe all the possible states with just the `isPending` and `isApproved` flags (plus the `isError` for errors)

However, if this assumption does not hold, meaning that there is a case when we could have:
```
isPending: false,
isError: false,
isApproved: false,
```
**before** the form is sent, then we'd have a problem because according to the diagram that I drew, this corresponds to the *unapproved* case, which would be wrong.

I think it should always be the case that `isPending: true` in the initial state. But in case that I'm wrong, I'm still against having 2 booleans with opposite names (`isApproved` and `isUnapproved`) because I think it's quite confusing.

E.g. should the user check `if (!isApproved) {}` or `if (isUnapproved) {}` - it's not clear and it's easy to make a mistake where both are false or both true and that should, of course, be impossible 😅 . In that case I'd rename the `isUnapproved` to `isReceived` or something similar, to indicate that the comment was received by WP, is not pending anymore, but is not approved yet.

packages/wp-comments/src/__tests__/mocks/helpers.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * It instantiates a `node-fetch` response object using the passed information
+ * with the standard `Response` type.
+ *
+ * @param body - Body of the `Response` object.
+ * @param init - Other `Response` properties.
+ *
+ * @returns `Response` object.
+ */
+export const mockResponse = (body, init?: ResponseInit): Response =>
```
michalczaplinski, line 14: This is just a comment for a future improvement: We could have a sort of `test-utils` or `shared` package like react.js has because there already exists the same function in the wp-source (twice in fact, once in helpers, other in populate.tests.ts 😅 )

packages/wp-comments/src/index.ts
```diff
@@ -0,0 +1,153 @@
...
+        // Check first if the connected source is a WP.com site.
+        if (state.source.isWpCom) {
+          return warn(
+            "Sending comments to a WordPress.com site is not supported yet."
+          );
+        }
+
+        // Return if a submission is pending.
+        if (state.comments.forms[postId]?.submitted?.isPending) {
+          return warn(
```
nicholasio, line 23: Should we return false so that the caller knows it didn't work? Or are we just relying on state update/re-rendering?

packages/wp-comments/src/index.ts
```diff
@@ -0,0 +1,153 @@
...
+          // Get first the body content.
+          const body = await response.text();
+
+          // Set error properties
+          form.submitted.isPending = false;
+          form.submitted.isError = true;
+
+          // If the body is empty `postId` is invalid;
+          // `author` or `email` are invalid otherwise.
+          form.submitted.errorMessage = !body
```
nicholasio, line 91: This is more of a side question, are there any plans on supporting internationalization? Having something like `__()` to localize static strings

This is something that can be handled at the theme level but there are probably a bunch of hardcoded messages across the codebase and having a standard way of handling this kind of thing would be useful.

nicholasio (commented):
nice work.

packages/wp-comments/src/index.ts
DAreRodz: We are relying on state update/re-rendering here because actions don't return anything. We try to follow the flux pattern so what actions only do is to modify the state. This is how actions work in Frontity by design.

Well, for asynchronous actions, they actually return a Promise that resolves once the action has finished, but it's just an empty Promise.

packages/wp-comments/src/index.ts
DAreRodz: Yeah, it would be useful!

It is not something in our roadmap right now but you can create a topic in the [Feature Discussions](https://community.frontity.org/c/feature-discussions/33) section for that. :slightly_smiling_face:

luisherranz: Oh, yes, sorry. A standard internationalization for Frontity packages is on the roadmap. Just no FD yet. I have some ideas though :)

DAreRodz:
I just renamed `isUnapproved` by `isOnHold`, trying to use the same names that the WordPress REST API returns for the `status` field in comments, which are
- `approved` - for published comments
- `hold` - for unmoderated comments
- `spam` - for those marked as spam
- `trash` - for those that were deleted

@michalczaplinski, you can review it again. :blush:

DAreRodz:
I also added content in the README.md file and changed the changeset to release a beta version (0.x) instead of a stable one (1.0), the same we did for our analytics packages.

Regarding the README, I used the same template we used for other packages but I didn't add a link to the documentation yet (it doesn't exist). I guess we have to first create a page for this package in https://github.com/frontity/docs.

michalczaplinski (approved):
Great work David 👍

PR #499, opened 2020-06-16
Add an extra check to update README.md in github PR template

michalczaplinski:
Since we have proper README.md files now, we should keep them up to date when adding new features!

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -45,6 +45,7 @@ merge of your pull request!
 - [ ] Unit tests
 - [ ] End to end tests
 - [ ] TypeScript tests
+- [ ] Update package README.md
 - [ ] Update starter themes
 - [ ] Update other packages
 - [ ] Link to PR in [Documentation](https://github.com/frontity/gitbook-docs/)


luisherranz:
Hmm... interesting.

- Shouldn't we just add a link to the relevant docs in order to avoid the need to maintain two places up to date?
- Is there a clear benefit for the extra work that means keeping up to date the latest features in both the docs and the readme?

My opinion is that we should try to avoid adding more work to our workflows because we are still a very small team but @juanmaguitar @mburridge, what do you think? Maybe we need to make the link to the docs on the READMEs even more prominent?

michalczaplinski:
I don't really know what the devrel team has planned for those README's .

But I agree that we should just concentrate the work in one place 🙂

The way I see it was that: README.md files should just have a 1-2 sentence description and link to the docs. So, it should happen very rarely that we need to update those READMEs (only if we change something very essential a package that changes its whole purpose). But it can still be necessary, so that's why I made this change 🙂 .

@juanmaguitar Since you were the one updating the READMEs for all packages, can you let us know if that agrees with your vision?

juanmaguitar:
@michalczaplinski @luisherranz I wouldn't add this extra check:
- `README.md` are written in the most generic way possible linking to the docs for specific details (with the exception of a very brief introduction of how to use the specific package) so they shouldn't change very often
- In the ["Code Releases" workflow description](https://github.com/frontity/gitbook-docs/wiki/Code-Releases) the DevRel team assumes...

> DevRel team will be in charge of opening the proper docs PR in the documentation repository (and all extra tasks or actions required)

So, once we have understood the purpose of each PR we'll decide the actions to be taken so this change is properly understood by the community (minimum action will be update of the docs, but maybe also a DevRel Talks video, update of the README, a blog post.... or any other action we consider is needed to properly communicate this)

michalczaplinski:
All right, thanks @juanmaguitar I'm closing this in that case :)

PR #501, opened 2020-06-18
Remove ts-node from `frontity` and move it to `core` only for `dev and `build`

### LINKED ISSUE #355: Get rid of ts-node in our CLI [5pt]
**Description**

Most commands on our CLI are simply dev tools, so performance is not important. But `frontity serve` is the command that will be run in production, so it's critical that it is a lean as possible. Right now, we are using `ts-node` internally in our whole CLI, which is fine for most commands, but it's unnecessary for `frontity serve` and it degrades its performance.

We added `ts-node` to be able to use the `frontity.settings.js` with ES modules (import & export) and to be able to use `frontity.settings.ts`. Those two files are outside Webpack.

It may also cause bugs, like the described in here: https://community.frontity.org/t/error-did-you-forget-to-run-frontity-build/1426/11

**Steps to reproduce the issue**

- Clone https://github.com/luisherranz/frontity-react-slick
- Run `npx frontity build`
- Run `npx frontity serve`

**Possible solution**

I think the best solution is to move `ts-node` inside the commands that need it: `dev` and `build` (they belong to `core`).

Apart from that, it'd be interesting to see if thanks to `ts-node`, we could get rid of the build step of `core`. Of course, that means that `ts-node` needs to be required a file before using commonjs, and that the code of `serve` must also be commonjs (no ES modules).

michalczaplinski:
**What**:

Remove ts-node from `frontity` and move it to `@frontity/core`. The is only for the `dev` and `build` commands`

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript

**Unrelated Tasks**
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages
- Community discussions
- Unit tests

**Additional Comments**

I have looked into removing the build step from `core` as part of this PR as well. This would be possible because both `dev` & `build` are using ts-node. The `serve` command could then be converted from TS to JS and use commonjs. 

I feel like it's not worth it because both `serve` and `dev` share some code in `createServer` which would have to be converted to JS and while we gain lack of build step, we lose type safety.


michalczaplinski:
@luisherranz @DAreRodz  I'm not sure about the best approach to testing `build` & `dev` 🤔 

I've started on a creating a test case using `mock-fs` in this PR and I think that it might work, but I think that we will have to mock all the "frontity" dependencies in node_modules, which is not ideal. 

For example, in `mars-theme-example`, we have all these symlinked deps:

Screenshot 2020-06-18 at 21 01 11

So, we'll have to mock each one of them with https://github.com/tschaub/mock-fs#mocksymlinkproperties. I think that should work, but not 100% sure yet.

~~Finally, I there is the `.bin/frontity` file - how does it get generated?~~
OK, I see it's created because of the:
```
"bin": {
    "frontity": "dist/src/cli/index.js"
  },
``` 
in `package.json` in `packages/frontity`, so it's not really relevant for the test case.

I guess the bottom line is this. Since there has been no tests before, we can do one of 3 things:

1. continue with this, but I'm not 100% sure it's gonna work
2. merge without new tests 💩 
3. a third way?

Thoughts? 🙂

luisherranz:
I think you have the best position to know if it makes sense to add tests for this now, or wait for the e2e system.

Have you tested this manually?

michalczaplinski:
@DAreRodz You can review it then :) 

Like I've mentioned earlier - I've tested this manually. 
Also, the codecov for project is going to fail, because I've added new lines that are not covered.

DAreRodz:
I see that now this is the only place where `ts-node` is used in the `frontity` package:

https://github.com/frontity/frontity/blob/1d61233006dfc2a93a696eb3c7d9c8b5d2c6352b/packages/frontity/package.json#L25-L26

I guess we should move `ts-node` from `dependencies` to `devDependencies` if we want to preserve that script, right? I can change it before approving the PR.

cc: @luisherranz

luisherranz:
Yeah, sure.

DAreRodz (approved):
Approved! :ok_hand:

PR #503, opened 2020-06-23
Fix when prop child switch

### LINKED ISSUE #362: Filter the `when` prop from the Switch component children
**_Bug report_**

**Expected behavior**

The Switch component uses a `when` prop to determine which component it should render:

```js
<Switch>
  <Loading when={data.isFetching} />
  <Home when={data.isHome} />
  <Archive when={data.isArchive} />
  <Post when={data.isPostType} />
  <ErrorPage /> {/* rendered by default */}
</Switch>
```

More info on: https://docs.frontity.org/api-reference-1/frontity-components#switch

The `when` prop is only needed for `Switch`, but it shouldn't be passed down to the rendered component.

**Observed behavior**

The `when` prop is sent to the rendered component.

**Possible solution**

`Switch` can remove the `when` prop from the props of the child that is going to render. I guess it would need to get access to the props object and do something like:

```js
const { when, ...otherProps } = component.props;
// ...
return <component {...otherProps} />;
```

The code is here:
https://github.com/frontity/frontity/blob/dev/packages/components/switch.tsx#L28-L32

A test for this can be a modification of this test, where the prop `when` is missing: https://github.com/frontity/frontity/blob/add-test-for-switch-props/packages/components/__tests__/switch.tests.tsx#L55-L66

goiblas:
**What**:

In the `Switch` component rendered child with `when` prop.

**Why**:

The `when` prop is only needed for `Switch`, so that it shouldn't be passed down to the rendered component.

**How**:

Render a new component without `when` prop.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests
- [x] Changeset

**Unrelated Tasks**

- TSDocs
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages
- Community discussions

**Additional Comments**

I tried to use `cloneElement` but the behavior was not what I expected, I do not know why.
The issue fixed is [https://github.com/frontity/frontity/issues/362](https://github.com/frontity/frontity/issues/362)

### DIFF
--- a/.changeset/polite-ladybugs-relax.md
+++ b/.changeset/polite-ladybugs-relax.md
@@ -1,0 +1,5 @@
+---
+"@frontity/components": patch
+---
+
+Filter the `when` prop from the Switch component children.
--- a/packages/components/__tests__/switch.tests.tsx
+++ b/packages/components/__tests__/switch.tests.tsx
@@ -61,7 +61,6 @@ describe("Switch", () => {
 
     expect(SwitchComponent.root.findByType(TestArchive).props).toEqual({
       someProp: "prop value",
-      when: true,
     });
   });
 
--- a/packages/components/switch.tsx
+++ b/packages/components/switch.tsx
@@ -1,6 +1,11 @@
 import { connect, warn } from "frontity";
 import { Package } from "frontity/types";
-import React, { isValidElement, ReactNodeArray, ReactElement } from "react";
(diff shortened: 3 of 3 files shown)


luisherranz:
It's perfect @goiblas 🙂

I've added a changeset on your behalf so the fix will appear in our changelog in the next release in the next release.

Issue #331, opened 2020-03-04
If a handler pattern isn't matched we should return a Server Error


SantosGuillamot:
**_Bug report_**

- [x] I've searched for existing issues
- [x] I've checked documentation: https://docs.frontity.org
- [x] I've checked community forum: https://community.frontity.org

**Describe the expected behavior**
If I navigate to a url where a handler pattern is not matched, I expect to receive a Server Error explaining the case, something similar as [this one](https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/libraries/handlers/postType.ts#L39-L42), explaining that no handler was matched.

**Describe the observed behavior**
Right now, we receive an "Internal Server Error" without detailed explanation because the `getMatch` function inside `wp-source` returns null at [this line](https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/actions.ts#L57).

luisherranz:
It should be a 404 error because the URL was not found.

luisherranz:
test

PR #506, opened 2020-06-26
Fix: Every not known URL should return a 404 

### LINKED ISSUE #366: Unexpected URL should return 404 page [5pt]
**_Bug report_**

- There is a related issue (https://github.com/frontity/frontity/issues/290), closed in favor of this one.

**Expected behavior**

Some URL should return 404 instead of 500.

**Observed behavior**

My Google Search Console report that the server is return 500.

bug

For example: 
https://stream-hub.com/undefined/nap-he-dieu-hanh
https://stream-hub.com/kiem-tra-toc-do-mang/kiem-tra-toc-do-mang

When I run at localhost, the message is somehow unclear to me.

localhost

The actual error in development is this one:
```
  TypeError: Cannot read property 'split' of undefined
      at __webpack_exports__.default.str (webpack-internal:///./node_modules/@frontity/wp-source/src/libraries/handlers/utils/capitalize.ts:2:107)
      at Object.eval (webpack-internal:///./node_modules/@frontity/wp-source/src/libraries/handlers/postType.ts:10:205)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

The error is thrown in the following line because `type` is undefined:
https://github.com/frontity/frontity/blob/107d3543ce5463186809b7e6f50ca31ffbdc107d/packages/wp-source/src/libraries/handlers/postType.ts#L46 

Navigate at link http://localhost:3000/undefined/nap-he-dieu-hanh

https://gitlab.com/concalahan/stream-hub-frontity

**Info about your system**

**Possible solution**

It's a problem in the `@frontity/wp-source` package, specifically in the `postType` handler, but it's also happening with every handler that has a similar pattern (like for categories).

We can either:

* Add here a condition that would check if the fetched link and the link of the received post are the same:
  https://github.com/frontity/frontity/blob/107d3543ce5463186809b7e6f50ca31ffbdc107d/packages/wp-source/src/libraries/handlers/postType.ts#L38

* Change the pattern for that handler to make it more restrictive (right now, the pattern would match almost every link) and let `actions.router.fetch` to throw a 404 error when #331 has been implemented.

---

Thank you very much :D

michalczaplinski:
**What**:

Fix #366 . Previously, in cases when the link was matched but did not return valid data, the server would crash.

**How**:

Add a check to verify if the fetched link and the link of the received post / taxonomy are the same.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] Community discussions and issues: 
  - https://community.frontity.org/t/my-personal-blog-site/1962
  - https://community.frontity.org/t/custom-post-types-returning-500-error/1983
  - https://community.frontity.org/t/setting-homepage-in-frontity-wp-source-error/1955
  - https://community.frontity.org/t/multi-language-website-with-frontity/869/6
  - https://community.frontity.org/t/error-in-capitalize/1238
  - https://github.com/frontity/frontity/issues/331

**Unrelated Tasks**

- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages

**Additional Comments**


michalczaplinski:
I think it's sufficient to update only the postType and category handlers because they are the only ones that use a wildcard in the patterns which can cause the erroneous match:
https://github.com/frontity/frontity/blob/107d3543ce5463186809b7e6f50ca31ffbdc107d/packages/wp-source/src/libraries/patterns/wp-org.ts#L22 and https://github.com/frontity/frontity/blob/107d3543ce5463186809b7e6f50ca31ffbdc107d/packages/wp-source/src/libraries/patterns/wp-org.ts#L64

@DAreRodz 

1.  I keep getting the e2e test errors from google-analytics for some reason. Do I need to merge `master` into my local branch after the release...? 🤔
2. The codecov is also showing the dropped coverage in the google-analytics package - I also think this is related to the release. Could you have a look?

.eslintrc
```diff
@@ -8,11 +8,15 @@
     "plugin:jest/recommended",
     "plugin:cypress/recommended"
   ],
-  "plugins": ["jsdoc"],
+  "plugins": ["jsdoc", "@typescript-eslint"],
```
michalczaplinski: Looks like this should have been added before because it was causing the eslint action to fail with `Definition for rule '@typescript-eslint/no-unused-vars' was not found`.

luisherranz: That's because the typescript eslint plugins/rules are on an override. I've moved the rule to the override: https://github.com/frontity/frontity/pull/506/commits/8bcd609180a09c22959a9c04030fd21f65d4e645

DAreRodz (changes requested):
Great job, Michał! :raised_hands:

Check my comments, I think we need to change a small thing.

 Nice TSDocs, by the way. :blush:

.eslintrc
michalczaplinski: oh gotcha 👍

packages/wp-source/src/libraries/handlers/postType.ts
```diff
@@ -33,11 +74,24 @@ const postTypeHandler = ({
...
       if (populated.length > 0) {
-        isHandled = true;
-        break;
+        // We have to check if the link property in the data that we
+        // populated is the same as the current route.
+        if (populated[0].link === route) {
+          isHandled = true;
+          break;
+        } else {
+          isMismatched = true;
```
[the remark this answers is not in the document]
michalczaplinski: You're right. But now I've realized that probably we should do:

```js
if (populated[0].link === route) {
  isHandled = true;
  isMismatched = false; // <--- added this
  break;
} else {
  isMismatched = true;
}
```

Because it might happen that 
- we match a route,
- `api.get()` the entity e.g. `post`
- that entity was matched "by accident" `(populated[0].link !== route)`
- we `api.get()` another entity, e.g. `media`
- this entity is the "correct" one (the link matches)
- so we should not throw an error.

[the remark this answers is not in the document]
DAreRodz: Right, right. Yeah, let's implement it that way.

DAreRodz (approved):
Everything's great. :smile:

PR #500, opened 2020-06-17
Add eslint-plugin-tsdoc and basic configuration

luisherranz:
**What**:

Add the `eslint-plugin-tsdoc` to our eslint configuration.

**Why**:

To have some linting regarding some questions we had, like if we should use `@return` or `@retuns`, `@default` or `@defaultValue` and so on.

**How**:

By installing the package in the root and adding it as a plugin in our `eslintrc` file.

**Tasks**:

- [x] eslint-plugin-tsdoc
- [x] eslint-plugin-jsdoc
- [x] GitHub action

**Unrelated Tasks**

- TSDocs
- TypeScript
- Unit tests
- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages
- Link to PR in [Documentation](https://github.com/frontity/gitbook-docs/wiki/Code-Releases)
- Community discussions
- Changeset (with link to its [Feature Discussion](https://community.frontity.org/c/33) if it exists)

PR #507, opened 2020-06-29
Feature/link component

nicholasio:
**What**:

Introduces a `<Link>` component in `@frontity/components`.

See https://community.frontity.org/t/link-component/885

**Why**:

> `<Link>` is a component needed in all the themes. Even tho it’s not complex to create one tailored to the developer needs, we can provide a default one.

https://community.frontity.org/t/link-component/885

**How**:

A new `<Link>` component has been created in `@frontity/components`.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] Update mars-theme
- [x] Open issue in Documentation: https://github.com/frontity/docs/issues/149
- [x] Community discussions: https://community.frontity.org/t/link-component/885/19

**Unrelated Tasks**

- End to end tests
- TypeScript tests
- Update other packages
- Update twentytwenty-theme

**Additional Comments**


packages/components/link.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { MouseEvent } from "react";
+import connect, { useConnect } from "@frontity/connect";
```
nicholasio: I'm not sure if it's expected but when I was importing `connect` from `frontity` package I was getting typescript errors saying connect only expects 1 argument.

orballo: That was fixed in a beta version, but not sure if it ended in the release. Are you using the last version of the `frontity` package?

orballo: That problem comes from this declaration https://github.com/frontity/frontity/blob/8dcedad029fedbb0cfb0799d15a1f7abb924b1d4/packages/types/src/connect.ts#L13-L20

It was fixed in a beta version for the infinite scroll hook, but maybe it never made it to the release. I don't know how the move from beta to release is being done.

This is how it should look like, in case it helps
```ts
export interface ConnectFunction {
  <Comp extends FunctionComponent<any>>(comp: Comp): FunctionComponent<
    FilterInjectedProps<ComponentProps<Comp>>
  >;
  <Comp extends FunctionComponent<any>>(
    comp: Comp,
    options: { injectProps: boolean }
  ): FunctionComponent<FilterInjectedProps<ComponentProps<Comp>>>;
  <Comp extends ComponentClass<any>>(comp: Comp): ComponentClass<
    FilterInjectedProps<ComponentProps<Comp>>
  >;
}
```

You can implement the change to fix the issue and just create two changesets and merge both with the same PR, I think, but @luisherranz will know better :)

luisherranz: Our fault. Thanks for the report guys. I've made a quick fix here: https://github.com/frontity/frontity/pull/508. Once it's merged we can update this PR and import it again from `frontity` instead of `@frontity/connect`.

This duplication of types between `@frontity/connect` and `frontity` is something I want to solve as part of our migration to the new version of `react-easy-state` in https://github.com/frontity/frontity/pull/415.

michalczaplinski (approved):
Great work @nicholasio ! 💯 I've just added some small improvements:

- Added some extra comments
- Added back the `aria-current` prop that was present on the `Link` in mars-theme
- Remove the `className` prop (we are passing all the other props as `...anchorProps`)
- Add an index type `[key: string]: any` that represents all extra props that we can pass to this `Link` component
- `.toMatchSnapshot()` -->` .toMatchInlineSnapshot()`. This way we get rid of the snapshot file by using inline snapshots and the expected values are a tiny bit clearer because they're in the same file.

michalczaplinski:
We can merge this after:
- The #508 is merged. Then we can do:
```
// components/link.tsx
import { connect, useConnect } from "frontity";
```
- The JSDocs are added. We've just merged additional eslint rules requiring JSDocs. I will add them myself 🙂 
- We open the issue in the docs repo

In the future, we should add a default for prefetching like @luisherranz suggested: https://community.frontity.org/t/auto-prefetch-data/886

michalczaplinski:
@luisherranz I'm not sure why the ESLint is still failing as the error message is unrelated to the TSDocs it seems... Do you have any idea why that might be?

Additional comments about the TSDocs:

1. I've added the `eslint-disable-next-line` in 2 places:

https://github.com/frontity/frontity/blob/f2bd015818153523ee33c0506ab4f20997f6d1cc/packages/mars-theme/src/components/link.js#L19-22

https://github.com/frontity/frontity/blob/cb06184552f0a60b75839634e30dec82da4fc4f3/packages/components/link.tsx#L65-67

because I think it makes no sense to add a comment there. Can we disable the checks for functions that are defined within another function?

2. I had to disable the TSDoc for the Link component because the JSDoc parser does not handle object destructuring inside of function arguments 🙁 . See the video: 

https://www.loom.com/share/09e5737fc89948669cd55aa567bda08e

3. Ive added the TSDocs for the mars-theme because of the rule, but I think that maybe it doesn't make that much sense to add them to the components in our themes. Is it possible to disable the TSDoc on a per-folder basis?

EDIT: Sorry, It seems that the embeds are not working on github, you ll have to click on the links I posted 🙂

packages/components/package.json
```diff
@@ -25,6 +25,7 @@
   },
   "dependencies": {
     "frontity": "^1.9.0",
+    "@frontity/connect": "^1.1.2",
```
luisherranz: Remove `@frontity/connect` dependency.

packages/mars-theme/src/components/link.js
```diff
@@ -1,41 +1,34 @@
 import React from "react";
-import { connect } from "frontity";
+import { connect, useConnect } from "frontity";
```
luisherranz, line 2: Use `frontity` for these imports.

luisherranz (dismissed):
I will take a look at your issues @michalczaplinski later today, but in the meantime, could you please finish the remaining tasks?

Also, I've done this review to not forget to make these changes once https://github.com/frontity/frontity/pull/508 is merged.

And I've added another task to also use the new component in the `twentytwenty-theme`.

luisherranz:
@michalczaplinski, I've updated the TSDocs to follow @darerodz approach for React components as we agreed 🙂

Feel free to report the eslint parser bug: https://github.com/gajus/eslint-plugin-jsdoc

I'll try to finish #508 as soon as possible and I'll check why the Eslint action

---

@nicholasio sorry for all the noise in your PR, we just [started using TSDocs to document the code along with a GitHub action to run eslint](https://github.com/frontity/frontity/pull/500) and we are still ironing things out 😅

nicholasio:
> @michalczaplinski, I've updated the TSDocs to follow @DAreRodz approach for React components as we agreed 🙂
> 
> Feel free to report the eslint parser bug: https://github.com/gajus/eslint-plugin-jsdoc
> 
> I'll try to finish #508 as soon as possible and I'll check why the Eslint action
> 
> @nicholasio sorry for all the noise in your PR, we just [started using TSDocs to document the code along with a GitHub action to run eslint](https://github.com/frontity/frontity/pull/500) and we are still ironing things out 😅

That's totally fine!

It looks like this is ready to go now, (pending the merge of the blocking PR) but let me know if there's anything else needed here.

Thanks.

michalczaplinski:
@luisherranz I think that we can add the component to the twentytwenty theme later on.

In the twentywenty theme, the `Link` component is doing prefetching and we still havent figured out what is the optimal approach to it: https://community.frontity.org/t/auto-prefetch-data/886. I don't want to add the Link to twentytwenty without the prefetching because we would be removing features from the theme this way.

Besides, I think it's actually not bad if the twentytwenty has a move advanced example of how to build your own link and we also have a standard component that we use in the mars theme.

Could you remove the task if you agree, otherwise let's discus before implementing?

----

I've also reported the eslint parsing bug and made a fix (cause it wasn't too hard 😅 ): https://github.com/gajus/eslint-plugin-jsdoc/pull/599

luisherranz:
> I think it's actually not bad if the twentytwenty has a move advanced example of how to build your own link

Ok, let's keep the advanced `Link` for now on `twentytwenty-theme` 👍

> I've also reported the eslint parsing bug and made a fix

Awesome 🙂

luisherranz:
Merged. @nicholasio Many thanks for the contribution! 🎉 👏 

This will be included in our next release.

----

_The Eslint action was failing because this PR is from a fork and the GH token only has read-only permissions. I'll try to find a workaround for that in a different PR._

nicholasio:
> Merged. @nicholasio Many thanks for the contribution! 🎉 👏
> 
> This will be included in our next release.
> 
> _The Eslint action was failing because this PR is from a fork and the GH token only has read-only permissions. I'll try to find a workaround for that in a different PR._

Awesome. Thanks for the fast turn around here.

Poliuk:
Thanks for the contribution @nicholasio 🎉

PR #512, opened 2020-07-06
Remark dependency of the wordpress plugin

juanmaguitar:
Remark dependency of the wordpress plugin in README


luisherranz:
@juanmaguitar could you please follow the PR template? Even for non-code PRs it will help us to have more context (with the "What", "Why", "How" sections) and make sure we don't forget anything (thanks to the "Tasks"/"Unrelated Tasks" sections). For example, this PR is missing a changeset.

Thanks!

michalczaplinski (approved):
@juanmaguitar I've added the changeset and will merge this now.

+1 to let's keep the standard PR template in the future like Luis mentioned even for trivial PRs so that we can see the context and make sure we didn't forget anything 🙂

PR #513, opened 2020-07-06
Bugfix: Dates out of range should return 404 and not 500

michalczaplinski:
**What**:

A small fix that catches an edge case. When the year / month / day are out of range in request, e.g.: `fetch("2020/99/99")` it causes a server to crash. 

**Why**:

So that we don't crash the server and return 500 but return 404 like WordPress does.

**How**:
Do a range check on month and day of the year.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- End to end tests
- TypeScript tests
- Update starter themes
- Update other packages
- Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases)
- Update community discussions

**Additional Comments**


packages/wp-source/src/libraries/handlers/date.ts
```diff
@@ -22,6 +39,13 @@ export const dateHandler: Handler = async ({
   const month = params.month && parseInt(params.month);
   const day = params.day && parseInt(params.day);
 
+  if (typeof month === "number" && (month < 1 || month > 12)) {
+    throw new ServerError(`${month} is a wrong month number`, 404);
+  }
+  if (typeof day === "number" && (day < 1 || day > 31)) {
```
DAreRodz: Maybe there is a way to use the real number of days here for the given month an year...

I've tested http://localhost:3000/2016/02/31/ and it's showing all posts published in March, 2016. :sweat_smile:

DAreRodz (changes requested):
Nice catch, Michal! Although I think we can improve it a bit. :wink:

packages/wp-source/src/libraries/handlers/date.ts
michalczaplinski: fair enough :)

We can check for leap year with: https://stackoverflow.com/a/16353241/2638310

michalczaplinski:
I will look into the e2e test failure later on - it looks like we are shouldn't be matching the date handler when running those tests: 

```js
it("should return a 500 status code", async () => {
    const res500 = await cy.request({
      url: "http://localhost:3001/500?name=wp-source-errors",
      failOnStatusCode: false
    });
    expect(res500.status).toBe(500);
  });
  ```
  
  and 
  
  ```js
   it("should have data populated correctly with 500 errors", () => {
    cy.visit({
      url: "http://localhost:3001/500?name=wp-source-errors",
      failOnStatusCode: false
    });
    cy.get("[data-test-id='isError']").should("contain.text", "true");
    cy.get("[data-test-id='is404']").should("contain.text", "false");
    cy.get("[data-test-id='status']").should("contain.text", "500");
  });

**Comment by michalczaplinski** [comment]
> I will look into the e2e test failure later on - it looks like we are shouldn't be matching the date handler when running those tests:
> 
> ```js
> it("should return a 500 status code", async () => {
>     const res500 = await cy.request({
>       url: "http://localhost:3001/500?name=wp-source-errors",
>       failOnStatusCode: false
>     });
>     expect(res500.status).toBe(500);
>   });
> ```
> 
> and
> 
> ```js
>  it("should have data populated correctly with 500 errors", () => {
>   cy.visit({
>     url: "http://localhost:3001/500?name=wp-source-errors",
>     failOnStatusCode: false
>   });
>   cy.get("[data-test-id='isError']").should("contain.text", "true");
>   cy.get("[data-test-id='is404']").should("contain.text", "false");
>   cy.get("[data-test-id='status']").should("contain.text", "500");
> });
> ```

I've fixed it by using the url like `http://localhost:3001/?code=500&name=wp-source-errors` so that the status code is not part of the route but rather a separate variable in the query string.

File: packages/wp-source/src/libraries/handlers/utils/validateDate.ts [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+ * Check that the given combination of year, month & day is valid.
+ *
+ * @param year - Year as integer.
+ * @param month - Month of the year as integer (should start with 1).
+ * @param day - Day of the month as integer (should start with 1).
+ *
+ */
+function validateDate(year: number, month: number, day: number) {
+  // This is just an arbitrary sanity check.
+  // We ll generously assume that frontity will continue to exist for a hundred years 😅
```
DAreRodz, line 25: That made me laugh. :joy:

DAreRodz:
Michal, let me know when this is ready to review again.

michalczaplinski:
> Michal, let me know when this is ready to review again.

Oh, I thought that I had re-requested the review already. It's ready!

DAreRodz (approved):
It's great! :clap:

PR #514, opened 2020-07-07
Codecov: use in informational mode

michalczaplinski:
Use codecov in informational mode only. Now we only get the comment but never prevent the PR from merging.

It seems that the "check" in the PR will always be present and green and they can't just be disabled.

Also, I guess that the codecov gets confused when we merge the branches for the release and it's still showing the missing coverage for google analytics. I think we can ignore that or at least it's not worth spending more time investigating.


luisherranz:
Any idea why Codecov is still reporting files from the analytics packages? Maybe it got corrupted at some point? It doesn't make much sense...

luisherranz:
Yeah, there's definitely something going on. All the new PRs are getting a wrong comment: https://github.com/frontity/frontity/pull/515#issuecomment-655370679

@michalczaplinksi, could you please investigate?

luisherranz:
I updated the branch and now it's showing the file from this other PR: https://github.com/frontity/frontity/pull/515 which is the only one merged in `dev` since the last release.

luisherranz:
Now it's fine! 🎉

luisherranz:
I guess this is ready for review again and revisit if it happens again.

DAreRodz (approved):
:ok_hand:

PR #518, opened 2020-07-11
Feature/auto prefetching

nicholasio:
**What**:

Introduces auto prefetching for links.

**Why**:

Auto prefetch allows the link to open instantaneously by prefetching all the data required to render the page ahead of time.

**How**:

This PR implemented the user stories described here https://community.frontity.org/t/auto-prefetch-data/886

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] Update starter themes
- [x] Update other packages
- [x] Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases): https://github.com/frontity/docs/issues/149

**Unrelated Tasks**
- End to end tests
- TypeScript tests

**Additional Comments**


nicholasio:
The "wp-source-errors.spec.js " are also failing locally for me, looks like it has to do with SSL certs issue so I'm assuming it's unrelated to this PR.

Other than that this should be ready for review!

nicholasio:
Another thing I was thinking of: there's an opportunity to combine multiple requests into one: for example, when fetching multiple categories links we could probably make one HTTP request passing an array of categories and then process the results to map them back to state.

That would definitely add more complexity so probably a separate feature/PR. Assuming the solution I'm proposing here is the way we want to proceed (versus relying on browser prefetch)

luisherranz:
> The "wp-source-errors.spec.js " are also failing locally for me, looks like it has to do with SSL certs issue so I'm assuming it's unrelated to this PR.

Yes, it was unrelated. The certificate of [httpstat.us](https://httpstat.us) expired. We should not rely on third party services but we are still working on a way to [start a WordPress container for our e2e tests](https://community.frontity.org/t/e2e-testing-that-requires-wordpress-instances/1756). It should be fixed now.

> Another thing I was thinking of: there's an opportunity to combine multiple requests into one

Let's talk about this on [the community thread](https://community.frontity.org/t/auto-prefetch-data/886/3), to keep the conversation in one place. I'll answer you there 🙂

packages/components/__tests__/link.tests.tsx
```diff
@@ -236,3 +252,283 @@ describe("Link", () => {
...
+          <Link link={linkUrl2} className="my-link-2">
+            This is a link
+          </Link>
+        </Provider>,
+        container
+      );
+    });
+
+    const anchor = document.querySelector("a.my-link");
+    mockIsIntersecting(anchor, true);
```
luisherranz, line 506: Nice trick!

luisherranz (approved):
I'm going to approve the PR because the code and the functionality is great. If you still want to give the `requestIdleCallback` approach a try as we discussed in [the community thread](https://community.frontity.org/t/auto-prefetch-data/886/7) please go ahead. If you don't or can't do it now let us know and we'll merge it as it is.

nicholasio:
> I'm going to approve the PR because the code and the functionality is great. If you still want to give the `requestIdleCallback` approach a try as we discussed in [the community thread](https://community.frontity.org/t/auto-prefetch-data/886/7) please go ahead. If you don't or can't do it now let us know and we'll merge it as it is.

I took a quick look at implementing `requestIdleCallback` and I think it would be better to create a new FD. There are some other things for us to consider like adding a shim for it. TypeScript also does not recognize `requestIdleCallback` as it is officially experimental so we would have to figure that out as well. It doesn't feel right to add a shim and typescript definitions for this in the Link component so I'd rather do that a part of a new FD.

michalczaplinski:
Awesome, thanks for the contribution, @nicholasio ! 🙂

PR #520, opened 2020-07-20
Feature/content link client routing

nicholasio:
**What**:

Introduces a processor to replace content links with the Link component.

**Why**:

It's needed to allow frontity to handle internal links automatically (without causing a new page load)

**How**:

A processor was created to replace all anchor tags that have an href with our Link component.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] Open an issue for this feature in the docs repo: https://github.com/frontity/docs/issues/248

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] End to end tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


e2e/e2e.js
```diff
@@ -42,7 +42,9 @@ let isDockerRunning = false;
 
 // Validate CLI args.
 validateArgs(target, { possibleValues: ["es5", "module", "both"] });
-validateArgs(browser, { possibleValues: ["firefox", "chrome", "edge"] });
+validateArgs(browser, {
+  possibleValues: ["firefox", "chrome", "edge", "electron"],
```
nicholasio, line 46: I'm not sure if there was a reason to leave `electron` out but I find it useful for being able to easily run tests in WSL 2 without having to install chrome. BTW the e2e stuff worked fine in WSL 2 with Docker for Desktop (Windows).

packages/html2react/src/index.ts
```diff
@@ -1,12 +1,13 @@
...
 import Component from "./libraries/component";
 import script from "../processors/script";
+import link from "../processors/link";
 
 const html2react: Html2React = {
   name: "@frontity/html2react",
   libraries: {
     html2react: {
-      processors: [script],
+      processors: [script, link],
```
luisherranz: I guess we have to make a decision here: should the `link` processor be added by default or should be added by the theme creator like the `image` and `iframe` processors?

My opinion now is that we should leave it out for the moment, as the `<Link>` component itself is optional in the themes. But we should improve the way people share processors. The conversation is happening here: https://community.frontity.org/t/how-to-share-html2react-processors/3069. It would be awesome if you could give us your feedback @nicholasio 🙂.

The best idea we had so far is to make them similar to the way `fills` work now. It is a bit verbose, but it is consistent and very flexible.

`script` is not optional because if not the `<script>` tags of the `post.content` don't work anymore.

packages/html2react/processors/__tests__/link.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+store.libraries.html2react.processors.push(link);
+
+jest.mock("@frontity/components/link", () => "mocked-link");
+
+describe("Link Processor", () => {
+  it("should replace all anchor tags with links", () => {
+    const { container } = render(
+      <Html2React
+        html="<a href='http://my-link.com/post-name'>My Link</a><a href='http://my-link.com/post-name'>My Link 2</a>"
+        {...store}
```
luisherranz: @DAreRodz maybe we can simplify the processor tests now that you have added a `processors` prop to `Html2React`.

What do you think? Would something like this work?

```js
<Html2React
  html="<a href='http://my...ame'>My Link 2</a>"
  processors={[link]}
/>
```

packages/components/link/utils.ts
```diff
@@ -111,3 +111,17 @@ export const onHover = (el: HTMLAnchorElement, cb: () => void) => {
...
+
+/**
+ * Removes the WP URL.
+ *
+ * @param link - The link URL.
+ * @param wpURL - The WP URL.
+ *
+ * @returns The URL without the WP URL.
+ */
+export const removeWPUrl = (link: string, wpURL: string) => {
```
luisherranz: Maybe it would be great to add some unit tests for this util? It would be easier to fix edge cases and it would make sure it's not messed up in future refactorings. What do you think @nicholasio?

luisherranz (changes requested):
Hey, awesome work @nicholasio. Many thanks for this PR 🙂

The only thing I would change is `replaceWPUrls` to `replaceSourceUrls` because, although now only `wp-source` exists, theoretically the link can be used by any source, even sources other WordPress. Also, it relates more to the `state.source.url` setting, which is what uses underneath to figure out what to replace.

packages/components/link/utils.ts
nicholasio: Yep, will add tests for this.

packages/html2react/processors/link.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * A link processor that replaces all anchor tags with the {@link Link} component.
+ *
+ * The allows internal links within content to work with frontity's internal router.
+ */
+const link: Processor<LinkElement> = {
+  priority: 10,
+  name: "link",
+  test: ({ node }) => node.component === "a",
+  processor: ({ node }) => {
+    if (node.props?.href && !node.props?.href?.startsWith("#")) {
```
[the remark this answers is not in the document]
nicholasio: I'll move the conditional to the test function.

Regarding the hash links - not sure of the benefits of handling it through router/link component but I can certainly include hash links.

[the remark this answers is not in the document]
luisherranz: I'm fine merging this as it is, I just wanted to draw the attention of orballo to this processor to make sure it works fine with his implementation 🙂

packages/components/link/utils.ts
luisherranz: Great. Thanks :)

packages/html2react/processors/__tests__/link.test.tsx
DAreRodz: Sure, that should work fine. When the `processors` prop is present, the content of `libraries.html2react.processors` is just ignored.

packages/html2react/src/index.ts
nicholasio: Makes sense to leave it out for now - but we need to make this very clear in the documentation as this is definitely going to be a must-need feature for most themes.

nicholasio: I was going to say that it makes sense but if the script is not optional it means that all themes are already shipping with frontity/components package, so I don't see a reason why not to make the link processor be enabled by default.

packages/html2react/processors/__tests__/link.test.tsx
nicholasio: FYI: TS complains about types not matching if we just pass processors since the other props aren't optional.

packages/html2react/src/index.ts
nicholasio: Actually, I see your point, you're concerned more about the way processors are shared and not themes needing to declare `frontity/components` as a dependency. I'll make it optional and include in the default themes.

.changeset/itchy-moles-enjoy.md
```diff
@@ -0,0 +1,8 @@
+---
+"@frontity/components": minor
+"@frontity/html2react": minor
+---
+
+Introduces a `replaceWPUrls` prop to the `Link` component and `link` processor that replaces all anchor tags with the `Link` component to make internal links work.
```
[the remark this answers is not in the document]
nicholasio: Done.

packages/html2react/src/index.ts
nicholasio: Done.

packages/components/link/utils.ts
nicholasio: Done.

packages/html2react/processors/__tests__/link.test.tsx
DAreRodz: Oh, I see. That happens because the _disconnected_ `Html2React` is being imported

```ts
import { Html2React } from "../../src/libraries/component";
```
instead of the _connected_ version, which is the default export:

```ts
import Html2React from "../../src/libraries/component";
```

That one should not require the Frontity store props to be passed. :slightly_smiling_face:

nicholasio: That worked, updated 👍

luisherranz (approved):
Looks perfect to me! Thanks for the PR and thanks for refactoring also the other processor tests!

I have added tests for the case when `state.source.url` contains a folder and I have refactored the `removeWPUrl` function to make it work. Could you please review [the new implementation](https://github.com/frontity/frontity/pull/520/commits/8e9eac4ab65dd64d4e78975883d09390f4032d53#diff-a2a24789b184eb4ccec4bf21e40d3a76006682d94c7f95d56e76c8a3bd7445a3R124-R128) and let me know if that is how you think it should work? Once you give me your approval I will merge the PR 🙂

packages/html2react/src/index.ts
luisherranz: Thanks @nicholasio :)

luisherranz:
BTW, it seems like an E2E test started failing but it's not related: https://github.com/frontity/frontity/issues/619

nicholasio:
> Looks perfect to me! Thanks for the PR and thanks for refactoring also the other processor tests!
> 
> I have added tests for the case when `state.source.url` contains a folder and I have refactored the `removeWPUrl` function to make it work. Could you please review [the new implementation](https://github.com/frontity/frontity/pull/520/commits/8e9eac4ab65dd64d4e78975883d09390f4032d53#diff-a2a24789b184eb4ccec4bf21e40d3a76006682d94c7f95d56e76c8a3bd7445a3R124-R128) and let me know if that is how you think it should work? Once you give me your approval I will merge the PR 🙂

Looks good to me! 👍

luisherranz:
Awesome. Merged now. Thanks again @nicholasio! 🎉

PR #521, opened 2020-07-21
Add WP comments handler

### LINKED ISSUE #440: Add handler to fetch the state with the comments [21pt]

DAreRodz:
**What**:

Implement a `wp-source` handler to fetch all comments from the give post.

This handler has uses `"comments/:postId(\\d+)"` as pattern so, for example, to fetch all comments published in the post with ID `60` you would do:

```ts
await actions.source.fetch("comments/60");
```

All retrieved comments are stored in `state.source.comment` and a data object is created containing a tree structure of comments and their replies. 

**Why**:

To release the first version of `@frontity/wp-comments` package.

Feature Discussion: https://community.frontity.org/t/wordpress-comments-package/1267

**How**:

I followed a similar approach @SantosGuillamot  used in https://github.com/frontity/frontity/pull/225

As he did, I had to modify `wp-source` to add a schema for comments even though this is a different package. Currently there's no other way to extend schemas. :slightly_frowning_face: 

Hopefully, we'll get rid of schemas in the v2 of `wp-source`.

Ah, and I'm not using Cypress tests in this PR. Instead, I did the same we did before to test `wp-source` default handlers.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases): https://github.com/frontity/docs/issues/174
- [x] Community discussions (https://community.frontity.org/t/wordpress-comments-package/1267/50)

**Unrelated Tasks**

- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages
- [ ] Add a changeset (with link to its [Feature Discussion](https://community.frontity.org/c/33) if it exists) (already added in #495)

**Additional Comments**

This PR has changes from #495, so it has to be merged first before reviewing this.


luisherranz:
@DAreRodz mention test :)

packages/wp-comments/src/libraries.ts
```diff
@@ -0,0 +1,119 @@
...
+        .map((_, i) => fetchComments({ postId, page: i }))
+    );
+
+    // Add other pages to rawItems.
+    rawItems = rawItems.concat(...otherPages.map(({ populated }) => populated));
+  }
+
+  const commentsMap: Record<string, CommentItem> = {};
+  const items = rawItems
+    .map(({ type, id }) => (commentsMap[id] = { type, id }))
```
michalczaplinski, line 91: I would use `.forEach` here instead of `.map` because you are creating a side effect (creating a value in `commentsMap` for each item) and not actually mapping the values to any other values. 

I was surprised that this works actually but then I realized that it's because the assignment of `commentsMap[id] = { type, id }` returns the value assigned (`{ type, id }`), so you're just mapping `{ type, id } => { type, id }` in this case.

packages/wp-comments/src/libraries.ts
```diff
@@ -0,0 +1,119 @@
...
+        const parentItem = commentsMap[parent];
+        parentItem.children = parentItem.children || [];
+        parentItem.children.push(item);
+      } else {
+        // Add it to the root list.
+        root.push(item);
+      }
+
+      return root;
+    }, [] as CommentItem[]);
```
michalczaplinski, line 107: Do you think this would be clearer as a for loop? I find the`.reduce()` usually a bit difficult to comprehend like this one here 😅  Sth like:

```ts
let items: CommentItem[] = []
for (let item of rawItems) {
  // Get the parent ID if this comment is a reply.
  const { parent } = state.source[item.type][item.id];

  if (parent !== 0) {
    // Add it to its parent's children list if it is a reply.
    const parentItem = commentsMap[parent];
    parentItem.children = parentItem.children || [];
    parentItem.children.push(item);

  } else {
    // Add the item to the list.
    items.push(item);
  }
}
```

michalczaplinski (commented):
Really nice work @DAreRodz 👍 

I just noticed that according to the proposal, the comments link should be preceded with a `@`. So, the fetch should be like: `source.fetch("@comments/123");`.

Is there a reason why we dropped the `@` ?

packages/wp-comments/src/libraries.ts
DAreRodz: It is clearer, indeed. I think I overuse `map` and `reduce` sometimes. :sweat_smile:

DAreRodz:
> Is there a reason why we dropped the `@`?

The only reason to use the `@` was to clarify that you are not fetching a URL but something different, and I think we can just point that out in the docs so I removed it.

Do you think it would be useful to maintain it?

michalczaplinski:
> > Is there a reason why we dropped the `@`?
> 
> The only reason to use the `@` was to clarify that you are not fetching a URL but something different, and I think we can just point that out in the docs so I removed it.
> 
> Do you think it would be useful to maintain it?

I think it would be very useful to keep the special character 🙂  Especially as going forward we will have more non-url data like menus and attachments and they could also be prefaced with `@` so we can distinguish them from "URL" data. 

I assumed that this was already decided based on: https://community.frontity.org/t/handlers-for-non-url-data/714/17 and the next comment but I think that this discussion was a bit hidden and was not referenced in the wp-comments thread for some reason.

michalczaplinski:
By the way, other than what I mentioned above I think it's good to merge so no need to wait till I come back from holidays 🙂

DAreRodz:
> I assumed that this was already decided based on: https://community.frontity.org/t/handlers-for-non-url-data/714/17 and the next comment but I think that this discussion was a bit hidden and was not referenced in the wp-comments thread for some reason.

Oh, I forgot about that conversation! I'll add the `@` again then, thanks for pointing it out. :blush:

luisherranz:
Great work guys 🙂

PR #525, opened 2020-07-27
Fix package names that start with a number

### LINKED ISSUE #524: Frontity fails if custom package name starts with numbers.
**_Bug report_**
Frontity does not work with package names such as `@10up/frontity-elasticpress` even though it is a valid npm package.

image

**Expected behavior**

Any valid npm package name would work.

**Observed behavior**

Valid npm packages that start with a number does not work with frontity.

**Info about your system**

**Possible solution**

I think the solution can be as simple as prefixing all packages with a valid variable name prefix when referencing them:

`import frontity_package_{normalized_package_name} from "@npm/package-name`

luisherranz:
**What**:

Packages that start with a number, like `@123/package` didn't work.

**Why**:

Because we generate variables out of those names in our entry point files and we were not contemplating that case.

**How**:

Fixing the function that turns package names into variable names.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] Unit tests
- [x] Add a changeset

**Unrelated Tasks**

- End to end tests
- TypeScript
- TypeScript tests
- Update starter themes
- Update other packages
- Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases)
- Update community discussions

### DIFF
--- a/.changeset/tasty-zebras-breathe.md
+++ b/.changeset/tasty-zebras-breathe.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Fix for package names that start with a number, like `@123/package`.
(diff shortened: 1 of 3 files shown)


DAreRodz (approved):
That was fast! Great job, @luisherranz. :raised_hands:

PR #531, opened 2020-08-04
Fix react test rendered types

luisherranz:
**What**:

I've fixed the types of some tests that use the `app.toJSON()` helper of `react-test-renderer`.

**Why**:

Because the latest version of `@types/react-test-renderer` returns `ReactTestRendererJSON | ReactTestRendererJSON[]` for `app.toJSON()` and TypeScript is confused and it doesn't know which type to use.

**How**:

I've added two different type guard functions that return the proper type for each situation.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


luisherranz:
I wonder if there's an easier way to do this... I guess I could have done a manual casting for each line where we use `app.toJSON()`, but seemed like worse than using type guard functions.

Right now I cannot think of a better/more elegant way to solve this. Ideas are welcomed 😅

packages/connect/src/index.d.ts
```diff
@@ -45,7 +45,7 @@ type ResolveActions<Act extends Actions> = {
     ? () => ReturnType<Act[P]>
     : Act[P] extends Actions
     ? ResolveActions<Act[P]>
-    : never;
+    : any;
```
DAreRodz, line 48: What does this line fix? There is nothing related to `connect` in the PR description. :thinking:

packages/frontity/src/utils/slot-and-fill/__tests__/__snapshots__/use-fills.test.tsx.snap
```diff
@@ -1,5 +1,54 @@
...
+`;
+
+exports[`Slot should render children as a fallback 1`] = `"fallback"`;
+
+exports[`Slot should work in the most basic case 1`] = `
+<div>
+  Im a Fill
+</div>
+`;
+
```
DAreRodz, line 51: I guess the snapshots were missing because tests were failing, right? Did we merge this PR using `force` or something? Just curious. :detective:

DAreRodz (approved):
I think using type guards is a great solution. :+1:

packages/frontity/src/utils/slot-and-fill/__tests__/__snapshots__/use-fills.test.tsx.snap
luisherranz: I have no idea. I also noticed that and I investigated a bit but didn't find any reason why those snapshots were missing 🤷

packages/connect/src/index.d.ts
luisherranz: It fixes another TypeScript problem, related to these lines https://github.com/frontity/frontity/blob/dev/packages/core/src/server/index.tsx#L95. It was complaining that `never` was not callable.

I'll try to improve the TypeScript coverage of that part once I reorganize our types. Those types are a bit of a mess right now.

PR #532, opened 2020-08-04
Try to fix the Google Analytics e2e tests

### LINKED ISSUE #535: e2e tests for the Google Analytics package fail in a random way
**_Bug report_**

*NOTE: This is not exactly a bug because the code seems to be fine but it prevents some PRs to be merged as it randomly makes e2e workflows to fail, having to re-run workflows until the tests pass.*

**Expected behavior**

The e2e tests for the Google Analytics package should pass as expected if the code is fine.

**Observed behavior**

They fail sometimes in a random way.

It can be reproduced in a local environment:

- Clone the `frontity` repo.
- Go to `e2e` and run `npm run e2e:open`.
- Once the Cypress app is open, run google-analytics spec file.
- Re-run that spec until it fails (it can take a long time).

**Possible solution**

I found out that the problem is the implementation of the e2e tests. I assumed that the `analytics.js` library always uses an `Image` element to create the request to Google Analytics. The thing is that it sometimes does an `XMLHTTPRequest` instead of using an `Image` and, as the tests are not expecting that case, it fails.

We could mock an `analytics.js` function instead of `Image` constructor for these tests.

DAreRodz:
**What**:

**Why**:

**How**:

**Tasks**:

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Unrelated Tasks**

**Additional Comments**


luisherranz:
@Darerodz could you please fill up the PR description to have a bit more context on the goal of this PR? Thanks!!

luisherranz:
Superseded by https://github.com/frontity/frontity/pull/546.

PR #533, opened 2020-08-05
Adds isError to infinite scroll hooks

orballo:
**What**:

Adds `isError` boolean to `useArchiveInfiniteScroll` and `usePostTypeInfiniteScroll`.
Also adds the intersection observer options as parameters for the same hooks.

**Why**:

The `isError` option is useful to write logic that depends on that state.
The observer options are needed inside e2e tests because when running `IntersectionObserver` inside an iframe, the `root` option needs to be set manually.

**How**:

The flag `isError` will be true when either the last post has failed to fetch or the last page has failed to fetch.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests
- [x] TSDocs

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


orballo:
I'm running into the following error when trying to create a changeset:
```
🦋  error Error: Failed to find where HEAD diverged from infinite-scroll-beta. Does infinite-scroll-beta exist?
🦋  error     at getDivergedCommit (/Users/orballo/Code/frontity/node_modules/@changesets/git/dist/git.cjs.dev.js:51:11)
🦋  error     at async getChangedFilesSince (/Users/orballo/Code/frontity/node_modules/@changesets/git/dist/git.cjs.dev.js:69:22)
🦋  error     at async Object.getChangedPackagesSinceRef (/Users/orballo/Code/frontity/node_modules/@changesets/git/dist/git.cjs.dev.js:108:24)
🦋  error     at async add (/Users/orballo/Code/frontity/node_modules/@changesets/cli/dist/cli.cjs.dev.js:369:29)
🦋  error     at async /Users/orballo/Code/frontity/node_modules/@changesets/cli/dist/cli.cjs.dev.js:950:5
```
Any ideas?

luisherranz:
> I'm running into the following error when trying to create a changeset:

It didn't happen here. I'll add it myself, although if it happens again on your computer let us know.

luisherranz (changes requested):
Hey @orballo, great work!

I really like the e2e tests with the stubbing of the `fetch` (finally 🎉).

I've simplified the fixtures and moved them to the fixtures folder. I've also used the new divided types pattern we are following now (more on that here: https://community.frontity.org/t/typescript-use-two-different-types/1681).

I've also merged everything from `dev` because there were some things we fixed there and I think we were to back behind with this beta.

The only problem is that the e2e tests are failing in Electron, which is the thing that runs in GH actions. Would you mind checking them out with Electron? It is complaining that the `root` is not valid for the Intersection Observer...

You can use Electron by switching it here:

Screenshot 2020-08-11 at 23 39 43

BTW, if you see package problems, do a clean install. I did a lot of changes to the package-locks and so on.

luisherranz:
Ups, now that I've merged `dev`, we also need to add proper TSDocs to the modified files 😅 

Once you've fixed that let me know and I'll release a new beta.

luisherranz:
The e2e tests are fixed by switching to Chrome in our GH actions. Let's do that instead of trying to fix them for Electron 👍

@orballo, then only the TSDocs are left.

orballo:
TSDocs finished 😄

luisherranz (approved):
Awesome 🙂

Feel free to merge and update the feature discussion with the latest changes: https://community.frontity.org/t/infinite-scroll-hooks/2055/16

I will release a new beta and I'll announce it both here and in the FD.

orballo:
@luisherranz let me know when the new version is released :)

PR #540, opened 2020-08-19
Get rid of short flags in the CLI

luisherranz:
**What**:

This PR gets rid of all the short flags, like `-h` or `-s`.

**Why**:

They were not consistent. For example, in a command, `-n` meant `--namespace` and `--no-prompt` didn't have any short flag, but in another command `-n` meant `--no-prompt`.

Some of them are not even valid, like `-h` for the `--theme`, because `-h` is also used for help:

```bash
-h, --help  output usage information
```

I don't think it makes sense to expose these types of args just for the shake of using them. I think we should get rid of them now, and then maybe introduce them in the future after agreeing on a standardize way to create them.

**How**:

I just got rid of them in the `commander` configuration.

I also added the alias `--prod` for `--production` and `--dev` for `--development`.

**Tasks**:

- [x] Code
- [x] Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases): https://github.com/frontity/docs/issues/199

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

### DIFF
--- a/.changeset/small-islands-reply.md
+++ b/.changeset/small-islands-reply.md
@@ -1,0 +1,5 @@
+---
+"frontity": minor
+---
+
+Remove all the short flags from the CLI commands.
--- a/packages/frontity/src/cli/index.ts
+++ b/packages/frontity/src/cli/index.ts
@@ -40,7 +40,7 @@ program
 
 program
   .command("dev")
-  .option("-p, --production", "Builds the project for production.")
+  .option("--prod, --production", "Builds the project for production.")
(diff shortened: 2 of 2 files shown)


michalczaplinski (approved):
I was gonna suggest the same thing, actually. You beat me to it 😄 

Looks good to me 👍

luisherranz:
Thanks Michal!

PR #542, opened 2020-08-21
Update the wp-comments package to use the WP REST API directly

### LINKED ISSUE #439: Add action to create new comments [13pt]

michalczaplinski:
**What**:

Update the implementation of the `wp-comments` package to use the REST API directly. 

**Why**:

The previous implementation has a problem with following redirects as outlined in the Feature Discussion: https://community.frontity.org/t/wordpress-comments-package/1267/53?u=mmczaplinski

**How**:

- Updating the`wp-comments` package to use the REST API directly
- Add the data returned by the REST API when POSTing a new comment directly to `state.source.data` and `state.source.comment`. This is handled inside of the `submit()` action.
- Store the validation errors returned from WordPress in the state. The REST API returns the validation errors for each of the fields in the comment. E.g. if you send a request with a malformed `author_email`
- Add a new type of `data` to `@frontity/source`
- Removing the `query`, `route` & `link` from the non-URL entities that are fetched from the REST API.

Also additionally:
- Add a new `load-db.js` script for local testing using the e2e testing system.
- It also includes some improvements to the e2e testing system (adding retries & migrating the plugins file and the wp-comments e2e test file to typescript)

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] End to end tests
- [x] Unit tests
- [x] Update other packages
- [x] Update community discussions https://community.frontity.org/t/wordpress-comments-package/1267/82
- [x] Open an issue for this feature in the docs repo: https://github.com/frontity/docs/issues/174

**Unrelated Tasks**

- [x] TypeScript tests
- [x] Update starter themes


e2e/plugins/index.ts
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * The utility type which facilitates extracting the types for the
+ * individual cypress tasks.
+ */
+export type taskTypes = <T extends keyof typeof tasks>(
+  event: T,
+  arg?: Parameters<typeof tasks[T]>[0],
+  options?: Partial<Cypress.Loggable & Cypress.Timeoutable>
+) => Cypress.Chainable<any>;
```
michalczaplinski, line 139: This type is used over here to get autocompletion and typesafety when calling cypress tasks.
It's admittedly not the most elegant way but since this is internal code and not a public API I think it's not a big problem.

https://github.com/frontity/frontity/pull/542/files#diff-24302ad5d3de5c2128e9f18d816c21eeR1-R3

packages/source/types/data.ts
```diff
@@ -1,4 +1,6 @@
+/* eslint-disable */
```
michalczaplinski, line 1: This is temporal because @luisherranz has already created the TSDocs for this file in https://github.com/frontity/frontity/pull/564

packages/wp-comments/src/__tests__/__snapshots__/commentsHandler.test.ts.snap
```diff
@@ -75,7 +75,7 @@ Object {
 
 exports[`commentsHandler should handle replies correctly 3`] = `
 Object {
-  "areComments": true,
+  "isComments": true,
```
michalczaplinski, line 78: I think `isComments` is correct if you consider "Comments" to be a proper noun like "Rolling Stones" or "eighties" 🙂

michalczaplinski:
We are receiving same error on this PR as reported by @luisherranz previously in https://github.com/bradennapier/eslint-plus-action/issues/45 
`Failed to generate URL to download artifact` 😕

michalczaplinski:
The e2e tests are failing on the previous version of WP - I will look into this, but in the meantime you can already review it @DAreRodz :)

e2e/cypress.json
```diff
@@ -11,5 +11,9 @@
   "ignoreTestFiles": ["**/__snapshots__/*", "**/__image_snapshots__/*"],
   "projectId": "ed7hed",
   "defaultCommandTimeout": 10000,
-  "taskTimeout": 60000
+  "taskTimeout": 60000,
+  "retries": {
+    "runMode": 3,
```
michalczaplinski, line 16: More info on this: https://www.cypress.io/blog/2020/08/19/introducing-test-retries-in-cypress-5-0/

michalczaplinski:
> The e2e tests are failing on the previous version of WP - I will look into this, but in the meantime you can already review it @DAreRodz :)

The test failures were really weird and had multiple causes:

1. In WordPress 5.4 & 5.3, the default comment with ID == 1 that is present in the database (The one that starts with `This is a test. If you want to get started with moderation...` is not being returned from the REST API. You can fetch it using `GET post/1/comments/1`, but it's missing from the list of all comments in `GET post/1/comments/`! I

   I'm not quit sure why that is the case, but I've just worked around it by deleting the default comment and just adding my own to the database.

2. Another problem seems to be that the approved / hold status of the comments depends on some factors that we have not anticipated. In the e2e tests, when I submit a test comment, the same comment appeared to be `approved` in `5.4` but is `hold` in `5.5` !

   I have first tried to work around it by checking the WP version and basically check that: `wpVersion < 5.5 ? "approved" : "hold"`. However, with some more testing it turns out that sometimes the same comment also gets stuck on `hold` in 5.4. 
My only conclusion is that the `approve`/`hold` status of the comments depends on some other factors that are hard to control like time since last posted a comment from a particular IP or that the classification is non-deterministic.

DAreRodz (approved):
It looks great, @michalczaplinski! :ok_hand:

e2e/integration/wordpress-01/wp-comments.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * Strangely, the default comment with ID === 1 is not returned from the REST API in WP 5.4 and below.
+ * Because of that, this comment is deleted and by default, only the comment with ID === 2 is present in the database.
+ *
+ * Any subsequent comments added to WordPress, will have a comment ID of 3 or above.
+ *
+ */
+describe("wp-comments", () => {
+  before(() => {
+    cy.task("installPlugin", {
```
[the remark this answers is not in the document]
michalczaplinski: ah, nono, just an honest mistake 🙂 thanks!

packages/wp-comments/src/index.ts
```diff
@@ -1,6 +1,102 @@
...
+    const totalPagesIncrement = commentData.total % 100 === 0 ? 1 : 0;
+
+    Object.assign(commentData, {
+      total: commentData.total + 1,
+      totalPages: commentData.totalPages + totalPagesIncrement,
+      items,
+    });
+  } else {
+    // This means that it's a new comment so we just create all the new data.
+    // We also have to set the `isReady` and `isFetching` flags appropriately.
```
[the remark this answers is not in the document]
michalczaplinski: Actually, you're right! Previously I was setting the flags on that line, so that comment is outdated. Thanks!

michalczaplinski:
I will merge this now, despite the eslint failure as I have no eslint issues locally so it was most likely just a problem with the github action.

Note that I have [added the `/* eslint-disable */` to silence the warnings](https://github.com/frontity/frontity/blob/a40a82fa9a16c8e8177be0073014cd730ba5ef9d/packages/source/types/data.ts#L1) as Luis has already implemented the TSDocs in another PR

PR #543, opened 2020-08-25
Update Cypress and Jest versions and remove Node 8 support

luisherranz:
**What**:

I've updated Cypress and Jest to their latest versions.

I've deprecated the `URL` import from `"frontity"`, which was the only thing we had for backward compatibility with Node 8.

I've also fixed a couple of issues with Cypress e2e tests:

- Remove WP CLI containers after `run`.
- Remove all the containers (WP, MySQL, WP CLI) after Cypress has sucessfully run.
- Remove all the containers after an error.

**Why**:

We've made several changes to our test system, I think it's good to keep it up to date with the latest Jest/Cypress versions.

Nobody supports Node 8 anymore. I think it's not even on maintenance anymore. All serverless platforms have updated to either Node 10 or Node 12. https://nodejs.org/en/about/releases/

**How**:

Simply by updating the dependencies and creating a wrapper around `URL` to warn about the deprecation.

**Tasks**:

- [x] Code
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

I've added a dummy test to the end of the `wp-basic-tests` e2e tests, because if not Cypress was ignoring the previous test. I guess this is a bug in Cypress, maybe related to the `after` function.

### DIFF
--- a/e2e/packages/ads/src/index.tsx
+++ b/e2e/packages/ads/src/index.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Slot, connect, URL } from "frontity";
+import { Slot, connect } from "frontity";
 import { Connect } from "frontity/types";
 import TestAds, { Packages } from "../types";
 
--- a/e2e/packages/ads/src/index.tsx
+++ b/e2e/packages/ads/src/index.tsx
@@ -1,3 +1,8 @@
+/* eslint-disable */
+/*
+ * TSDocs will be added in this branch:
+ * https://github.com/frontity/frontity/tree/package-name
+ */
 import React from "react";
 import { Slot, connect } from "frontity";
 import { Connect } from "frontity/types";
--- a/e2e/plugins/index.js
+++ b/e2e/plugins/index.js
@@ -25,7 +25,7 @@ module.exports = (on, config) => {
     installPlugin({ name, version }) {
       return (async () => {
         await execa.command(
-          `docker-compose run wpcli wp plugin install ${name}${
+          `docker-compose run --rm wpcli wp plugin install ${name}${
             version ? ` --version=${version}` : ""
           }`,
           { stdio: "inherit" }
@@ -85,9 +85,12 @@ module.exports = (on, config) => {
      */
     removeAllPlugins() {
       return (async () => {
-        await execa.command(`docker-compose run wpcli wp plugin delete --all`, {
-          stdio: "inherit",
-        });
+        await execa.command(
+          `docker-compose run --rm wpcli wp plugin delete --all`,
+          {
+            stdio: "inherit",
+          }
+        );
         return null;
       })();
     },
--- a/packages/frontity/src/utils/decode/__tests__/decode.test.ts
+++ b/packages/frontity/src/utils/decode/__tests__/decode.test.ts
@@ -9,7 +9,7 @@ const mockedHe = he as jest.Mocked<typeof he>;
 describe("decode", () => {
   beforeEach(() => {
     mockedHe.decode.mockReset();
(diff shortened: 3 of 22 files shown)


e2e/packages/ads/src/index.tsx
```diff
@@ -1,5 +1,10 @@
+/* eslint-disable */
+/*
+ * TSDocs will be added in this branch:
+ * https://github.com/frontity/frontity/tree/package-name
```
michalczaplinski, line 4: ~~Is this correct? I see that the last commit on that branch was 8 days ago.~~

Oh, I see that it's in other files as well. I assume that you had a good reason for doing this but in general I would say that we shouldn't keep track of tasks this way because it's asking for trouble once it's merged and someone forgets 😅

e2e/plugins/index.js
```diff
@@ -25,7 +25,7 @@ module.exports = (on, config) => {
     installPlugin({ name, version }) {
       return (async () => {
         await execa.command(
-          `docker-compose run wpcli wp plugin install ${name}${
+          `docker-compose run --rm wpcli wp plugin install ${name}${
```
michalczaplinski, line 28: 👍 when running locally it leaves lotsa containers behind, good point.

packages/frontity/src/utils/decode/__tests__/decode.test.ts
```diff
@@ -9,7 +9,7 @@ const mockedHe = he as jest.Mocked<typeof he>;
 describe("decode", () => {
   beforeEach(() => {
     mockedHe.decode.mockReset();
-    mockedHe.decode.mockImplementation(require.requireActual("he").decode);
+    mockedHe.decode.mockImplementation(jest.requireActual("he").decode);
```
michalczaplinski, line 12: Just leaving this for reference: https://github.com/facebook/jest/pull/9854. This change is because of jest deprecating `require.requireActual()`

michalczaplinski (approved):

michalczaplinski (commented):
Not strictly related to the changes in the PR, but I've noticed that the `URL` constructor is not supported at all in IE so as far as I can tell that means that we don't support IE at all.

Based on https://community.frontity.org/t/frontity-support-for-ie9/2540/5?u=mmczaplinski we decided to support IE11 SSR-only so that would break this promise.

e2e/packages/ads/src/index.tsx
luisherranz: You're right, it's not a good approach. I should've copied the TSDocs from that branch 👍

luisherranz:
> Based on https://community.frontity.org/t/frontity-support-for-ie9/2540/5?u=mmczaplinski we decided to support IE11 SSR-only so that would break this promise.

The client-code it's not executed in IE because it's server-side only. So `URL` is safe to use because it will run only in the server 👍

This hasn't changed in this PR. We were already using `window.URL` for the client before: https://github.com/frontity/frontity/blob/dev/packages/core/src/config/webpack/externals.ts#L12

---

Thanks for the review and the comments @michalczaplinski! 🙂

michalczaplinski:
> The client-code it's not executed in IE because it's server-side only. So `URL` is safe to use because it will run only in the server 👍

oh, I think I was tired yesterday since the answer was in my own question basically 🤦‍♂️ Thanks @luisherranz :)

luisherranz:
> oh, I think I was tired yesterday since the answer was in my own question basically

Haha, that's true 😄

PR #544, opened 2020-08-25
Add Link types and mars example using a `frontity.settings.ts` file

luisherranz:
**Why**:

I was trying to figure out the "type merge" problem I explained here that @orballo discovered in the last [Infinite Scroll hooks PR](https://github.com/frontity/frontity/pull/533).

I reported the problem in the forum because it was caused by the merge I proposed here: https://community.frontity.org/t/typescript-use-two-different-types/1681/14

I noticed that the `Link` component didn't have proper types. This was something we haven't done before because the `Link` component is not a package, but still needs to get types from the packages it interacts with, like `Source` or `Router`.

So I went ahead to see how typing a component like `Link` would look like in Frontity. It turns out that we can use the same approach we are using for packages.

Interestingly, this package also exposes a new setting for the themes: `state.theme.autoPrefetch`. I also exported that type, so it can be used in theme types that use this component. I used the `@frontity/components/link/types` path and I think it makes sense.

I also refactored `Link` into different files because with the TSDocs it was getting huge. Due to that refactoring, I refactored the processing queue in a class.

Finally, I added types for MarsTheme to be able to test the `frontity.settings.ts` file (TypeScript). I used the `AutoPrefetch` type defined in `Link`. That part is explained in this video: https://www.loom.com/share/c73a7addf54e41ac9db6cd6bcb55dd2a

**How**:

Added created types for `Link` just like if it were a package and used those types in the component, just like components that belong to a package.

For the `frontity.settings.ts` file, I added a new example `mars-theme-ts-example`. It is not deployed to Now (because it will be exactly the same as the `mars-theme-example`).

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript

**Unrelated Tasks**

- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

There is no changeset because the functionality didn't change.


luisherranz:
@michalczaplinski this is ready for review.

Eslint is failing because of the problem you described in https://github.com/gajus/eslint-plugin-jsdoc/issues/598 so don't worry about that. I checked and it is fine 👍 

By the way, I see you fixed this in https://github.com/gajus/eslint-plugin-jsdoc/pull/599. Was it released?

packages/components/link/__tests__/link.tests.tsx
```diff
@@ -541,7 +542,7 @@ describe("Link prefetching", () => {
     act(() => {
       render(
         <Provider value={storeInViewMode}>
-          {links.map((link, i) => (
+          {links.map((link) => (
             <Link key={link} link={link} className={"my-link-${i}"}>
```
nicholasio: I noticed you removed the `i` variable; I believe `className` is broken now (i.e `my-link-undefined`)

packages/components/link/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+ * @param el - The element to watch for hover events.
+ * @param cb - The callback that should run should an hover event happen.
+ *
+ * @returns A callback that removes the event listeners.
+ */
+export const onHover = (el: HTMLAnchorElement, cb: () => void) => {
+  let timeout: ReturnType<typeof setTimeout> = null;
+
+  /**
+   * When the mouse is moved over the element, if delays the call of the
```
nicholasio:
```suggestion
   * When the mouse is moved over the element, it delays the call of the
```

packages/components/link/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+   */
+  const mouseOverHandler = () => {
+    timeout = setTimeout(() => {
+      cb();
+    }, config.hoverDelay);
+  };
+
+  /**
+   * Removes the timeout to avoid executing the callback if when the mouse is
+   * moved out of the element.
```
nicholasio:
```suggestion
   * Removes the timeout to avoid executing the callback if the cursor is
   * moved out of the element before `config.hoverDelay`
```

nicholasio (approved):
Nice work, just notice one potential bug and made some suggestions for some of the comments.

packages/components/link/__tests__/link.tests.tsx
michalczaplinski: Thanks @nicholasio. I think this showed up a TS warning originally because the `"my-link-${i}"` was not a template string, so the `i` was not actually use anywhere 😅  Fixed now 🙂

packages/components/link/utils.ts
michalczaplinski: 👍

packages/components/link/utils.ts
michalczaplinski: 👍

packages/components/link/__tests__/link.tests.tsx
nicholasio: oh yeah, that makes sense!

michalczaplinski (approved):
Looks great @luisherranz 👍 👍 

Super minor adjustments: 
- I've merged the suggestions about comment wording from @nicholasio as well as added some of mine.
- Fix minor issue in a `Link` test: https://github.com/frontity/frontity/pull/544#discussion_r476686805
- I've changed the version of `mars-theme-ts-example` to from `1.2.0` to `1.0.0`. I guess this was just a copy paste mistake.

michalczaplinski:
By the way, just wanted to note that [one of the test runs](https://github.com/frontity/frontity/pull/544/checks?check_run_id=1028269761#step:5:1146) failed with:

```
Cypress failed to make a connection to the Chrome DevTools Protocol after retrying for 50 seconds.

This usually indicates there was a problem opening the Chrome browser.

The CDP port requested was 43629.

Error details:

Error: connect ECONNREFUSED 127.0.0.1:43629
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1128:14)
```

However, it passed when I pushed another commit 🤷‍♂️

Let's keep an eye on it in the future. There's an issue open for this in the cypress repo with a suggested workaround: https://github.com/cypress-io/cypress/issues/6540#issuecomment-641087834

packages/components/link/__tests__/link.tests.tsx
luisherranz: Ups, absolutely my fault. Eslint complained that it was not used and I deleted it assuming Eslint was right. Sorry about that, and nice catch!

Thanks for the fix @michalczaplinski 👍

luisherranz:
Thanks for the reviews @michalczaplinski and @nicholasio  🙂

> There's an issue open for this in the cypress repo with a suggested workaround: cypress-io/cypress#6540 (comment)

I think I saw it once too. I've subscribed to the issue as well 👍

PR #545, opened 2020-08-26
Fix CLI env variable names and no-prompt option

### LINKED ISSUE #434: `npx frontity create-package <packageName> --no-prompt` should not prompt user for input
**Expected behavior**
When running `frontity create-package <packageName> --no-prompt`, it should not prompt user for any input but use the defaults for the package namespace.

**Observed behavior**

Despite the flag, the CLI still prompts the user for input: 

```
⬢  npx frontity create-package my-custom-theme --no-prompt
? Enter the namespace of the package: theme
✔ Adding package.json.
✔ Adding src/index.js.
✔ Installing package my-custom-theme.

New package "my-custom-theme" created.

⬢  npx frontity create-package my-custom-theme2
? Enter the namespace of the package: theme
✔ Adding package.json.
✔ Adding src/index.js.
✔ Installing package my-custom-theme2.

New package "my-custom-theme2" created.
```

```
> npx frontity create my-app
> cd my-app
my-app> npx frontity create-package my-package --no-prompt
```

**Possible solution**

The `create-package` command should take into account the default value for the package namespace

### LINKED ISSUE #538: Prompted even if `--no-prompt` option is set.
**_Bug report_**

Hello,
When passing the following command :

`$ npx frontity create --no-prompt --theme @frontity/mars-theme myapp`

or

`$ npx frontity create -n --theme @frontity/mars-theme myapp`

I'm prompted back

`? Do you want to receive framework updates by email? (y/N)`

**Expected behavior**

The cli should apply a default behaviour which must be set NO as answer to the question.

**Observed behavior**

A question is asked, while `--no-prompt` or `-n` option is set.

**Info about your system**

Reproduced on Ubuntu 20.04 and MacOS 10.15

**Possible solution**

Reading the code `packages/frontity/src/cli/create.ts` it seems that no default values are set for cli parameters if they are missing and -no-prompt is set.  
They are defined only if `promptUser` is True.  
Default values for name and theme should be defined as variables at top of the file, and not *hard-coded* in Question templates.
The value of `promptUser` must be evaluated before everything else and return with well set values if true.

I will try to submit a PR in the next days to fix it. It's possible that it will fix Issue #434 as well.

luisherranz:
**What**:

I've fixed the env variables of the CLI commands.

I also took the opportunity to refactor the code related to the `--no-prompt` arg and fix these two bugs: https://github.com/frontity/frontity/issues/538 and https://github.com/frontity/frontity/issues/434.

**Why**:

I was [working with @juanma on the explanation of how the CLI works with env variables](https://github.com/frontity/docs/issues/198) and I noticed that some of the CLI commands didn't have the correct env variable names.

If we don't fix that now, and it is documented, people can start using it, and then we could not fix it without introducing backward compatibility.

**How**:

- I fixed the env variable names.
- I refactored and fixed the `--no-prompt` code.
- I added TSDocs to all the CLI commands.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases): https://github.com/frontity/docs/issues/210

**Unrelated Tasks**

- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

As we don't have tests for the CLI yet, I've made some manual tests in my local environment.


michalczaplinski (approved):
LGTM 👍 

I've just only updated some comment wording in a couple places

luisherranz:
Thanks @michalczaplinski!

PR #548, opened 2020-08-31
Recreate Yoast package

### LINKED ISSUE #41: REST API - Head Tags doesn't work with Yoast SEO ^14.0 [5pt]
**_Bug report_**

Our plugin integration with the Yoast plugin stopped working since the release of its 14.0 version.

They made a major release and they have practically rewritten the whole code; all the functions and classes we are using are now deprecated. Having said that, we should rewrite `setup` and `reset` functions in the Yoast integration (nothing of the following code is valid):

https://github.com/frontity/wp-plugins/blob/f92193c6dc460012e9adfd4664d497ff404a82ca/plugins/frontity-headtags/includes/integrations/class-frontity-headtags-yoast.php#L23-L53

**Expected behavior**

I should include the Yoast meta tags in the `head_tags` field for all entities, not just for the first one.

**Observed behavior**

It only works for the first entity.

The current code is removing all hooks attached to the `wpseo_head` action in the `reset` function:

https://github.com/frontity/wp-plugins/blob/f92193c6dc460012e9adfd4664d497ff404a82ca/plugins/frontity-headtags/includes/integrations/class-frontity-headtags-yoast.php#L45

and calling this the Yoast reset function after that

https://github.com/frontity/wp-plugins/blob/f92193c6dc460012e9adfd4664d497ff404a82ca/plugins/frontity-headtags/includes/integrations/class-frontity-headtags-yoast.php#L53

which was adding all hooks once again, apart from other stuff. That function simply does nothing since the 14.0 version of Yoast SEO.

Just install the latest version of both Head Tags and Yoast SEO plugins.

**Possible solution**

I don't see an easy solution in this case. We will have to study the [Yoast SEO code](https://github.com/Yoast/wordpress-seo/tree/trunk/src) in-depth and look for public functions and classes we could use to get the meta tags rendered in the `wp_head` action for a given `WP_Query`.

At least their code has improved a lot! :smile:

DAreRodz:
**What**:

Release v2 of `@frontity/yoast` package.

A `@frontity/yoast` package was already released some time ago, and deprecated in favor of the `@frontity/head-tags` package.

Now that it is not possible to use it anymore with Yoast as [our Head Tags plugin does not support Yoast SEO ^14.0](https://github.com/frontity/wp-plugins/issues/41), the `@frontity/yoast` package is re-implemented in this PR to use the `yoast_head` field the Yoast SEO plugin already exposes in the REST API since that version.

**Why**:

To support Yoast SEO ^14.0.

**How**:

Using the `yoast_head` field present in the entities that come from the WordPress REST API.

As that field is in string format, it must be parsed first and converted to React. For that reason, this package depends on the `@frontity/html2react` package, which was modified to allow developers to pass an array of processors as a property, using them instead of those defined in `libraries.html2react.processors`.

It also reuses some functions from `@frontity/head-tags package to transform links in case the decoupled mode of Frontity is being used.

Closes https://github.com/frontity/wp-plugins/issues/41

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests for Yoast +14
- [x] End to end tests for `@frontity/head-tags` and Yoast -13
- [x] End to end tests for `@frontity/head-tags` and WP SEO
- [x] End to end tests for `@frontity/head-tags` and AllInOneSEOPack
- [x] Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases): https://github.com/frontity/docs/issues/220
- [x] Update community discussions (https://community.frontity.org/t/support-for-yoast-plugin-rest-api-fields/2626/13)
- [x] Move `getEntity()` from `head-tags` to `source`
- [x] Change `onlyInSSR` by `renderTags` property

**Unrelated Tasks**

- TypeScript tests
- Update starter themes
- Update other packages


luisherranz:
As part of this PR we could add some E2E tests for the `head-tags` package as well. What do you think @santosguillamot?

SantosGuillamot:
>As part of this PR we could add some E2E tests for the head-tags package as well. What do you think @SantosGuillamot?

I assume they are going to be pretty similar, so I guess it makes sense in that case. We should aim to complete all the tests that are lacking so I feel this is a good opportunity to do these ones.

We would need to do the same tests but use with different plugins: All in One, WordPress SEO and previous versions of Yoast. Right?

luisherranz:
> We would need to do the same tests but use with different plugins: All in One, WordPress SEO and previous versions of Yoast

Seems right 👍

I have updated the PR tasks.

DAreRodz:
OK, guys. Tests for everyone. :raised_hands:

e2e/support/commands.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable jest/valid-expect */
 
-import "cypress-plugin-snapshots/commands";
+require("cypress-plugin-snapshots/commands");
```
DAreRodz, line 3: This prevents an ESLint error.

e2e/cypress.json
```diff
@@ -11,5 +11,7 @@
   "ignoreTestFiles": ["**/__snapshots__/*", "**/__image_snapshots__/*"],
   "projectId": "ed7hed",
   "defaultCommandTimeout": 10000,
-  "taskTimeout": 60000
+  "taskTimeout": 60000,
+  "chromeWebSecurity": false,
```
DAreRodz, line 15: This is needed by the `cy.visit` command to work properly (that command was extended in e2e/support/commands.js).

e2e/cypress.json
```diff
@@ -11,5 +11,7 @@
...
+  "baseUrl": "http://localhost:3001"
```
DAreRodz: This line solves a problem I had with e2e tests starting again once all wp plugins were installed, reinstalling them again.

Also, not setting it is an [anti-pattern](https://docs.cypress.io/guides/references/best-practices.html#Setting-a-global-baseUrl).

luisherranz: Nice. Should we use `cy.visit("/xxx")` instead of `cy.visit("https://localhost:3001/xxx")` from now on?

luisherranz:
Hey @darerodz, I'm going to assume this is ready for review again 🙂 Let me know if it's not.

EDIT: Oh, you said so in the FD: https://community.frontity.org/t/support-for-yoast-plugin-rest-api-fields/2626/21?u=luisherranz. Thanks!! 😄

luisherranz (approved):
Amazing work @darerodz!! 👏👏

The only change I did was to use `useConnect` inside your hook to avoid passing `state` around. I've also taken the opportunity to move `state.source.api` to the types of `source` instead of `wp-source`. I don't know why we decided to have it there in the past but I don't think it makes sense anymore, especially when [we want to switch to `state.source.url`](https://community.frontity.org/t/make-the-backend-url-a-global-setting/2381/7).

Other than that, I think we cannot leave `state.source.entity()` without TypeScript coverage. We need to fix that, although I think we can merge this PR now and work on that in the same PR we are going to open for the `Data` TypeScript fix. I've added more info in the FD: https://community.frontity.org/t/support-for-yoast-plugin-rest-api-fields/2626/21

PR #550, opened 2020-09-01
Expose Webpack's errors and warnings in `frontity build`

luisherranz:
**What**:

I've added error handling to the `npx frontity build` command.

Fatal Webpack errors, like wrong configurations:

Screenshot 2020-09-01 at 12 01 19

Compilation errors, like missing modules, syntax errors, etc:

Screenshot 2020-09-01 at 12 02 08

Compilation warnings, like dynamic modules, performance issues, etc:

Screenshot 2020-09-01 at 12 03 07

**Why**:

It's something that we didn't do due to the lack of time when we launched the v1.0 but that it is necessary.

Specifically, it's making hard to see the real errors on the Vercel deploy because they don't seem to be reproducible in the local environment using `npx frontity dev --prod`: https://community.frontity.org/t/deploying-to-vercel-issues/2695/19

Apart from that, it's very important for CI's where the command needs to stop the deployment if something goes wrong.

**How**:

I modified the function that runs `webpack` programmatically to handle the different types of errors.

**Tasks**:

- [x] Code
- [x] TSDocs

**Unrelated Tasks**

- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

It's not tested because we don't have tests for the CLI yet. I thought about adding unit tests but I don't think they would be useful as the tests will be tightly coupled with the implementation.

### DIFF
--- a/.changeset/mighty-queens-leave.md
+++ b/.changeset/mighty-queens-leave.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": minor
+---
+
+Expose Webpack errors and warnings in the `npx frontity build` command.
--- a/.changeset/pink-fans-promise.md
+++ b/.changeset/pink-fans-promise.md
@@ -1,0 +1,5 @@
+---
+"frontity": patch
+---
+
+Update chalk version.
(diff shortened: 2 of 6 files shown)


DAreRodz (approved):
Everything seems to be fine. :+1:

luisherranz:
Thanks @DAreRodz! :)

PR #551, opened 2020-09-02
Add script to load Google Tag Manager container in its package

### LINKED ISSUE #536: Google Tag Manager click events aren't working in @frontity/google-tag-manager-analytics
**_Bug report_**

**Expected behavior**

After installing `@frontity/google-tag-manager-analytics`, I expect to be able to use the built-in click events triggers from Google Tag Manager.

**Observed behavior**

Although the pageviews are working properly, and if I send a custom event is working as well, the click triggers doesn't fire in a Frontity project.

1. Create a new Frontity project, install `@frontity/google-tag-manager-analytics` package and add an existing containerId.
2. Configure a Tag that is triggered on a click event in your Google Tag Manager workspace.
3. Visit your page and check that links are not working (you can use Google Tag Manager preview mode to do this).

**Possible solution**

I don't know a possible solution, but these are some insights after researching it a bit.

* At the beginning I thought it was a problem in frontity.org, but it seems it doesn't work in `twentytwenty-theme` and `mars-theme` after a clean create.
* I wasn't sure if `html2react` was affecting somehow but this doesn't seem to be the case. I removed all the components and add just a link tag that was working in another web, with the same trigger (community.frontity.org), and it wasn't working, so I assume is not `html2react` but Frontity. This is the HTML I was using for the tests:
```
<a
  href="https://docs.frontity.org/getting-started"
  title="Docs"
  target="_blank"
  id="home-button"
  className="text-menu-item"
  rel="noopener noreferrer"
>
  DOCS
</a>
```
* If I add an `onClick` property with an `actions.analytics.event( {"gtm.linkClick", {}})` it works.
* I investigated a little bit why the `onClick` works but the default click event doesn't and I found [this article](https://www.simoahava.com/analytics/dont-gtm-listeners-work/) interesting. It says that when a click event occurs on a link, the event starts to bubble up the DOM tree, until it reaches the topmost node (event propagation). Any elements along this bubbly path with listeners attached to them will fire when the event reaches them. However, if we use `onClick`, it waits for a click event on that specific tag.

This last point made me think that event propagation may affect somehow, and maybe one of the top nodes is stopping the propagation and making it fail.

SantosGuillamot:
**What**:

This Pull Request adds a `<script>` to the `@frontity/google-tag-manager-analytics` package to loads the container that is needed to listen to the default Google Tag Manager events like clicks.

**Why**:

This container, that isn't loaded right now, is needed to listen to the default Google Tag Manager events like clicks. There is more information about that in [this issue](https://github.com/frontity/frontity/issues/536#issuecomment-685465051).

**How**:

Adding the part of the [GTM recommend script](https://developers.google.com/tag-manager/quickstart) that is missing. Apart from that I've updated the snapshot for the unit tests and the indexes of the `dataLayer` array in the e2e tests.

**Tasks**:

- [x] Code

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

### DIFF
--- a/.changeset/tall-timers-mate.md
+++ b/.changeset/tall-timers-mate.md
@@ -1,0 +1,5 @@
+---
+"@frontity/google-tag-manager-analytics": patch
+---
+
+Add the script to load the Google Tag Manager container, that is needed to listen to default Google Tag Manager events.
--- a/e2e/integration/frontity-01/google-tag-manager-analytics.spec.js
+++ b/e2e/integration/frontity-01/google-tag-manager-analytics.spec.js
@@ -27,28 +27,28 @@ describe("Google Tag Manager", () => {
   });
 
   it("should have sent the first pageview", () => {
-    cy.window().its("dataLayer").its(0).should("deep.equal", pageviewHome);
+    cy.window().its("dataLayer").its(1).should("deep.equal", pageviewHome);
   });
 
   it("should sent a pageview if the page changes", () => {
     cy.get("button#change-link").click();
-    cy.window().its("dataLayer").its(1).should("deep.equal", pageviewSomePost);
+    cy.window().its("dataLayer").its(2).should("deep.equal", pageviewSomePost);
   });
 
   it("should sent pageviews when going back or forward", () => {
     cy.get("button#change-link").click();
     cy.go("back");
-    cy.window().its("dataLayer").its(2).should("deep.equal", pageviewHome);
+    cy.window().its("dataLayer").its(3).should("deep.equal", pageviewHome);
 
     cy.go("forward");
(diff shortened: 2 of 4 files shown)


SantosGuillamot:
@DAreRodz I am not sure if the changes I did to the tests are correct. The snapshot looks a bit weird, but it is what is returning right now. Apart from that, I had to update the indexes of the `dataLayer` array in the e2e tests because there is a new element now.

DAreRodz (approved):
I think it's fine, @SantosGuillamot.

The snapshots are correct and also updating the index for the events.

PR #564, opened 2020-09-08
Support the `?preview=true` query

### LINKED ISSUE #331: If a handler pattern isn't matched we should return a Server Error

### LINKED ISSUE #481: isHome should be false when isSearch is true
**_Bug report_**

**Expected behavior**

When we are in the home, but there is a search and `isSearch` is `true`, `isHome` should be `false`. This is how WordPress works.

**Observed behavior**

Right now, `isHome` remains `true` when `isSearch` is also `true`.

Note: This only happens for `isHome`. If you search inside a taxonomy or other archive, both `isCategory` and `isSearch` flags are `true`.

Do a search in the home, like `"/?s=something"`.

### LINKED ISSUE #562: Add e2e for the preview. [13pt]
**Feature:** [WordPress preview support](https://community.frontity.org/t/wordpress-preview-support/2419).
**User Story**: Be able to use a reliable version of the preview.

[**Sprint Card**](https://github.com/orgs/frontity/projects/9#card-44889149)

luisherranz:
**What**:

Add support for the `?preview=true` query in the post type handlers of `@frontity/wp-source`.

**Why**:

It means Frontity will be able to support the WordPress preview button, because the `preview` query indicates that the last revision should be loaded, instead of the published content.

**How**:

- Support the `preview=true` query in the `postType` handler.
- Add a new handler for post types with plain query permalinks (`?p=ID`) with support for the `preview=true` query.

  This handler is required because posts that are not published yet do not have a permalink and the URL of the preview uses the plain query permalink: `/?p=13&preview=true`. And another one for pages: `?page_id=ID`.

- Add a way to match queries in the `pattern`.

  For this I've added support for regular expressions in the `pattern` field of the handler, like `"RegExp:(\\?|&)preview=true"`. The `path-to-regexp` library we use for the normal `pattern` fields has a very limited support for queries. This change is in the `getMatch` function.

  I've also added support for named capture groups, but I still need to check if they work with our current Babel configuration. If they work, they will be exposed in the same way we are exposing the `path-to-regexp` params:

  - `/category/:slug`: `/category/nature` -> `{ params: { slug: "nature" } }`
  - `RegExp:\\?p=(?<id>\\d+)`: `/?p=13` -> `{ params: { id: 13 } }`

Minor changes:

- I've refactored the internal `wp-source` libraries to rename `path` (deprecated) and `route` (link without pagination) for `link`.
- I've added alphabetical sorting to the string query. For example, `?k1=v1&k2=v2` is now the same than `?k2=v2&k1=v1`.
- I've chaged how we compute `isHome` (although I still need to fix [this bug](https://github.com/frontity/frontity/issues/331)).
- I've added `queryString` to the list of link params we get/set using `libraries.source.parse` and `libraries.source.stringify`. It's sorted alphabetically.

**Tasks**:

- [x] Add a handler for pages with plain query permalinks: `?page_id=ID` - @DAreRodz
- [x] Fix the [If a handler pattern isn't matched we should return a Server Error](https://github.com/frontity/frontity/issues/331) bug _(because we've changed the way patterns match, I think it's a good moment to revisit this problem)_. - @DAreRodz
- [x] End to end tests - @DAreRodz
- [x] Test the plain query permalinks handler with custom post types - @luisherranz 
- [x] Check that we can use named capture groups in the regular expressions - @luisherranz
- [x] Add support for `preview=true` to the `postType` handler.
- [x] Add a way to match patterns with queries by allowing regular expressions in the pattern.
- [x] Add a handler for post types with plain query permalinks: `?p=ID`
- [x] Initialize plain query permalinks for custom post types.
- [x] Fix the [`isHome` is true when `isSearch` is true](https://github.com/frontity/frontity/issues/481) bug _(because we've changed the way `isHome` works, I think it's a good moment to revisit this problem)_ - @luisherranz
- [x] TSDocs
- [x] TypeScript
- [x] Fix Unit tests - @luisherranz

**Unfinished Tasks**

- [x] Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases): https://github.com/frontity/docs/issues/232

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

Out of the scope of this PR:

- Implement a standard way to add headers to the REST API request, as exposed in [Support for custom headers in Source packages](https://community.frontity.org/t/support-for-custom-headers-in-source-packages/2678).

  I will do this in a separate PR. Right now the code is hardcoded inside the handlers.

- Rename the `token` query.

  The name is to general. It should start at least with `frontity_` to avoid collisions and include more information to allow for other types of tokens and versions, like `frontity_preview_jwt_token_v1=XXX`.

- Modifications to the Embedded mode PHP plugin.

  I haven't done anything to the PHP plugin other than what David already did: https://github.com/frontity/frontity-embedded-proof-of-concept/tree/preview-poc


luisherranz:
I confirm that the [named capturing groups](https://babeljs.io/docs/en/babel-plugin-transform-named-capturing-groups-regex) babel plugin is installed and working.

luisherranz:
@DAreRodz I've added some changeset, but the one for the https://github.com/frontity/frontity/issues/331 bug is missing. Please add it yourself once you fix that bug 🙂

luisherranz:
I'm working on the "Test the plain query permalinks handler with custom post types" task.

luisherranz:
Finished.

@darerodz, I think we should add at least these e2e tests:

- Preview with unpublished post.
- Preview with published post.
- Preview with unpublished page.
- Preview with published page.
- Preview with unpublished custom post type.
- Preview with published custom post type.

What do you think?

DAreRodz:
I don't think of any other case we should test for the preview.

Those ones are fine, @luisherranz. :+1:

DAreRodz:
> @DAreRodz, I think we should add at least these e2e tests:

After implementing the handlers for posts and pages with the plain permalink I think we should differentiate between unpublished post/page and edited post/page, so e2e tests are going to be:

- Preview with an unpublished post.
- Preview with a published post.
- Preview with an edited post.
- Preview with an unpublished page.
- Preview with a published page.
- Preview with an edited page.
- Preview with unpublished custom post type.
- Preview with published custom post type.
- Preview with edited custom post type.

DAreRodz:
@luisherranz, ready to review!

luisherranz:
@santosguillamot, we have finished our work here.

I still have to update the community threads and the documentation issue, but I've moved those tasks to an "Unfinished tasks" section and I'll do them later.

You can approve and merge this PR and then do a release 🙂

SantosGuillamot (approved):
Great :slightly_smiling_face:

luisherranz:
I've created the docs issue and updated the community treads 👍

PR #568, opened 2020-09-12
Support for auth header in Source packages

### LINKED ISSUE #561: Support for auth headers in Source packages [8pt]
**Feature:** [Support for custom headers in Source packages](https://community.frontity.org/t/support-for-custom-headers-in-source-packages/2678).
**User Story**: Be able to access custom headers that source packages must send when requesting data.

[**Sprint Card**](https://github.com/orgs/frontity/projects/9#card-44888109)

michalczaplinski:
**What**:

Support for adding the auth header to source packages.

Following the Feature Discussion: https://community.frontity.org/t/support-for-custom-headers-in-source-packages/2678/12

**Why**:

So that users can e.g. add authentication headers to use Basic authentication.

**How**:

- [x] Allow the user to pass the auth info in `state.source.auth`
- [x] `state.source.auth` should be removed in the `afterSSR` action:
- [x] `frontity_source_auth` & `FRONTITY_SOURCE_AUTH` being passed to state.source.auth should be done in the `init()`. 
- [x] Add support for the `.env ` file
- [x] Rename `name=site` to `frontity_name=site`
- [x] Add the new (optional) `auth` param to `libraries.source.api.get()`
- [x] Remove frontity_source_auth and frontity_name from the `initialLink` and create a new field for those queries called `URLOptions`

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] Open an issue for this feature in the docs: https://github.com/frontity/docs/issues/228

**Unrelated Tasks**

- [x] End to end tests
- [x] TypeScript tests
- [x] Update starter themes
- [x] Update other packages


michalczaplinski (commented):
@DAreRodz You can review this :) 

Note that I have not added an e2e test to check if `state.source.auth` is indeed being removed in the `afterSSR` action because I thought that this would be easier to include in the e2e tests that you are writing for the preview.

I also have tested loading env variables from the `.env` files manually and it's 👌

packages/core/src/server/utils/initial-state.ts
```diff
@@ -2,13 +2,41 @@ import deepmerge from "deepmerge";
...
+}
+
+/**
+ * Initialize the state of a Frontity project.
+ *
+ * @param options - Defined in {@link StateOptions}.
+ * @returns The Frontity state object.
+ */
+const state = ({ settings, url }: StateOptions) => {
+  const paramsToDelete = ["frontity_source_auth", "frontity_name"];
```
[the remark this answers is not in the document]
luisherranz: I don't think that's a good idea because these options are meant to be used also by packages, like `frontity_source_auth`, and the blacklist is hardcoded in the core. So each time a package creator wants to use one of these options in their package, they'd have to do a PR to the core to add it.

Do you see any benefit on maintaining this blacklist?

packages/wp-source/src/actions.ts
```diff
@@ -122,6 +122,14 @@ const actions: WpSource["actions"]["source"] = {
 
     libraries.source.api.init({ api, isWpCom });
 
+    // If the URL contains an auth token, then add it to the state.
+    // This is normally the case e.g, when accessing the post preview.
+    const auth = state.frontity?.options?.frontity_source_auth;
+    const authFromEnv = process.env.FRONTITY_SOURCE_AUTH;
```
luisherranz: Interesting. Which one should take precedence here, the query, or the env variable? 🤔

luisherranz: Just to be sure, have you checked that the token doesn't end up in the `server.js` bundle?

It shouldn't, but I'd prefer to check it out to be sure 😅

packages/wp-source/src/libraries/__tests__/auth-header.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  // Initialize wp-source package
+  store.actions.source.init();
+
+  expect(store.state.source.auth).toBe("This is used indeed");
+});
+
+it.todo("should check that the token is erased correctly in afterSSR");
+
+it.todo("should check that we handle the frontity_name param correctly");
```
luisherranz: Did you forget to add these or they are meant to be added in a future PR?

luisherranz (changes requested):
Hey Michal, great work!! 👏👏👏

I know you assigned David for the review, but I wanted to review it as well because I added the "Frontity Query Options" to the implementation proposal at the last moment and I think that maybe it was not clear enough in some aspects.

I have one request regarding that, which is to not hardcode the Frontity Query Options in the core (deleting and types), to make them available to be used by packages. Other than that, I've left some other comments here and there with some suggestions and some questions for both you and David 🙂

packages/core/src/server/utils/__tests__/initial-state.tests.ts
```diff
@@ -87,4 +87,30 @@ describe("initialState", () => {
...
+    const url = new URL(
+      "https://site.com/page/2?frontity_name=mySite&frontity_source_auth=hello"
+    );
+
+    // The frontity_name & frontity_source_auth should be present in the options
+    expect(initialState({ settings, url }).frontity.options)
+      .toMatchInlineSnapshot(`
+      Object {
+        "frontity_name": "mySite",
+        "frontity_source_auth": "hello",
```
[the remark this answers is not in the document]
michalczaplinski: Hmm, I don't have a strong opinion because on one hand camelCase == consistency across frontity packages. 

On the other, keeping them in snake_case perhaps makes it more clear that they were part of the URL? So this way, they are consistent across the URL and the options object.

I lean towards agreeing with Luis here.

packages/wp-source/src/actions.ts
michalczaplinski: My thinking was that the value from the query string should take precedence over the value defined from the env variable. 

I think that the opposite would not really make sense. It would mean that if you set the FRONTITY_SOURCE_AUTH in your environment then you could not use the `frontity_source_auth` param anymore to set it.

packages/wp-source/types.ts
```diff
@@ -425,6 +442,11 @@ interface WpSource extends Source {
   };
 }
 
+/**
+ * Packages used internally by wp-source.
+ */
+export type Packages = WpSource & Package;
```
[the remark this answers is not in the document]
michalczaplinski: How would it be different from the current `Package` interface? I'm sorry, it's not clear to me.

packages/core/src/server/utils/__tests__/initial-state.tests.ts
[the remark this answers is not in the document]
luisherranz: I absolutely agree with you on both points. I don't have a strong opinion either, I just felt we should ask ourselves this question before making this public.

@darerodz, what do you think?

packages/wp-source/src/actions.ts
luisherranz: I see, thanks.

I was a bit afraid of the opposite case:

- You use the Frontity PHP plugin, which always sends the `frontity_source_auth` query.
- You use the `FRONTITY_SOURCE_AUTH` token to auth with the [JWT plugin](https://github.com/WP-API/jwt-auth) because you prefer that.
- `state.source.auth` is populated with the query.

But that's incidental, just because it happens that in this case, the token in the query has fewer permissions than then token in the env variable. The opposite case could also be true.

In our CLI we decided to use [the env variable only if the argument is not present](https://docs.frontity.org/frontity-cli/environment-variables) so I guess it makes sense to follow the same here.

All good then, thanks Michal!

packages/wp-source/types.ts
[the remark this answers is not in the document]
luisherranz: It would be something like this:

```ts
interface Frontity extends Package {
  state: {
    frontity: {
      url: string;
      name: string;
      initialLink: string;
      //...
    };
  };
}
```

That way you only use it when your package needs to know about those. Right now there's no actions or libraries in the core, but there could be in the future.

We would get rid of those types in `Package`.

And finally, if you need the Frontity state in your package, you do this:

```ts
export type Packages = MyPackage & WpSource & Frontity;
```

instead of this

```ts
export type Packages = MyPackage & WpSource & Package;
```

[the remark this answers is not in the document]
DAreRodz: I think is a good idea. :slightly_smiling_face: 

By the way, did we take any decision regarding how package types should be merged? Because the intersection operator (`&`) was not working with TypeScript 3.9, was it?

packages/core/src/server/utils/__tests__/initial-state.tests.ts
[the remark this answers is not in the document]
DAreRodz: Yeah, well, I would choose camelCase mainly because I'm more used to it. :camel: 

Some JavaScript libraries already transform `kebab-case` or `snake_case` into `camelCase` (like what React does with component properties) so I think it would not be hard to understand for developers why the options are in `snake_case` in the URL and in `camelCase` in JavaScript/TypeScript.

packages/wp-source/types.ts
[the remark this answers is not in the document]
michalczaplinski: @DAreRodz Yes, @luisherranz has proposed that we'll need a type utility to merge the types instead of using the `&` operator: https://community.frontity.org/t/typescript-use-two-different-types/1681/16

I've tried to create it and I came up with:

```ts
export type AnyArray<T = any> = Array<T> | ReadonlyArray<T>;

type Head<T extends AnyArray> = T["length"] extends 0 ? never : T[0];
type Tail<T extends AnyArray> = T["length"] extends 0
  ? never
  : ((...t: T) => void) extends (first: any, ...rest: infer Rest) => void
  ? Rest
  : never;

type Merge<T extends AnyArray> = Omit<Head<T>, 'name'> & Merge<Tail<T>>
```

The problem is that TS does not allow circular references (at least in some of the types) so it doesn't work 😕

packages/core/src/server/utils/__tests__/initial-state.tests.ts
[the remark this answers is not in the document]
michalczaplinski: okay, I ll transform it to 🐫Case in that case 🙂

packages/wp-source/src/libraries/__tests__/auth-header.test.ts
michalczaplinski: Whoops, yes,  I should have deleted those todos.

`"the token is erased correctly"` - I've mentioned here: https://github.com/frontity/frontity/pull/568#pullrequestreview-493025308 that we can test this in a e2e test

`"we handle the frontity_name param correctly"` - this was referring to the test cases that I wrote in `@frontity/core`.

packages/wp-source/src/actions.ts
michalczaplinski: nope, it doesn't end up in the bundle.

I've set the env variable locally to something like `1234567` and just manually searched the bundle for it.

packages/wp-source/types.ts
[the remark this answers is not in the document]
michalczaplinski: Let's continue the conversation in https://community.frontity.org/t/typescript-use-two-different-types/1681/16

.changeset/fuzzy-insects-warn.md
```diff
@@ -0,0 +1,7 @@
+---
+"@frontity/types": minor
+---
+
+Add types for frontity options (`state.frontity.options.frontity_name` & `state.frontity.options.frontity_source_auth`)
```
DAreRodz: Just a reminder!

When you have changed the frontity options to camelCase, you will have to update this text.

michalczaplinski: good catch, thanks 🙂

michalczaplinski:
The tests were also failing now because httpstat.us is down 🙁 

https://github.com/Readify/httpstatus/issues/89

So, I've just made a quick app on glitch which does the same thing as httpstat.us and it works fine.

e2e/packages/wp-source-errors/src/index.tsx
```diff
@@ -23,7 +23,8 @@ const wpSourceErrors: WpSourceErrors = {
       init: ({ state, libraries }) => {
         const { query } = libraries.source.parse(state.frontity.initialLink);
         if (query.statusCode) {
-          state.source.api = `http://httpstat.us/${query.statusCode}?rest_route=/`;
+          // This is a simple app made on glitch.me which just returns the HTTP status code passed as param.
+          state.source.api = `https://ballistic-western-donkey.glitch.me/status/${query.statusCode}?rest_route=/`;
```
michalczaplinski: The tests were also failing now because httpstat.us is down 🙁 

https://github.com/Readify/httpstatus/issues/89

So, I've just made a quick app on glitch which does the same thing as https://httpstat.us and it works fine.

michalczaplinski:
@luisherranz You're welcome to review this again.

@DAreRodz Pinging you if you want to have a look as well to see how to integrate it once merged.

packages/wp-source/src/libraries/__tests__/auth-header.test.ts
luisherranz: Perfect, thanks Michal.

packages/wp-source/src/actions.ts
luisherranz: Great, thanks.

e2e/packages/wp-source-errors/src/index.tsx
DAreRodz: Hey, I'm going to cherry-pick and add this to `dev` (tests are failing also in my PR :sweat_smile:).

packages/wp-source/src/actions.ts
```diff
@@ -122,6 +122,14 @@ const actions: WpSource["actions"]["source"] = {
 
     libraries.source.api.init({ api, isWpCom });
 
+    // If the URL contains an auth token, then add it to the state.
+    // This is normally the case e.g, when accessing the post preview.
+    const auth = state.frontity?.options?.frontitySourceAuth;
```
DAreRodz: The `frontity` prefix is removed from query options, isn't it?

Shouldn't it be `sourceAuth` instead `frontitySourceAuth`?

packages/wp-source/src/actions.ts
```diff
@@ -265,5 +273,9 @@ const actions: WpSource["actions"]["source"] = {
       });
     }
   },
+
+  afterSSR: ({ state }) => {
+    delete state.source.auth;
```
DAreRodz, line 279: The `frontity.options.sourceAuth` prop is not been removed here and thus it would be present in the state. I think it is not a big deal because the auth token would be already in the URL anyway. Just so you know.

packages/wp-source/src/libraries/__tests__/auth-header.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  // Initialize wp-source package
+  store.actions.source.init();
+
+  expect(store.state.source.auth).toBe(authToken);
+});
+
+it("should pass the frontitySourceAuth option to state.source.auth", async () => {
+  const authToken = "test-2";
+
+  // eslint-disable-next-line @typescript-eslint/camelcase
```
DAreRodz: I guess this ESLint comment is not needed anymore.

packages/wp-source/src/libraries/__tests__/auth-header.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  store.state.frontity = { options: { frontitySourceAuth: authToken } };
```
DAreRodz: Same as I mentioned before. Shouldn't it be `sourceAuth` instead of `frontitySourceAuth`?

packages/wp-source/src/libraries/__tests__/auth-header.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  expect(store.state.source.auth).toBe(authToken);
+});
+
+it("should use the value from `options` if both the state.source.auth and FRONTITY_SOURCE_AUTH are set", async () => {
+  // Add the token to the env variable
+  process.env.FRONTITY_SOURCE_AUTH = "This is not used";
+
+  store.state.frontity = {
+    options: {
+      // eslint-disable-next-line @typescript-eslint/camelcase
```
DAreRodz: This ESLint comment is not needed either.

DAreRodz (dismissed):
I was looking at the code to know what changes I have to do in https://github.com/frontity/frontity/pull/564 and I saw a couple of things. @michalczaplinski, check it out. :slightly_smiling_face:

PS: sorry for the comments being outside of the code review, haha. :smile:

luisherranz:
@michalczaplinski I'll take care of this so you can focus on the Smart Ads package 🙂

michalczaplinski:
> @michalczaplinski I'll take care of this so you can focus on the Smart Ads package 🙂

thanks @luisherranz ! 👍

packages/wp-source/src/actions.ts
luisherranz: Solved 👍

packages/wp-source/src/libraries/__tests__/auth-header.test.ts
luisherranz: Solved 👍

packages/wp-source/src/libraries/__tests__/auth-header.test.ts
luisherranz: Solved 👍

packages/wp-source/src/libraries/__tests__/auth-header.test.ts
luisherranz: Solved 👍

packages/wp-source/src/actions.ts
luisherranz: Good catch David. I think we should remove it as well.

luisherranz: Solved 👍

luisherranz (approved):
Awesome work Michal! 👏 😄

PR #579, opened 2020-09-29
Respect the `{ force: true }` parameter in source.fetch() in postType handler

### LINKED ISSUE #577: Actions Fetch Force Parameter Not Respected in wp-source Package
**_Bug report_**

**Expected behavior**

Setting the `force` parameter to `true` when fetching a post type should make it to be fetched again.

**Observed behavior**

The parameter seems to be ignored and the post is not fetched.

**Possible solution**

This [line](https://github.com/frontity/frontity/blob/114bd4bbe87c99d4384a5fb3eb338ae6d72b67a0/packages/wp-source/src/libraries/handlers/postType.ts#L52) in `postType.ts` doesn't take `force` into account when trying to fetch new data from a post or page. I discovered this in [this thread](https://community.frontity.org/t/how-to-render-password-and-then-protected-posts/2938/12?u=jonobr1) when trying to call `actions.source.fetch(url, { force: true });` on password protected pages.

I believe a simple fix could be amending line 52 to:
```javascript
if (!state.source.get(route).id || force) {
```
But, I don't know `wp-source` well enough to know if that would be the correct amendment. This is why I'm filing at as an issue instead of a PR.

michalczaplinski:
**What**:

Fix #577 . 

**Why**:

The `{force: true}` was not being respected in the postType handler.

**How**:

Fetch the data again if `force === true` in the postType handler.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] Update starter themes
- [ ] Update other packages
- [ ] TypeScript tests
- [ ] End to end tests

### DIFF
--- a/.changeset/smart-boxes-know.md
+++ b/.changeset/smart-boxes-know.md
@@ -1,0 +1,5 @@
+---
+"@frontity/wp-source": patch
+---
+
+Fix the issue where the `{ force: true }` option was not respected in the postType handler.
--- a/packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/post-type.tests.ts.snap
+++ b/packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/post-type.tests.ts.snap
@@ -228,7 +228,7 @@ Object {
       "isReady": false,
     },
     "/post-1/": Object {
-      "id": 1,
+      "id": 2,
       "isFetching": false,
       "isPost": true,
       "isPostType": true,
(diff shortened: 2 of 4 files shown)


michalczaplinski:
@jonobr1 Check it out

DAreRodz (approved):
I see what you did. Sometimes it's better not to rely on snapshots, isn't it? haha 😁

Approved!

michalczaplinski:
> I see what you did. Sometimes it's better not to rely on snapshots, isn't it? haha 😁
> 
> Approved!

Yeah, the snapshot was obviously wrong before!

I hesitated for a moment before running `git blame` because I think I knew that it would show my name. I wasn't wrong 😅

PR #593, opened 2020-10-05
Fix the Script component when unmount

### LINKED ISSUE #592: The `useEffect` hook inside the `Script` component should not return a value
**_Bug report_**

**Expected behavior**

When using the `<Script>` component from `@frontity/components` &ndash; usually from the `@frontity/html2react`'s `script` processor added by default &ndash; the content passed in the `code` property should work as long as the code is fine.

**Observed behavior**

If the `code` property contains code that returns a value when it is evaluated, the `useEffect` hook returns that value, and React shows a warning similar to this one:

```
Warning: An effect function must not return anything besides a function, which is used for clean-up. You returned: 1
```

When the script is unmounted, a `TypeError` is thrown and **the whole app crashes**.

**Possible solution**

The solution would be as simple as fixing this line so the return statement doesn't return any value.

https://github.com/frontity/frontity/blob/821ae43b7141e5e5ee1b6b3d61d7fc0d11d8477c/packages/components/script.tsx#L16

DAreRodz:
**What**:

Fix the `<Script>` component so its `useEffect` doesn't return the result of the evaluation of the `code` prop.

Fixes #592 

**Why**:

This causes a critical bug that could make the whole React tree disappear when a script is unmounted.

**How**:

Just fixing the line where the evaluation result was returned.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] End to end tests

**Unrelated Tasks**

- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/.changeset/itchy-trees-share.md
+++ b/.changeset/itchy-trees-share.md
@@ -2,6 +2,6 @@
 "@frontity/components": patch
 ---
 
-Fix an error that happen when an inline script evaluation returns a value.
+Fix an error that happens when an inline script evaluation returns a value.
 
 See [#592](https://github.com/frontity/frontity/issues/592).
--- a/.changeset/itchy-trees-share.md
+++ b/.changeset/itchy-trees-share.md
@@ -1,0 +1,7 @@
+---
+"@frontity/components": patch
+---
+
+Fix an error that happen when an inline script evaluation returns a value.
+
+See [#592](https://github.com/frontity/frontity/issues/592).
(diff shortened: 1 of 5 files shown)


luisherranz (approved):
Thanks @darerodz!

By the way, let's wait until Mario is back and see what he thinks, but I don't believe a warning can ever be a critical bug because it doesn't stop the app from working, does it?

DAreRodz:
My bad, I thought it was only a warning, but when the `<Script>` component is unmounted, the **whole app crashes** because it tries to call the `useEffect` returned callback, which is not a function.

I updated the bug description.

luisherranz:
Oh, ok. I guess React only has the check to see if the returned callback is a function in "development" mode to show a warning but it executes it anyway because in production it doesn't have a check. Makes sense, although that should have been a React error, not a React warning! 🙂

Anyway, thanks for solving this @darerodz.

PR #599, opened 2020-10-08
Fix `isHome` when `homepage` option is set.

### LINKED ISSUE #598: `data.isHome` is not working for front pages since wp-source@1.9.0
**_Bug report_**

**Expected behavior**

When the [`state.source.homepage` option](https://docs.frontity.org/api-reference-1/wordpress-source#state-source-homepage) of the `wp-source` package is set, the `isHome` property of the homepage should be `true` as expected.

**Observed behavior**

It returns `undefined` instead.

Thread in the community describing the issue: https://community.frontity.org/t/data-ishome-is-not-working-since-update/3039

**Possible solution**

It's a bug that sneaked into the commit https://github.com/frontity/frontity/commit/bdfeaace7823507314e3782b48887cf837bd11a2#diff-b4f2e2daaa31ef8b04cffb39926c83c1L71, when we fixed `isHome` for handlers that use regex patterns. The variable `route` used to check if the page being fetched is the homepage is modified when the `homepage` option is set because it uses an internal redirection to trigger the `postTypeHandler` instead of the `postArchiveHandler`. Instead of using `route` directly, we should use `source.data[link].route` for that check.

DAreRodz:
**What**:

Closes https://github.com/frontity/frontity/issues/598

**Why**:

Fixes a critical bug. Right now, `data.isHome` returns `undefined` for the homepage when the `homepage` option is set.  

**How**:

Fixing how `isHome` is computed. It was using the wrong `route` variable.

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] Update community discussions:
   - https://community.frontity.org/t/data-ishome-is-not-working-since-update/3039
   - https://community.frontity.org/t/bug-on-search-after-having-some-handlers/2921/27

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/.changeset/metal-trains-divide.md
+++ b/.changeset/metal-trains-divide.md
@@ -1,0 +1,5 @@
+---
+"@frontity/wp-source": patch
+---
+
+Fix `data.isHome` when the `state.source.homepage` option is set.
--- a/packages/wp-source/src/__tests__/actions.tests.ts
+++ b/packages/wp-source/src/__tests__/actions.tests.ts
@@ -128,6 +128,46 @@ describe("actions.source.fetch", () => {
     store.actions.source.fetch("/blog/page/123");
   });
 
+  test('should set isHome in "/" when a redirection has matched', (done) => {
+    store.libraries.source.redirections = [
+      {
+        name: "homepage",
+        priority: 10,
+        pattern: "/",
+        func: () => "/front-page/",
+      },
+    ];
+
+    observe(() => {
+      const data = store.state.source.get("/");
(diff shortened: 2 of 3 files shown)


michalczaplinski (approved):
Approved! 👍 

But I wanna continue the conversation about how to handle the `isHome` problem more generally in: https://community.frontity.org/t/data-ishome-is-not-working-since-update/3039/

PR #601, opened 2020-10-12
3xx Redirections

### LINKED ISSUE #608: Implement 301/302 redirections for redirects that are stored in WP database [13pt]
Feature: [301 redirects stored in WordPress database](https://community.frontity.org/t/301-redirects-stored-in-wordpress-database/3032).

[Sprint Card](https://github.com/orgs/frontity/projects/11#card-48312358)

michalczaplinski:
**What**:

Handle the 301/302/307/308 Redirections

**Why**:

Because some of the redirections can be stored in the WordPress database.

**How**:

Modify `actions.source.fetch()` and `actions.router.beforeSSR()` and `@frontity/core`.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests
- [x] Open an issue for this feature in the docs repo: https://github.com/frontity/docs/issues/253
- Add a couple of E2E tests for redirections:
  - [x] Defined directly in `state.source.data`
  - [x] Created on-the-fly by a handler

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


.eslintrc
```diff
@@ -109,20 +109,15 @@
...
+        "tsdoc/syntax": "off"
       }
     },
     // Overrides for Cypress test files.
     {
       "files": ["**/integration/**/*.spec.{js,ts}"],
       "rules": {
-        "jest/valid-expect-in-promise": "off"
+        "jest/valid-expect-in-promise": "off",
+        "jest/expect-expect": "off"
```
michalczaplinski, line 128: I've turned this off for all cypress tests because otherwise, we get errors from https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/expect-expect.md in many cypress tests.

e2e/integration/frontity-01/image.spec.js
```diff
@@ -70,9 +70,8 @@ describe("Image lazy-loading (with native lazy-load)", () => {
 
   it("should render an image without loading if it doesn't have height", () => {
     cy.scrollTo("topLeft");
-    cy.get("img:not([height])")
-      .should("not.have.attr", "loading")
-      .should("not.be.visible");
+    cy.get("img:not([height])").should("not.have.attr", "loading");
+    cy.get("img:not([height])").should("not.be.visible");
```
michalczaplinski: I am not certain why that is the case but this test has been failing when run locally when chaining the `.should()` calls:

```js
// doesn't work
cy.get("img:not([height])")
  .should("not.have.attr", "loading")
  .should("not.be.visible");
```

```js
// this works fine 
cy.get("img:not([height])").should("not.have.attr", "loading")
cy.get("img:not([height])").should("not.be.visible");
```

e2e/integration/wordpress-01/__snapshots__/yoast-package.spec.js.snap
```diff
@@ -876,3 +877,51 @@ exports[`Yoast Package > Custom Taxonomy > should render the schema tag #0`] = `
...
+<meta data-rh="true" name="twitter:data1" content="admin" />
+`;
+
+exports[`Yoast Package > CPT > should render the Twitter tags #3`] = `
+<meta data-rh="true" name="twitter:label2" content="Est. reading time" />
+`;
+
+exports[`Yoast Package > CPT > should render the Twitter tags #4`] = `
+<meta data-rh="true" name="twitter:data2" content="0 minutes" />
+`;
```
michalczaplinski: Those snapshots were missing, but because by default the cypress snapshot plugin runs with

```js
{ 
  "autopassNewSnapshots": true,    // Automatically save & pass new/non-existing snapshots
}
```

It looks like those snapshots were never generated and as a result the tests never failed on the CI.

e2e/integration/wordpress-01/__snapshots__/yoast-package.spec.js.snap
```diff
@@ -122,7 +122,8 @@ exports[`Yoast Package > Post > should render the schema tag #0`] = `
           "inLanguage": "en-US",
           "url": "http://2.gravatar.com/avatar/55502f40dc8b7c769880b10874abc9d0?s=96&d=mm&r=g",
           "caption": "admin"
-        }
+        },
+        "sameAs": ["http://localhost:8080"]
```
michalczaplinski: This seems to be part of the Yoast schema, e.g.: https://yoast.com/yoast-seo-11-0/ but I don't know why this snapshot has not failed previously and only failed now when run locally.

e2e/project/frontity.settings.ts
```diff
@@ -381,6 +416,7 @@ const settings: Settings<
           },
         },
       },
+      "@frontity/tiny-router",
```
michalczaplinski: Because `wp-source` depends on tiny-router

e2e/project/frontity.settings.ts
```diff
@@ -46,6 +50,10 @@ const settings: Settings<
         name: "@frontity/tiny-router",
         state: { router: { autoFetch: false } },
       },
+      {
+        name: "@frontity/wp-source",
+        state: { source: { url: "https://test.frontity.org/" } },
+      },
```
michalczaplinski: Because `tiny-router` has to access `state.source.get()` in: https://github.com/frontity/frontity/blob/6db0054df1a27813eb8888dd06e401480dccc5d2/packages/tiny-router/src/actions.ts#L52 we need to add both tiny-router and wp-source.

packages/core/src/server/index.tsx
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable @typescript-eslint/ban-ts-ignore, require-atomic-updates */
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
michalczaplinski: @typescript-eslint/ban-ts-ignore was replaced by ban-ts-comment in https://github.com/typescript-eslint/typescript-eslint/releases/tag/v3.0.0

packages/tiny-router/package.json
```diff
@@ -23,7 +23,8 @@
   "dependencies": {
     "@frontity/router": "^1.1.2",
     "@frontity/source": "^1.4.0",
-    "frontity": "^1.13.0"
+    "frontity": "^1.13.0",
+    "query-string": "^6.13.7"
```
michalczaplinski: The `wp-source` already has `query-string` as a dependency so this is not adding to the bundle size.

e2e/integration/wordpress-01/redirections.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    cy.get("#302-redirection").click();
+
+    cy.location("href").should(
+      "eq",
+      "http://localhost:3001/hello-world-redirected/"
+    );
+    cy.get("#post").should("exist");
+  });
+});
```
michalczaplinski, line 380: @luisherranz Just FYI: I'm still missing the test case for 307 & 308 Redirections here which should be analogous to the ones for 301 & 302 (for both client and server).

.eslintrc
luisherranz: Nice. Although I am going to leave the rule for Jest tests :+1:

packages/wp-comments/src/__tests__/commentsHandler.test.ts
```diff
@@ -12,7 +12,10 @@ let store: InitializedStore<Packages>;
 let api: jest.Mocked<Api>;
 
 beforeEach(() => {
-  const packages: any = mergeDeepRight(clone(wpSource()), clone(wpComments));
+  const packages: any = mergeDeepRight(
+    mergeDeepRight(clone(wpSource()), clone(wpComments)),
+    { state: { router: { redirections: "no" } } }
+  );
```
luisherranz: Why does `wp-comments` require setting `redirections` to "no"? Is not that the default value?

luisherranz (changes requested):
Hey, awesome work Michal. This was not an easy one.

I only see two improvements that we could make:

1. Move `state.router.redirections` to `state.source.redirections`:

I have been thinking about this and with this final implementation I think it doesn't make sense to have the settings in `state.router.redirections` anymore, but in `state.source.redirections`. I have explained why in the FD: https://community.frontity.org/t/301-redirects-stored-in-wordpress-database/3032/16?u=luisherranz

That would also mean that `wp-source` won't be dependent on `tiny-router` anymore, which is good. And we can get rid of all the extra code in the types and the tests.

2. Store the status in `redirectionStatus` and return it to the server.

Apart from that, I have left a few comments here and there.

luisherranz:
@michalczaplinski do you know where these snapshots are obsolete? https://github.com/frontity/frontity/pull/601/commits/17b4cd3d72a58159f53cf572442cfad7b29b1279

packages/tiny-router/src/actions.ts
```diff
@@ -115,17 +130,41 @@ export const beforeSSR: TinyRouter["actions"]["router"]["beforeSSR"] = ({
...
+  }
+
+  await actions.source.fetch(state.router.link);
+  const data = state.source.get(state.router.link);
+
+  // Re-create the Frontity Options.
+  const options = {};
+  for (const [key, value] of Object.entries(state?.frontity?.options || {})) {
+    options[`frontity_${key}`] = value;
+  }
```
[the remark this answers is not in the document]
michalczaplinski: yeah, I see. Agreed it's a good idea 👍

packages/tiny-router/src/actions.ts
```diff
@@ -115,17 +130,41 @@ export const beforeSSR: TinyRouter["actions"]["router"]["beforeSSR"] = ({
...
+
+  if (isRedirection(data)) {
+    // The query is always added to `data` and it's always an object, but
+    // can be empty if there were no search params.
+    const location =
+      Object.keys(data.query).length > 0
+        ? data.location + "&" + stringify(options)
+        : data.location + "?" + stringify(options);
+
+    ctx.redirect(location);
```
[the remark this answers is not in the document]
michalczaplinski: You're right, it was on my list of changes to make but it slipped through. Koa by default will set the status to 302 when you call the `ctx.redirect()` so we have to do this.

packages/wp-source/package.json
```diff
@@ -21,7 +21,9 @@
     "url": "https://github.com/frontity/frontity/issues"
   },
   "dependencies": {
+    "@frontity/router": "^1.1.1",
     "@frontity/source": "^1.4.0",
+    "@frontity/tiny-router": "^1.2.2",
```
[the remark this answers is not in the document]
michalczaplinski: This was because I was accessing the `state.router.redirections` in `source.actions.fetch()` here: https://github.com/frontity/frontity/blob/617e400e8c56444a3e773d9db4a9057da73fe87d/packages/wp-source/src/actions.ts#L82

However, I think I actually made a mistake and it is not necessary to include tiny-router as a dependency after all. I could have made and extra null check with the optional chaining operator like: `state?.router?.redirections` and if there is no router, then there are also no redirections. 

But that is all a moot point as we are going to move the redirections to `state.source.redirections` so this dependency can be removed.

packages/wp-comments/src/__tests__/commentsHandler.test.ts
michalczaplinski: The tests in wp-comments are calling `actions.source.fetch()` which is accessing `state.router.redirections`. This is the same case as mentioned in https://github.com/frontity/frontity/pull/601#discussion_r526495008.

It can be removed now that we're moving the redirections to `state.source.redirections`.

packages/wp-source/src/actions.ts
```diff
@@ -105,6 +145,48 @@ const actions: WpSource["actions"]["source"] = {
...
+          }
+
+          if (head?.redirected) {
+            const pathname = new URL(head.url).pathname;
+
+            const searchParams = stringify(query);
+
+            const location = `${normalize(pathname)}${
+              searchParams && "?"
+            }${searchParams}`;
```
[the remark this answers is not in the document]
michalczaplinski: My thinking was that the redirections should match some pathname like `/url/some-post` and always redirect for that pathname regardless of the query or hash. But you are right that a user can set up a redirection which includes a query string or hash as well like: `/url/some-post?key=value` --> `/url/some-new-post`. We have to add code to account for that use case.

But I still think that we should preserve the query and hash when doing a redirection if there is no query or hash present in the redirected link. Otherwise, the user would have to set up a redirection for every possible query and hash that can be added to the link and I think that might not always be practical.

So, I think that we have 3 cases here:

1. `/url/some-post` --> `/url/redirected` (no query or hash)
  
   In this case, we should preserve the hash and query if they are present in the link, because and redirect the user. So, `/url/some-post?key=value` redirects to `/url/redirected?key=value`.

2. `/url/some-post?key=value` --> `/url/redirected` (original link has a query)

   If the original link includes query or hash, we have to extract it and omit it from the query params in the redirected link. So, `/url/some-post?key=value&p=123` redirects to `/url/redirected?p=123`.

3. `/url/some-post` --> `/url/redirected?key=value` (The redirected link has hash or query)

   This is simple and similar to case 1: We add whatever query and hash is present on the link when accessing it, just making sure to use the `&` instead of `?` when glueing together the final link.

michalczaplinski:
> @michalczaplinski do you know where these snapshots are obsolete? [17b4cd3](https://github.com/frontity/frontity/commit/17b4cd3d72a58159f53cf572442cfad7b29b1279)

I have no idea but I ll have a look into it.

michalczaplinski:
> @michalczaplinski do you know where these snapshots are obsolete? [17b4cd3](https://github.com/frontity/frontity/commit/17b4cd3d72a58159f53cf572442cfad7b29b1279)

Oh, it looks like you've deleted those test cases in 617e400e8c56444a3e773d9db4a9057da73fe87d 🙂  I assume this was an error then?

packages/wp-source/package.json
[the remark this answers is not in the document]
luisherranz: Yes, I think it can be removed!

But if we were going to keep it I think the dependency should be with `@frontity/router` instead of `@frontity/tiny-router` because if I am not mistaken all you want are the types 🙂

packages/wp-source/src/actions.ts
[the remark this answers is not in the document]
luisherranz: I am sorry but I don't think that is right Michal. We should not mess with the `Location` that is returned by the server. If they want to preserve all the queries, part of the queries, or none of the queries, that is up to the redirection logic.

Take into account that in the "server-side rendered world" there is no "extra step" like we have here to modify the redirection. So whatever is returned by the `Location` header is used by the browser.

Redirections, when set properly, take into account the queries and pass them to the next link:

- Apache: `RedirectMatch ^/images/(.*)$ http://images.example.com/$1`
- NGINX: `rewrite ^/images/(.*)$ http://docs.domain2.com/$1 redirect;`

But this the standard, they can be way more complex. This one changes the value of `some-query` when it is `some-value`:

```nginx
location /some-path {
   if ($args ~* "(.*)some-query=some-value(.*)") {
       set $args "$1some-query=other-value$2";
       return 301 $scheme://$host$uri$args;
   }

}
```

Vercel and Netlify have simpler redirection setups where they pass the queries by default, but Netlify also has a more complex way to define the redirections where you can modify the queries: https://docs.netlify.com/routing/redirects/redirect-options/#query-parameters

The Redirections plugin has options for what to do with the queries as well. And of course, you can also use RegExps.

Screenshot 2020-11-19 at 08 59 37

So I really think we should respect the `Location` header (with the exception of the `frontity_` queries in the server, of course).

packages/wp-comments/src/__tests__/commentsHandler.test.ts
luisherranz: Ok, thanks for the explanation 🙂

luisherranz:
> Oh, it looks like you've deleted those test cases in 617e400 🙂 I assume this was an error then?

But did you delete them for a reason or was it a mistake?

michalczaplinski:
> > Oh, it looks like you've deleted those test cases in [617e400](https://github.com/frontity/frontity/commit/617e400e8c56444a3e773d9db4a9057da73fe87d) 🙂 I assume this was an error then?
> 
> But did you delete them for a reason or was it a mistake?

oh, looks that you have accidentally deleted the test cases that create those snapshots in 617e400 so that's why the snapshots are gone 😅

packages/wp-source/package.json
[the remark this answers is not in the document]
michalczaplinski: > But if we were going to keep it I think the dependency should be with @frontity/router instead of @frontity/tiny-router because if I am not mistaken all you want are the types 🙂

yes, exactly :) I'm not sure why I added the tiny-router as well as the types TBH.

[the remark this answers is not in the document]
luisherranz: Ok. No problem and thanks Michal!

e2e/project/frontity.settings.ts
michalczaplinski: That is not the case anymore

michalczaplinski:
@luisherranz You can review it again, I think it should be pretty good to merge now.

Very brief overview of the changes:
- Moved the `redirections` from `state.router.redirections` to `state.source.redirections`. 
- Store the status returned from the redirection in `redirectionStatus`.
- I've decided to not do anything with `libraries.source.redirections` in this PR. We can remove it in the docs or rename it, we can continue discussing in the FD: https://community.frontity.org/t/301-redirects-stored-in-wordpress-database/3032/12
- Restored the test cases and snapshots that got deleted.
- Removed the dependency on `router` & `tiny-router` from `wp-source`, so that allowed me to simplify the `frontity.settings.ts` in the e2e tests.
- I've added additional e2e tests to check how redirections work with the query string:
  - When the redirection is matching a route with a query param in it.
  - When the redirection has the "Ignore and pass params to target" setting.

michalczaplinski:
@luisherranz There is one small detail that I did not add and I was curious of your opinion. 

If the redirections === `all` but the original fetch does not fail, we do not actually cancel the `fetch()` of the redirection. I'm not sure if that's a potential problem or not. If it's not, it's maybe not worth the effort.

On the other hand, it would not be hard to use the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) to cancel that promise. We would have to use a package though, because `AbortController` is not available in node, but the implementation is only about 300B: https://bundlephobia.com/result?p=abort-controller@3.0.0

packages/wp-source/src/actions.ts
```diff
@@ -93,11 +90,11 @@ const actions: WpSource["actions"]["source"] = {
           });
         }
         // Or it can be "all".
-      } else if (routerRedirections === "all") {
+      } else if (redirections === "all") {
         redirectionPromise = fetch(state.source.url + link, { method: "HEAD" });
```
luisherranz: Uhm... `state.source.url` can contain a trailing slash, so maybe this is a problem in servers that do not handle double slashes:

- `state.source.url` = https://domain.com/
- `link` = /some-post
- The fetch is done to https://domain.com//some-post, which could result in a 404 or a 30X in some server configurations.

In Vercel, it returns a 308:

Screenshot 2020-11-23 at 09 10 17

I guess we better handle that case?

```js
fetch(state.source.url.replace(/\/$/, "") + link, { method: "HEAD" });
```

Maybe we need to improve this in the future with a hook, to make sure people don't need to do this over and over.

_By the way, there are more places where this is used. Maybe we should check them all._

packages/wp-source/types.ts
```diff
@@ -452,7 +451,7 @@ interface WpSource extends Source<Packages> {
 /**
  * Packages used internally by wp-source.
  */
-export type Packages = WpSource & Package & Router;
+export type Packages = WpSource & Package;
```
luisherranz: We can use `Frontity` now. It is more explicit 🙂

```js
import { Frontity } from "frontity/types";

export type Packages = WpSource & Frontity;
```

luisherranz (changes requested):
> If the redirections === all but the original fetch does not fail, we do not actually cancel the fetch() of the redirection. I'm not sure if that's a potential problem or not.

Oh, you are absolutely right, we cannot do those requests in parallel, the redirection check needs to be done first, and only once that has finished, we do the handler fetch.

If not, we would be causing a race condition. Imagine this case:

1. The URL `/some-post` contains a real post (it is not a 404).
2. The user uses `redirections: all` and sets a redirection to `/other-post` in the Redirections plugin.

Then, this could happen in the browser:

- A user visits the home.
- A user clicks on a link to `/some-post`.
- Both the fetch for the redirection check and the REST API start.
- The REST API finishes first, and `actions.source.fetch` populates the data object.
- The user sees the old `/some-post` instead of the new `/other-post`.
- The redirection ends (I am not sure what happens at this point, if the user gets redirected to `/other-post` or if this is ignored).

---

EDIT: Ok, it looks like the redirection is ignored if the REST API fetch doesn't fail. This means the `redirections === "all"` option is not useful right now. I think it needs to be handled serially:

- Check if there is a redirection for `/some-post`:
  - Yes: populate the data object with 301.
  - No: use the handler to populate the data object.

I know that performance-wise this is horrible, but right now there is no other way to do this because the handlers mutate the state directly.

This is something that is on the list of things to change in Source v2: Handlers should not modify the state directly. Instead, they should return the data objects and entities to `actions.source.fetch`. That way, `actions.source.fetch` has the control to mutate the state whenever it wants. So once we make that change we could do this redirection fetch in parallel. But until then, it needs to be done serially.

EDIT 2: Can you please add a test for the case when the URL exists but it has a redirection?

packages/wp-source/src/__tests__/3xx-redirections.tests.ts
```diff
@@ -0,0 +1,457 @@
...
+    )
+  );
+
+  handler = {
+    name: "always",
+    priority: 10,
+    pattern: "/(.*)",
+    func: jest.fn(async ({ route, state }) => {
+      await Promise.resolve();
+      Object.assign(state.source.data[route], {
```
luisherranz: Oh, better to use `link` here, `route` is deprecated 🙂

```js
func: jest.fn(async ({ link, state }) => {
      await Promise.resolve();
      Object.assign(state.source.data[link], {
```

michalczaplinski:
> EDIT: Ok, it looks like the redirection is ignored if the REST API fetch doesn't fail. This means the `redirections === "all"` option is not useful right now. I think it needs to be handled serially:

I see, you're right! 

I incorrectly assumed that when you add a redirection, the post should not be available in the REST API anymore under the `/old-slug` but that is not the case. It's only the case if you update the post slug directly. If you just add a redirection in the plugin, it doesn't work like that.


> EDIT 2: Can you please add a test for the case when the URL exists but it has a redirection?

👍

michalczaplinski:
@luisherranz Just wanted to clarify that the behaviour in the case when `state.source.redirections` is a RegExp or RegExp[] should change as well as for when `state.source.redirections === "all"`, correct?

By that I mean: 

If the redirections is a RegExp or RegExp[] AND the current link matches that RegExp we should do as you outlined earlier:
- Check if there is a redirection for /some-post:
  - Yes: populate the data object with 301.
  - No: use the handler to populate the data object.

luisherranz (changes requested):
I have noticed another problem in the design: `redirectionResponse.status` is always 200.

The redirection status is not exposed to the client due to some security issues: https://fetch.spec.whatwg.org/#atomic-http-redirect-handling

So we have to use different logic for the client and the server:

```js
if (state.frontity.platform === "server") {
  redirectionResponse = await fetch(state.source.url + link, {
    method: "HEAD",
    redirect: "manual",
  });
  isRedirection = /30[1278]/.test(redirectionResponse.status);
  redirectionStatus = redirectionResponse.status;
} else {
  redirectionResponse = await fetch(state.source.url + link, {
    method: "HEAD",
  });
  isRedirection = redirectionResponse.redirected;
  redirectionStatus = 301; // Default value.
}
```

I have explained it in more detail in these videos:

- https://www.loom.com/share/dd6308199b2b46f88c591a337baa0f3e
- https://www.loom.com/share/de16193ad31d45c98f0e780fa495d940

**And very important**: if I am right with this one, the status response from the server for any redirection is 200. That means that the E2E tests should be failing, but they are passing. Could you please fix that?

michalczaplinski:
I've followed this but I've just realized now that while yes you get the status of the request on the server when you make the request with `{ redirect: 'manual' }`, you cannot get the final location header on the server 😕

I looked at the spec again and it states that the redirect should actually have status of 0: https://fetch.spec.whatwg.org/#concept-filtered-response-opaque-redirect so this looks like the node-fetch implementation deviates from the spec in that regard.

I ll go ahead and try the approach of creating the separate `server` and `client` files like you mentioned earlier.

michalczaplinski:
Update:

I've experimented a little further and I've realized that it is still possible to use node-fetch and get both the link and the status of the redirection because node-fetch helpfully populates the `location` header with the url of the redirection (which is also different from the spec): https://github.com/node-fetch/node-fetch/blob/0d35ddbf73/src/index.js#L117

So we have to do something like: 

```js
let response: Response;
let isRedirection: boolean;
let status: number;
let location: string;

// On the server we have to fetch with `redirect: manual` so that we can get
// the 30x status code of the redirection.
if (platform === "server") {
  response = await fetch(link, {
    method: "HEAD",
    redirect: "manual",
  });
  isRedirection = /30[1278]/.test(response.status.toString());
  status = response.status;
  location = response.headers.get("location");
} else {
  // On the client it's not possible to get the actual status code of the
  // redirection because of https://fetch.spec.whatwg.org/#atomic-http-redirect-handling
  // The `redirect: manual` option on the client returns an "opaque response"
  // object which always has the status of 0.
  response = await fetch(link, {
    method: "HEAD",
  });
  isRedirection = response.redirected || false;
  status = 301; // Default value.
  location = response.url;
}

```

luisherranz:
I think you missed my second video, where I explain that 🙂 !
https://www.loom.com/share/de16193ad31d45c98f0e780fa495d940

luisherranz:
Could you please add a couple of e2e tests for:

1. Redirections defined directly in the state:

```js
const state = {
  source: {
    data: {
      "/url-with-redirection": {
        isRedirection: true,
        // ...
      },
    },
  },
};
```

2. Redirections defined on-the-fly by a handler:

```js
const handler = {
  pattern: "/urls-with-redirections/:slug",
  func: ({ state, link }) => {
    state.source.data[link] = {
      isRedirection: true,
      // ...
    }
  }
}
```

luisherranz:
> I've experimented a little further and I've realized that it is still possible to use node-fetch and get both the link and the status of the redirection because node-fetch helpfully populates the location header with the url of the redirection

Oh, ok, you meant using `response.url` and `response.headers.get("location")` 👍 🙂

michalczaplinski:
> Oh, ok, you meant using `response.url` and `response.headers.get("location")` 👍 🙂

Yup, sorry, I think I didn't make that clear.

package.json
```diff
@@ -35,7 +35,7 @@
     "prettier": "^2.0.1",
     "replace-in-files-cli": "^1.0.0",
     "ts-jest": "^26.2.0",
-    "typescript": "^4.0.5"
+    "typescript": "4.0.3"
```
michalczaplinski: I've pinned the version of typescript because TS `4.1+` is not working with ts-jest and causing tests to fail.

Officially, the version `4.0+` is not supported by ts-jest, but `4.0.x` they are working on this. Until it's resolved upstream, we should keep this pinned.

Here's the ts-jest issue: https://github.com/kulshekhar/ts-jest/issues/1887

michalczaplinski:
I'll let you re-review this @luisherranz even though David is the supporter because you've reviewed it before.

There is one outstanding (tiny) issue that I wasn't able to solve: On the client the 307 temporary redirect fails to fetch the redirection from WordPress because of a CORS issue. This is only on the client and only for 307 and I'm still not sure why it is happening. 

Here's a Loom showing it in practice: https://www.loom.com/share/39d6aa4c3c2148da8969550e6328a701

luisherranz (changes requested):
Michal, I am really sorry but I have just noticed that there is yet a use case that we have not taken into account: redirections to external domains.

Imagine a redirection from `https://my-domain.com/my-post` to `https://external-domain.com/another-post`. The way the logic is coded right now, the user will end up in `https://my-domain.com/another-post` instead of `https://external-domain.com/another-post`.

These needs to work both in the server, returning the proper external URL in the `location` header, and in the client, by using `window.location.replace(link)` instead of `actions.router.set(link, { method: "replace" })`.

I think we can do the match against both `state.frontity.url` and `state.source.url` because they are kind of "the same domain" in this case.

Maybe what we could do is to bake that logic in the data object so packages using the redirection don't have to do the match against `state.frontity.url`/`state.source.url` themselves.

So what if we add a new prop to the redirection object called `type: "external" | "internal"` os `isExternal: boolean`?

@michalczaplinski if you agree, could you please add that logic add E2E tests for this use case?

packages/tiny-router/src/actions.ts
```diff
@@ -83,6 +90,20 @@ export const init: TinyRouter["actions"]["router"]["init"] = ({
...
+    // Observe the current data object. If it is ever a redirection, replace the
+    // current link with the new one.
+    observe(() => {
+      const data = state.source?.get(state.router.link);
+      if (data && isRedirection(data)) {
+        actions.router.set(data.location, {
+          // Use "replace" to keep browser history consistent.
+          method: "replace",
+          // Keep the same history.state that the old link had.
+          state: state.router.state,
```
michalczaplinski: @luisherranz I've discovered that passing the state back to the router will fail if the state is already proxified.

It seems that you cannot pass a `Proxy` object as the `state` parameter, you can try it yourself:

```
> const proxy = new Proxy({}, {})
undefined
> window.history.replaceState(p, "", window.location.href);
VM178:1 Uncaught DOMException: Failed to execute 'replaceState' on 'History': [object Object] could not be cloned.
    at <anonymous>:1:16
```

I think we have two options. We could either "clone" the state object as suggested in https://stackoverflow.com/questions/53102700/how-do-i-turn-an-es6-proxy-back-into-a-plain-object-pojo but I think that a better solution would be to add an "escape" hatch like discussed in https://stackoverflow.com/a/60668790/2638310. We add a special symbol to the state which returns the "original", unproxied object. Like:

```
const original = {foo: 'bar'};

const handler = {
  get(target, prop) {
    if (prop === Symbol.for('ORIGINAL')) return target;
    return Reflect.get(target, prop);
  }
};

const proxy = new Proxy(original, handler);
const unproxied = proxy[Symbol.for('ORIGINAL')];

console.assert(original === unproxied); // true
```

I think that this should be an implementation detail for now, but maybe in the future we could expose a more ergonomic API for this in `@frontity/connect`

What do you think?

luisherranz: You are absolutely right.

@orballo already stumbled upon this problem in the `infinite-scroll-beta` branchan and we decided to go with the `clone` approach because it is safer.

`react-easy-state`/`@frontity/connect` has a [`raw()` utility to expose the original object](https://github.com/frontity/frontity/blob/dev/packages/connect/src/index.js#L2), but it is not deep, so would have to create a `deepRaw()` that iterates over the object extracting all the originals, in case there are proxies inside proxies.

So we decided to go with the `clone` approach for now, as you can see in the `actions.ts` file of `tiny-router` of the `infinite-scroll-beta` branch: https://github.com/frontity/frontity/blob/infinite-scroll-beta/packages/tiny-router/src/actions.ts.

michalczaplinski:
@luisherranz 

Here's the status at the moment. The last remaining feature is to add support for redirecting to  external domains.

- I've added support for this on the **server** according to your comment here: https://github.com/frontity/frontity/pull/601#pullrequestreview-542104143 because I agree that it's a good idea to add an `isExternal` flag. There is only one passing test created, I think it would need at least a couple more. You can see all the changes related to implementing the support for external domains in ee62da9659af9f26caedb6f9fd7c5080d9a5985d commit
- Support for redirecting to external domains still remains to be implemented on the client.
- I have also noticed that there are some failing snapshots in the yoast package: https://github.com/frontity/frontity/pull/601/checks?check_run_id=1579699686#step:5:2184 but I didn't manage to investigate it further.
- With respect to the discussion about "passing the state back to the router will fail if the state is already proxified.": I have not used a `clone()` utility in order not to add additional bytes to the bundle, but rather simply done `JSON.parse(JSON.stringify(router.state))`. It seems to work just fine and simply strips out any non-serializable items.

luisherranz:
I think we are going to store the full URL in `data.location`, including the domain, because we are also storing `isExternal`.

It is a bit worse for the consumer because if it is an internal link it needs to discard the domain, but if we strip the domain away there is no way for a package to differentiate between links that have redirections set up for the backend or the frontend.

luisherranz:
It seems like we have a problem with external redirections in the client. I have explanied it in detail here: https://www.loom.com/share/adc741cdcdd04ef9a42d9c09eb806d3b

To be honest, right now I don't see that this is a problem that can be solved because CORS and obscure redirections are two security measures the browsers impose. Finding a way to do this would mean finding a way to bypass that security.

_By the way, I tried capturing the error returned by the failed `fetch` but it only says "TypeError: Failed to fetch" as I supposed._

@santosguillamot, @darerodz, @cristianbote: My take is that we should go ahead and release the redirections without this feature. But before that, I want to know if you have an idea about how to solve this 🙂

DAreRodz:
Oh, I thought that using `fetch(link, { method: "HEAD" })` you would get a `301` response, and so you could stop if a link in between was an external link, but the actual browser behavior is to follow all the redirections until the last one, right?

I don't know if we can do anything here. :shrug:

SantosGuillamot:
Oh, if it isn't possible I guess it isn't a big deal and we should go ahead. Is there a way to force SSR if a client-side redirection fails?

luisherranz:
> Is there a way to force SSR if a client-side redirection fails?

The problem is that we don't know if there is a redirection or not. All that we have is a `TypeError: Failed to fetch` error, but that doesn't mean that there is a 301.

---

These are my thoughts right now: 

1. My guess is that doing 301 redirections to websites that you don't own is not going to be common.

  The most common scenario will be doing a redirection from `mydomain.com/blog` to `blog.mydomain.com` for example. 

  In that case, the user owns both sites and can add CORS to them.

2. The only way to solve this is by proxying the request from a server.

  So if we find out that people do actually need to do 301 redirections to external sites they don't control, we can solve this with the Frontity server itself: 

  - We add a `/frontity/source/is-redirection/` endpoint using the upcoming Server Extensibility and we modify source to use that instead of doing the request on the clients-side.

---

Ok, so now, to finish this PR, I need an additional server with a different URL configured with CORS for the e2e tests.

We only have localhost:3001 and localhost:8080, so I'll try to figure out a way to do it. I wouldn't like to use an external server to not add an external dependency again. 

Any ideas are welcomed 😄

luisherranz:
> Ok, so now, to finish this PR, I need an additional server with a different URL configured with CORS for the e2e tests.

My bad, the test is passing because we have the `chromeWebSecurity` setting of Cypress turned off.

@darerodz, you turned it off here and left a comment: https://github.com/frontity/frontity/pull/548/files#r485459330

I wonder if we should try to turn it on again, because tests that should be failing can pass, or just leave it as it is.

What do you think?

luisherranz:
Ok, we are going to turn it on again and Cristian is going to investigate if we can either turn it off for the yoast tests using something like `Cypress.config()` or use a different approach for the Yoast SSRing tests.

I will try to see how to make that final test pass in this PR, doing the 301 redirection to an external server that has CORS.

DAreRodz:
> The problem is that we don't know if there is a redirection or not. All that we have is a `TypeError: Failed to fetch` error, but that doesn't mean that there is a 301.

Another thing we can do is, in case there is a fetch error, do another request with `redirect:  "manual"`, check if the response type is `"opaqueredirect"` and, if true, go to the link using SSR as @SantosGuillamot proposed.

Would it make sense?

luisherranz:
> Another thing we can do is, in case there is a fetch error, do another request with redirect: "manual", check if the response type is "opaqueredirect" and, if true, go to the link using SSR as @SantosGuillamot proposed.
> Would it make sense?

Oh, that's an awesome solution! Yes, we can do that 🙂

luisherranz:
Funny thing (or not that funny 😅): Cypress does not allow to change the URL to a different origin in the middle of a test (even with proper CORS headers), so I cannot do the external redirection test for the client.

And guess what, the workaround is: to turn off `chromeWebSecurity` 😆😆😆

Check the video and let me know what you think:
https://www.loom.com/share/b8dfae3a78504ca380459e924af759c4

I'll implement David's idea to fix the CORS problem in the meantime 👍

luisherranz:
Ok, I have a plan. Let's continue with the https://github.com/frontity/frontity/pull/671 to use chrome web security again and I'll use the `--suites` option here to run those tests in a new GH job that runs without security.

I think it's going to be safer in the long run to have the web security activated. CORS problems are always tricky and Frontity is especially susceptible to them.

luisherranz:
Ok, the e2e external redirection test in the client is working now without depending on external websites.

I will work tomorrow on: 
1. Implementing this change and adding more tests: https://github.com/frontity/frontity/pull/601#issuecomment-763507876
2. A way to run the client external redirection tests without chrome web security.

cristianbote:
Finally managed to catch-up and get some more context 😅 

Regarding:
> > The problem is that we don't know if there is a redirection or not. All that we have is a `TypeError: Failed to fetch` error, but that doesn't mean that there is a 301.
> 
> Another thing we can do is, in case there is a fetch error, do another request with `redirect: "manual"`, check if the response type is `"opaqueredirect"` and, if true, go to the link using SSR as @SantosGuillamot proposed.
> 
> Would it make sense?

Since we are executing/calling these fetch in the redirection context, if the calls fails with `failed to fetch` wouldn't be safe to assume that there's a client-side issue, and a full SSR nav should happen regardless? 🤔  is it Frontity's concern to resolve the redirection only if the client side call is successful and then redirect?

luisherranz (approved):

luisherranz:
@darerodz this is finally ready for review.

I already reviewed Michal's code, so you only have to review my modifications.

I have made a couple of videos explaining my changes so it is simpler for you to understand them:
- https://www.loom.com/share/754c7fbd84bf45f18db07a4e89e2f018
- https://www.loom.com/share/ce9809ed83764e08bee9cd1a29cd7555

The diff between Michal's code and mine is here: https://github.com/frontity/frontity/pull/601/files/ee62da9659af9f26caedb6f9fd7c5080d9a5985d..ffc8aa856d18198620a54715f6148fed29ececa9

I did some small last minute refactoring after that, nothing serious. I just moved a try/catch from `actions.source.fetch()` to inside `fetchRedirection` and I refactored the way we clone the `state` that is passed to `actions.router.set(link, { state })` and ends up in `history.state`: https://github.com/frontity/frontity/pull/601/files/ffc8aa856d18198620a54715f6148fed29ececa9..9057baeb1174bfe62085b19e16a1d8b05c7e71bd

Apart from that, the e2e tests are failing but only because I already turned the `chromeWebSecurity` flag ON again. So once the https://github.com/frontity/frontity/pull/671 PR is finished and merged, merging `dev` here should solve it.

If you have questions let me know 🙂

packages/tiny-router/src/actions.ts
```diff
@@ -138,17 +190,49 @@ export const beforeSSR: TinyRouter["actions"]["router"]["beforeSSR"] = ({
...
+  if (data && isRedirection(data)) {
+    // If the redirection is external, just redirect to the full URL here.
+    if (data.isExternal) {
+      ctx.redirect(data.location);
+      return;
+    }
+
+    // Recover all the missing query params from the original URL. This is
+    // required because we remove the query params that start with `frontity_`.
+    const location = new URL(data.location, "https://dummy-domain.com");
```
DAreRodz, line 242: Same here. Maybe I'm wrong... :thinking:

DAreRodz (approved):
Really awesome work, @michalczaplinski & @luisherranz, congrats!! :tada:

I've just written down some notes and questions below, take a look.

luisherranz:
Thanks David! 

I will review your comments on Monday 🙂

I have also pushed another update because I realized that the unit tests for tiny-router were missing. But I will explain those changes on Monday as well and I'll ask you to review them.

e2e/plugins/index.ts
```diff
@@ -116,7 +145,33 @@ const tasks = {
     })();
   },
 
-  updateOption({ name, value }: UpdateOption) {
+  /**
+   * Start a server that proxies requests from WordPress, but with a different
```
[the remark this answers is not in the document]
DAreRodz: I just saw that there were a lot of comments to fix. :see_no_evil:

Thanks for fixing all of them!

DAreRodz (approved):
Approved again! :heavy_check_mark:

BTW, could you explain more in-deep why the `resolved` function is needed? It is because the type of derived props is not resolved properly?

luisherranz:
Awesome 🙂

> BTW, could you explain more in-deep why the resolved function is needed? It is because the type of derived props is not resolved properly?

Yes. It's kind of a trick used to force a type change to a variable in a "clean way" for the final user.

It is explained a bit in this video: https://youtu.be/wNsKJMSqtAk?t=354, although he uses the function for the `extends` and I am actually changing the returned type. 

But for a final Frontity user, it can be a clean way to define actions and derived state on the fly, which is not straightforward right now because the type of the definition is different than the type of when the variable is used.

e2e/plugins/index.ts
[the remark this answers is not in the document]
luisherranz: True!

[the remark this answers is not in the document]
luisherranz: Done.

e2e/project/frontity.settings.ts
```diff
@@ -61,7 +61,10 @@ const settings: Settings<
     name: "wp-source-errors",
     packages: [
       "e2e-wp-source-errors",
-      "@frontity/wp-source",
+      {
+        name: "@frontity/wp-source",
+        state: { source: { url: "https://test.frontity.org/" } },
```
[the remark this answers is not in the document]
luisherranz: I think it's because I changed all the `state.source.api` to `state.source.url` in the frontity settings of the e2e project, but if it's not needed I'll remove it!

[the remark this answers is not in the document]
luisherranz: Done!

e2e/wp-data/.htaccess
```diff
@@ -4,6 +4,9 @@ RewriteRule ^(.*) - [E=HTTP_AUTHORIZATION:%1]
 SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1
 
 # BEGIN WordPress
+# The directives (lines) between "BEGIN WordPress" and "END WordPress" are
+# dynamically generated, and should only be modified via WordPress filters.
+# Any changes to the directives between these markers will be overwritten.
```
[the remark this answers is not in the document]
luisherranz: That is from a Michal's commit (https://github.com/frontity/frontity/commit/8456e9fb004b40dca62f9e2999c422c867ce0e2f) but if I'm not mistaken that text is automatically added by WordPress.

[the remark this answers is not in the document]
luisherranz: The docs for `.htaccess` are here: https://wordpress.org/support/article/htaccess/

packages/connect/src/__tests__/connect.tests.jsx
```diff
@@ -267,6 +266,7 @@ describe("useConnect", () => {
   });
 
   it("should log a warning if the component is not connected", () => {
+    // eslint-disable-next-line
```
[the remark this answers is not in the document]
luisherranz: You're right.

Done!

packages/tiny-router/src/actions.ts
```diff
@@ -44,8 +44,28 @@ export const set: TinyRouter["actions"]["router"]["set"] = ({
...
+  // The redirections are stored in source.data just like any other data.
+  const data = state.source?.get(link);
+  if (data && isRedirection(data)) {
+    if (data.isExternal) {
+      window.replaceLocation(data.location);
+    } else {
+      // If the link is internal, we have to discard the domain.
+      const { pathname, hash, search } = new URL(
+        data.location,
+        "https://dummy-domain.com"
```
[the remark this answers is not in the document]
luisherranz: Yeah, I thought so at the beginning because the standard (https://tools.ietf.org/html/rfc2616#section-14.30) says that `Location` should contain absolute links, but people can still use relative links and they work fine in the browsers, so we need to make sure they work.

_Most of the redirections we used in WordPress for the e2e tests are actually relative links._

packages/tiny-router/src/actions.ts
luisherranz: Explained in https://github.com/frontity/frontity/pull/601#discussion_r563676453.

luisherranz:
This merge needs some fireworks.

fw

Issue #596, opened 2020-10-07
`actions.source.fetch` fails with links with query params for certain handlers [5pt]


DAreRodz:
**_Bug report_**

- I've updated both my project and my packages to the latest Frontity versions: https://docs.frontity.org/guides/keep-frontity-updated
- I've searched for existing issues.
- I've checked the documentation: https://docs.frontity.org
- I've checked the community forum: https://community.frontity.org

Let's say we have these two handlers, one for `movie` entities and another for `movie` attachments:
```ts
[
  {
    pattern: "/movie/:movieSlug/:slug",
    name: "movie attachment",
    func: () => { ... },
    priority: 10,
  },
  {
    pattern: "/movie/:slug",
    name: "movie",
    func: () => { ... },
    priority: 10,
  },
]
```

**Expected behavior**

When doing
```ts
actions.source.fetch("/movie/the-terminator/?query=string")
```

you may expect that the `movie` handler is the one that matches.

**Observed behavior**

It is not only that the `movie attachment` handler is matched but also the `getMatch()` call fails because the params that `getMatch()` returns are extracted from the pathname and not from the query, which [is being removed](https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/libraries/get-match.ts#L20-L22) before getting those params, thus making the link not to match anymore with the matched pattern from `movie attachment` and throwing the following error in this line:
 and the following error is thrown (in [this line](https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/libraries/get-match.ts#L24)):

```
TypeError: Cannot read property 'slice' of null
```
The data object for that link ends being:

```ts
{
  isFetching: true,
  isReady: false,
  link: "/movie/the-terminator/?query=string",
  page: 1,
  query: { query: "string" },
  route: "/movie/the-terminator/",
}
```

**Steps involved to reproduce the problem**

1. `npx frontity create`.
2. Add the handlers mentioned above to `libraries.source.handlers` in /packages/mars-theme/index.js
3. Add the link `"/movie/the-terminator/?query=string"` to the menu in /frontity.settings.js
4. `npx frontity dev` and click on that link.
4. In the console you will see `TypeError: Cannot read property 'slice' of null`.
5. The theme will just show the loading bar.

**Possible solution**

I'm not sure how to solve this, actually.

As a workaround, users just would have to add a trailing slash to their handler patterns.

luisherranz:
Thanks David. This is the test to reproduce the issue:

```js
[...]

PR #603, opened 2020-10-16
Catch error whill returning result in getMatch function

jeanettepraninPF:
**What**:

Adding a try/catch in the wp-source get-match.ts file, because the "extractParameters" function can cause the following error:
TypeError: Cannot read property 'slice' of null
      at extractParameters (webpack-internal:///./node_modules/@frontity/wp-source/src/libraries/get-match.ts:16:34)

**Why**:

Instead of returning a 404 Not Found error, the code is returning a 500 Internal Server error. Please see https://trashisfortossers.com/2015/07/day-1-upstate-new-york.html?spref=pi as an example.

**How**:

On my own personal fork and tested.

**Tasks**:

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Unrelated Tasks**

**Additional Comments**

### DIFF
--- a/packages/wp-source/src/libraries/get-match.ts
+++ b/packages/wp-source/src/libraries/get-match.ts
@@ -77,8 +77,9 @@ export const getMatch = <Patt extends Pattern>(
     })
     .find(({ regexp }) => regexp.test(link));
 
-  return result
-    ? {
+  if (result) {
+    try {
+      return {
         func: result.func,
         params: result.pattern.startsWith("RegExp:")
           ? result.regexp.exec(link).groups
@@ -86,5 +87,11 @@ export const getMatch = <Patt extends Pattern>(
         name: result.name,
         pattern: result.pattern,
       }
-    : null;
+
+    } catch {
+      return null;
+    }
+  } else {
+    return null;
+  }
 };


luisherranz:
Hey @jeanettepraninPF, many thanks for the PR!!

May I know which pattern is making the `extractParameters` function to fail? I guess the URL is `/2015/07/day-1-upstate-new-york.html?spref=pi`, right?

I'd love to understand more why this is failing and if returning a 404 is the correct behavior 🙂

jeanettepraninPF:
Hi @luisherranz! The reason why I think this should be returning a 404 instead of a 500 is because the URL is a valid Wordpress URL, although this specific article has been hidden so it is not available to the public.

SantosGuillamot:
Thanks for the PR @jeanettepraninPF :slightly_smiling_face: In case it helps, it seems it's the query `?spref=pi` the one causing the 500:

* This returns a 404 -> https://trashisfortossers.com/2015/07/day-1-upstate-new-york.html
* This returns an Internal Server Error -> https://trashisfortossers.com/2015/07/day-1-upstate-new-york.html?spref=pi

I guess both should return a 404 right?

luisherranz:
This seems to work just fine: https://mars.frontity.org/not-existent-page/?spref=pi

@jeanettepraninPF please add a list of your patterns here so we can take a look.

jeanettepraninPF:
@luisherranz 
For my common settings in Wordpress, we have the "Post name" selected --> "https://www.trashisfortossers.com/sample-post/"

We have these patterns as well:
pattern: "/articles/:post_name/:id",
pattern: "/videos/",
pattern: "/style/",
pattern: "/home-and-living/",
pattern: "/diy-and-guides/",
pattern: "/ceo-diaries/"

These are all initialized in actions --> theme --> init in the twentytwenty-theme/src/index.js
`libraries.source.handlers.push(articleHandler, videosHandler, styleHandler, homeAndLivingHandler, diyHandler, ceoDiariesHandler);`

luisherranz:
Thanks @jeanettepraninPF. I was able to debug this directly in https://www.trashisfortossers.com/ and I found that the pattern that is catching `/2015/07/day-1-upstate-new-york.html?spref=pi` is `"attachment from post by day"` which is `"/:year(\d+)/:month(\d+)/:day(\d+)?/:postSlug(.*\D.*)/:slug"` and it's doing so because it's using the query `?spref=pi` as the final `:slug` part. Then, it fails because we are removing the query in the `extractParameters` function:

https://github.com/frontity/frontity/blob/7e99b9b4a7d1bb8d3e02d3b7bce587bb8f400bee/packages/wp-source/src/libraries/get-match.ts#L19-L23

So this is finally a duplicate of https://github.com/frontity/frontity/issues/596 and once that is solved, it should return a 404.

Jeanette, would you be willing to help us solve that bug instead? I'll add more information about how to solve it in the issue.

Regarding this PR, once we fix the #596 bug it shouldn't be necessary to try/catch the return of `getMatch` because it should never fail, so I prefer to leave it as it is because if it fails for someone else in the future it means there's another bug, so I'm closing it now. Feel free to reopen it if you if there's something else we're not aware of  🙂

jeanettepraninPF:
@luisherranz Got it. Reading that bug, it seems like DAreRodz's solution should work. Thanks for your help! Is this something that may come out in an update soon?

luisherranz:
It's rated as `priority: medium` right now so it'd be awesome if you could help us out with it 🙂

PR #606, opened 2020-10-22
Normalize upper/lower case paths

### LINKED ISSUE #604: Frontity doesn't "recognize" the source of data if the route is NOT 100% lower case
**_Bug report_**

**Expected behavior**

If my request (Frontity) is **/book/ABC**, I expect to see the same result when I request **/book/abc**.

**Observed behavior**

A page not found is loaded. WordPress is loading the correct content, but the condition on line 83 of `node_modules/@frontity/wp-source/src/libraries/handlers/postType.ts file` makes this fails (=== condition where the source is **/book/ABC** and the populated link is **/book/abc**.

Try to access any existing page/post on WordPress with upper case letters on the request/URL.

**Info about your system**

```
## System:
 - OS: macOS 10.15.7
 - CPU: (12) x64 Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz
 - Memory: 33.72 GB / 64.00 GB
 - Shell: 5.7.1 - /bin/zsh
## Binaries:
 - Node: 12.18.0 - /usr/local/bin/node
 - npm: 6.14.4 - /usr/local/bin/npm
## Browsers:
 - Chrome: 86.0.4240.80
 - Edge: Not Found
 - Firefox: 81.0
 - Safari: 14.0
## npmPackages:
 - @frontity/core: ^1.9.0 => 1.9.0 
 - @frontity/html2react: ^1.4.0 => 1.4.0 
 - @frontity/mars-theme: ./packages/mars-theme => 1.4.3 
 - @frontity/tiny-router: ^1.2.1 => 1.2.1 
 - @frontity/wp-source: ^1.9.0 => 1.9.0 
 - frontity: ^1.12.0 => 1.12.0 
## npmGlobalPackages:
 - frontity: Not Found
 - npx: Not Found
```

**Possible solution**

I had a long discussion with a community member where I explain the situation, show possible solutions, etc. Here is the link: https://community.frontity.org/t/how-make-slug-case-insensitive/3138

I've provided 2 suggestions. The first one I don't think is correct (it seems more a workaround), but the second one has a chance to be the solution:
1) https://community.frontity.org/t/how-make-slug-case-insensitive/3138/11
2) https://community.frontity.org/t/how-make-slug-case-insensitive/3138/14

Thank you so much.

nandotess:
**What**:

Normalize upper/lower case paths.

**Why**:

If my request (Frontity) is /book/ABC, I expect to see the same result when I request /book/abc.

**How**:

Improving the route-utils/normalize function to force lower case path.

**Tasks**:

- [x] Code
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

- Community discussion: https://community.frontity.org/t/how-make-slug-case-insensitive/3138
- Issue: https://github.com/frontity/frontity/issues/604

### DIFF
--- a/.changeset/quick-flowers-pretend.md
+++ b/.changeset/quick-flowers-pretend.md
@@ -1,0 +1,5 @@
+---
+"@frontity/wp-source": patch
+---
+
+Make `libraries.source.normalize()` to lowercase the pathname of links. This fixes a bug when links with uppercase letters are visited.
--- a/packages/wp-source/src/libraries/__tests__/route-utils.test.ts
+++ b/packages/wp-source/src/libraries/__tests__/route-utils.test.ts
@@ -276,6 +276,10 @@ describe("route utils - normalize", () => {
     expect(normalize("/some/path")).toBe("/some/path/");
   });
 
+  test("from path (upper case)", () => {
+    expect(normalize("/SOME/PATH/")).toBe("/some/path/");
+  });
+
   test("from path and page", () => {
     expect(normalize("/some/path/page/2")).toBe("/some/path/page/2/");
   });
--- a/packages/wp-source/src/libraries/__tests__/route-utils.test.ts
+++ b/packages/wp-source/src/libraries/__tests__/route-utils.test.ts
@@ -288,6 +288,10 @@ describe("route utils - normalize", () => {
     expect(normalize("/some/path?k1=v1&k2=v2")).toBe("/some/path/?k1=v1&k2=v2");
   });
 
+  test("from path and query (upper case)", () => {
+    expect(normalize("/some/PATH?K1=V1&k2=v2")).toBe("/some/path/?K1=V1&k2=v2");
+  });
+
   test("from path, page and query", () => {
     expect(normalize("/some/path/page/2?k1=v1&k2=v2")).toBe(
       "/some/path/page/2/?k1=v1&k2=v2"
--- a/packages/wp-source/src/libraries/route-utils.ts
+++ b/packages/wp-source/src/libraries/route-utils.ts
@@ -146,7 +146,7 @@ const paramsToLink = ({
   const pathAndPage = page > 1 ? `${path}page/${page}/` : path;
   const queryString = objToQuery(query);
 
-  return `${pathAndPage}${queryString}${hash}`;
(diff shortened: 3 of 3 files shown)


SantosGuillamot:
@DAreRodz Could you take a look at this please? According to what @nandotess said in the issue: _"It seems to be related to other parts of the code that are failing..."_

DAreRodz:
OK, I saw the problem. We cannot lowercase the whole link because we would end up changing the query params too.

I guess the link `"/author/author-1/?s=findAuthor"` was transformed to `"/author/author-1/?s=findauthor"` in this tests, making the data object not to be found.

```text
@frontity/wp-source: FAIL src/libraries/handlers/__tests__/author.tests.ts
@frontity/wp-source:   ● author › is requested with a search param
@frontity/wp-source:     TypeError: Cannot read property 'isSearch' of undefined
@frontity/wp-source:       206 | 
@frontity/wp-source:       207 |     expect(
@frontity/wp-source:     > 208 |       store.state.source.data["/author/author-1/?s=findAuthor"].isSearch
@frontity/wp-source:           |                                                                 ^
@frontity/wp-source:       209 |     ).toBe(true);
```

So, we only have to lowercase the pathname of a link. We can do so in the `paramsToLink` function that `normalize` uses internally:

https://github.com/frontity/frontity/blob/ff9af973484cb532d1dde73470244fdec2e34bd2/packages/wp-source/src/libraries/route-utils.ts#L149

Just change that line by this one:

```ts
return `${pathAndPage.toLowerCase()}${queryString}${hash}`;
```

DAreRodz:
BTW, @nandotess could you add a new test case for a link with an uppercased query to check that it is not affected? Or simply modify the one you added. :slightly_smiling_face:

nandotess:
@SantosGuillamot @DAreRodz thank you for your feedback and instructions. A new commit was made with the fix and a new test. Let's check the results once they are done... 🙂

DAreRodz (approved):
The code is fine. Thank you for your contribution, @nandotess!

PR #609, opened 2020-10-29
Fix serve command takes too much to start

### LINKED ISSUE #605: `npx frontity serve` command takes too much time to start the server [8pt]
**_Bug report_**

**Expected behavior**

I expect that, after running `npx frontity serve`, the Server starts in ~1sec at most.

**Observed behavior**

Right now, if we run it locally it can take ~10 seconds and even more (20-30) if we deploy it to serverless platforms like Google Cloud Run. This is not a big deal for local development or for a common server, but for serverless platforms, where the container is stopped and it has to start again from time to time, it really affects the performance.

Run `npx frontity serve` in any Frontity project. It should be instant and it isn't right now.

**Possible solution**

It seems this is caused by [`commander.js`](https://www.npmjs.com/package/commander), the library we're using for our CLI, so we could just not use commander for the `serve` command.

If we bypass commander and we run the serve script directly, the server takes much less to start. To test this, you just have to create a `server.js` file in the root of your project and run `node server.js` in the console instead of `npx frontity serve`. This way we can see the difference.

_server.js_:
```
const serve = require("@frontity/core/dist/src/scripts/serve").default;

serve({
  isHttps: false,
  port: 8080,
});
```
----
It seems that this is causing other problems as well, like the one explained in this [community topic](https://community.frontity.org/t/error-did-you-forget-to-run-frontity-build/1426/39?u=santosguillamot), and it is also solved bypassing commander.

cristianbote:
**What**:

Closes #605

**Why**:

By loading dynamically only the needed script we make sure to not overload the current process with the side-effects from the other scripts.

**How**:

Inside the `@frontity/core` package I changed the way the files were exported.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript

**Unrelated Tasks**
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**
None

### DIFF
--- a/.changeset/cold-jars-give.md
+++ b/.changeset/cold-jars-give.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+The `serve` command was taking too much to spin the server. That was due to the build and dev scripts which are having side-effects by loading ts-node and other scripts.
--- a/packages/core/src/scripts/build.ts
+++ b/packages/core/src/scripts/build.ts
@@ -56,7 +56,7 @@ import { webpackAsync } from "./utils/webpack";
 /**
  * The options of the build command.
  */
-interface BuildOptions {
+export interface BuildOptions {
   /**
    * The Webpack mode used, either "development" or "production".
    *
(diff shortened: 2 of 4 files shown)


luisherranz (approved):
Awesome work @cristianbote! Thanks for the PR 🙂

---

It seems like Eslint is not working for external forks. @darerodz has made some changes to Eslint in the [Update to TypeScript 4.0](https://github.com/frontity/frontity/pull/449) PR which may fix this issue but I'm going to merge this now.

PR #610, opened 2020-10-29
Add the `state.source.url` setting to wp-source

### LINKED ISSUE #589: Create a global setting for the backend [8pt]
**Feature:** [Make the backend URL a global setting](https://community.frontity.org/t/make-the-backend-url-a-global-setting/2381).

[**Sprint Card**](https://github.com/orgs/frontity/projects/10#card-46619778)

michalczaplinski:
**What**:

Add the setting for `state.source.url` according to https://community.frontity.org/t/make-the-backend-url-a-global-setting/2381/17

**Why**:

We want to be able to set the canonical URL of the WordPress backend in a frontity application. It will also be needed by future features, most notably in [301 Redirections](https://community.frontity.org/t/301-redirects-stored-in-wordpress-database/3032/3)

**How**:

Adding the relevant settings and their defaults to the `wp-source`.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] TypeScript tests
- [x] Unit tests
- [x] Open an issue for this feature in the docs repo:  https://github.com/frontity/docs/issues/245
- [x] Update starter themes
- [x] Update other packages
- [x] End to end tests
- [x] Update community discussions: https://community.frontity.org/t/make-the-backend-url-a-global-setting/2381/20

**Unrelated Tasks**

**Additional Comments**


packages/types/src/derived.ts
```diff
@@ -1,9 +1,16 @@
+/* eslint-disable */
```
michalczaplinski, line 1: I've added it because otherwise eslint is forcing me to add JSDocs for the `state`, `actions` and `libraries` which in my opinion actually makes it harder to read what the type is.

packages/types/src/derived.ts
```diff
@@ -1,9 +1,26 @@
+/* eslint-disable jsdoc/require-jsdoc */
```
michalczaplinski, line 1: I've added this because otherwise eslint requires me to add JSDocs for `state`, `actions` & `libraries` which in my opinion makes it actually harder to understand this type.

packages/types/src/utils.ts
```diff
@@ -1,13 +1,18 @@
+/* eslint-disable jsdoc/require-jsdoc */
```
michalczaplinski, line 1: Same as https://github.com/frontity/frontity/pull/610/files#r514747732

packages/connect/src/index.d.ts
```diff
@@ -28,9 +28,13 @@ type Store = {
   actions?: Actions;
 };
 
+type Primitive = number | string | boolean | symbol | undefined | null;
```
michalczaplinski, line 31: You will notice that this type is duplicated in the [https://github.com/frontity/frontity/pull/610/files#diff-936fe7efe4a257e580cf24c0a2084d2f1428e37b542edac7b6db3f932c3514d6R57](packages/types/src/utils.ts).

This should be cleaned in the PR when we fork react-easy-state.

packages/connect/src/index.d.ts
```diff
@@ -28,9 +28,21 @@ type Store = {
...
+export type Serializable =
+  | Primitive
+  | (Primitive | Record<string, Primitive>)[]
+  | Record<string, Primitive>;
+
 type ResolveState<State> = {
   [P in keyof State]: State[P] extends (state: object) => any
     ? ReturnType<State[P]>
+    : State[P] extends Serializable | ((state: object) => any)
+    ? Exclude<State[P], Derived<any, any>>
```
luisherranz, line 45: Good trick @michalczaplinski, a very smart way to solve this problem! 🙂

cc: @DAreRodz

luisherranz (approved):
Thanks Michal. Looks good to me 👍

Things I have done:

- I have moved `state.source.url` to the types of `@frontity/source` because it should be present on all the source packages.
- I have moved `state.source.api` and `state.source.isWpCom` to `state.wpSource` with backwards compatibility because they only make sense for `@frontity/wp-source`. Well, `state.source.api` makes sense for other packages, but they belong to each package namespace:
  - `state.wpSource.api`: `"https://backend.com/wp-json"`
  - `state.wpGraphqlSource.api`: `"https://backend.com/graphql"`
- I have added tests for:
  - When `state.source.url` contains a folder.
  - When `state.wpSource.url` is `?rest_route=`.
  - When `state.source.api` is set manually.
  - When `state.source.url` is from wordpress.com.
- I have added an error when `state.source.url` is accessed but neither `state.frontity.url` or `state.source.url` are set in the `frontity.settings.js` file.
- I have moved all the handler tests to use `state.source.url` instead of `state.source.api`.
- I have updated the typescript tests of `@frontity/source`.
- I have updated the `frontity.settings.js` of the `examples` folder.
- I have updated the `frontity.settings.js` generated in `npx frontity create`.
- I have updated the `frontity.settings.js` of the end-to-end project.

packages/connect/src/index.d.ts
michalczaplinski: Thanks @luisherranz 😊

PR #611, opened 2020-11-05
Add a test for special characters in URL

### LINKED ISSUE #432: Special character in permalink cause internal server error
**_Bug report_**

**Expected behavior**

If I visit an URL with special character in the permalink, it should open the post.

**Observed behavior**

Internal Server Error:
```
TypeError: Cannot read property 'split' of undefined
      at __webpack_exports__.default (webpack-internal:///./node_modules/@frontity/wp-source/src/libraries/handlers/utils/capitalize.ts:2:73)
      at Object.eval (webpack-internal:///./node_modules/@frontity/wp-source/src/libraries/handlers/postType.ts:10:205)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
      at eval (webpack-internal:///./node_modules/@frontity/wp-source/src/actions.ts:18:101)
```

`npx frontity create` -> Create a WP post, with the title: `test♯post`, Publish it, than visit it through Frontity Frontend (e.g.: http://localhost:3000/test%E2%99%AFpost or http://localhost:3000/test♯post). If navigated to the page, than press enter on the address bar. 
Example: https://szinhaz.online/test%E2%99%AFpost2

**Info about your system**

## System:
 - OS: Linux 5.3 Linux Mint 19.3 (Tricia)
 - CPU: (8) x64 Intel(R) Core(TM) i7-4810MQ CPU @ 2.80GHz
 - Memory: 7.36 GB / 15.54 GB
 - Shell: 4.4.20 - /bin/bash
## Binaries:
 - Node: 12.16.3 - /usr/bin/node
 - npm: 6.14.5 - ~/.npm_modules/bin/npm
## Browsers:
 - Chrome: 81.0.4044.138
 - Firefox: 75.0
## npmPackages:
 - @frontity/core: ^1.5.3 => 1.5.3 
 - @frontity/html2react: ^1.3.2 => 1.3.2 
 - @frontity/mars-theme: ./packages/mars-theme => 1.4.1 
 - @frontity/tiny-router: ^1.2.0 => 1.2.0 
 - @frontity/wp-source: ^1.7.0 => 1.7.0 
 - frontity: ^1.6.0 => 1.6.0 
## npmGlobalPackages:
 - frontity: Not Found
 - npx: Not Found

**Possible solution**

michalczaplinski:
**What**:

Add an e2e test to check that special characters work fine in permalinks. This was originally reported in #432 and fixed in https://github.com/frontity/frontity/pull/606/

**Why**:
To ensure that issues like https://github.com/frontity/frontity/issues/432 don't appear in the future.

**Tasks**:

- [x] Code

- [x] End to end tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


e2e/integration/wordpress-01/special-characters-in-url.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+describe("Special characters in the URL", () => {
+  before(() => {
+    task("loadDatabase", {
+      path: "./wp-data/post-with-special-characters/dump.sql",
+    });
+  });
+
+  it("Should load the post", () => {
+    cy.visit(
+      "http://localhost:3001/test%e2%99%afpost%f0%9f%98%8a/?frontity_name=wp-basic-tests"
```
DAreRodz: Shouldn't we use here the special characters in the URL? This is the encoded version and it is already in lowercase. Wouldn't it work in that case even without the bugfix? :thinking:

DAreRodz (changes requested):
Great idea to add an e2e test for this. :+1: 

I have a question though, see below.

e2e/integration/wordpress-01/special-characters-in-url.spec.ts
michalczaplinski: You're absolutely right, I think pushing this late in the evening was not the best idea.

Just one thing to note, my very first idea was to do just:

```
cy.visit("http://localhost:3001/test♯post😊/?frontity_name=wp-basic-tests");

cy.location("href").should(
      "eq",
      "http://localhost:3001/test%e2%99%afpost%f0%9f%98%8a/"
    );
```

However, the problem is that cypress fails with 
Screen Shot 2020-11-05 at 7 17 10 PM

The problem is that cypress does not encode the URL for you in the same way that the browser does when you just paste something in the search bar! 😅 More details: https://github.com/cypress-io/cypress/issues/5274#issuecomment-537659134

So, I've created the test case which does the navigation on the client side instead (works fine as expected).

michalczaplinski:
> ## ESLint Summary [View Full Report](https://github.com/frontity/frontity/pull/611/checks?check_run_id=1361393059)
> > Annotations are provided inline on the [Files Changed](https://github.com/frontity/frontity/pull/611/files) tab. You can also see all annotations that were generated on the [annotations page](https://github.com/frontity/frontity/pull/611/checks?check_run_id=1361393059).
> 
> Type	Occurrences	Fixable
> **Errors**	0	0
> **Warnings**	1	0
> **Ignored**	0	N/A
> * **Result:**       ✅ success
> * **Annotations:** [1 total](https://github.com/frontity/frontity/pull/611/checks?check_run_id=1361393059)
> 
> ## [warning] [jest/expect-expect](https://github.com/jest-community/eslint-plugin-jest/blob/v24.1.0/docs/rules/expect-expect.md)
> > Enforce assertion to be made in a test body
> 
> * [e2e/integration/wordpress-01/special-characters-in-url.spec.ts](https://github.com/frontity/frontity/blob/493fb5ac8d90920b3251dca18199c69980feaf44/e2e/integration/wordpress-01/special-characters-in-url.spec.ts#L12-L12) Line 12 - Test has no assertions
> 
> Report generated by [eslint-plus-action](https://github.com/bradennapier/eslint-plus-action)

I don't understand why this is happening, as I don't see any issues locally. I've added the override in the `.eslintrc` but it seems to do nothing.

DAreRodz (approved):
I just fixed the issue with ESLint.

Everything OK :ok_hand:

PR #615, opened 2020-11-06
Fix Yoast backward compatibility

luisherranz:
**What**:

If we release the data type guard of `isTerm` as it is, it won't work with term custom handlers.

**Why**:

There may be packages that start using `isTerm(data)` to check if a data object is a term or not, like Yoast. Those packages need to be compatible with custom handlers that may have not added `isTerm: true` yet.

**How**:

I added the `isTaxonomy === true` check to the `isTerm` and `isTaxonomy` type guards.

**Tasks**:

- [x] Code
- [x] TypeScript tests
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


luisherranz:
@DAreRodz what do you think? Is this safe or do you see any problems?

DAreRodz:
Oh, nice catch @luisherranz, it's completely safe.

BTW, I added tests for `entity` type guards but not for the `data` ones because they seemed to be pretty straightforward, but maybe in this case we should. Do you think it's necessary? I can add them if you want.

luisherranz:
> Do you think it's necessary? I can add them if you want.

Good idea. Yes please 🙂

DAreRodz:
@luisherranz I've added tests for the data type guards, including tests to check if `isTerm` and `isTaxonomy` work the same with old taxonomy data objects.

Take a look, and if you think everything is OK I'll approve and merge the PR. :blush:

DAreRodz (approved):
Everything OK.

I added tests for all the data guards if you want to take a look.

luisherranz:
Wow. Impressive tests David. Thanks for that! 👍 😄

PR #625, opened 2020-11-20
Fix/internal links multiple sites

nicholasio:
**What**:

Bug fix for https://community.frontity.org/t/client-routing-for-internal-content-links/887/10

**Why**:

The link component was not taking into account the `match` property when replacing internal links. This would lead to issues if multiple sites are using the same source url.

**How**:

Testing the link against the `match` property if it does not match do not replace it.

**Tasks**:

- [x] Code
- [x] TSDocs
- [X] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

Might be a good idea to add some e2e tests for this.


SantosGuillamot:
I have tested it in [the frontity.org repo](https://github.com/frontity/frontity.org) but I wasn't able to make it work. It seems that the `match` property is undefined. Not sure if we can use [`useConnect` to get the match setting](https://github.com/frontity/frontity/blob/b288143be0ad332c0bbd006e1da344c142635aee/packages/components/link/index.tsx#L40). Maybe we need to use [the `getSettings` function](https://github.com/frontity/frontity/blob/dev/packages/file-settings/src/getSettings.ts) from `@frontity/file-settings` package. It seems this is something we are doing in `@frontity/core` package- > [Example](https://github.com/frontity/frontity/blob/dev/packages/core/src/server/index.tsx#L10).

@michalczaplinski Could you confirm which is the correct approach here?

nicholasio:
> I have tested it in [the frontity.org repo](https://github.com/frontity/frontity.org) but I wasn't able to make it work. It seems that the `match` property is undefined. Not sure if we can use [`useConnect` to get the match setting](https://github.com/frontity/frontity/blob/b288143be0ad332c0bbd006e1da344c142635aee/packages/components/link/index.tsx#L40). Maybe we need to use [the `getSettings` function](https://github.com/frontity/frontity/blob/dev/packages/file-settings/src/getSettings.ts) from `@frontity/file-settings` package. It seems this is something we are doing in `@frontity/core` package- > [Example](https://github.com/frontity/frontity/blob/dev/packages/core/src/server/index.tsx#L10).
> 
> @michalczaplinski Could you confirm which is the correct approach here?

We could simply add the `match` property to the frontity state in https://github.com/frontity/frontity/blob/dev/packages/core/src/server/utils/initial-state.ts#L52, however not sure if it would make sense to expose this optional prop to any package.

I don't think we can simply run `getSettings` on the client-side, from what I can tell it is s server-only function as it reads the settings file

luisherranz:
> We could simply add the match property to the frontity state in https://github.com/frontity/frontity/blob/dev/packages/core/src/server/utils/initial-state.ts#L52

Yep, if you finally want to expose `match` that would be the way to go 👍

nicholasio:
> > We could simply add the match property to the frontity state in https://github.com/frontity/frontity/blob/dev/packages/core/src/server/utils/initial-state.ts#L52
> 
> Yep, if you finally want to expose `match` that would be the way to go 👍

Done!

luisherranz:
@nicholasio is this ready for review? 🙂

nicholasio:
> @nicholasio is this ready for review? 🙂

Yes it is!

luisherranz:
Awesome 🙂

luisherranz (approved):
Looks good to me!

The only thing to notice, maybe in the docs, is that Frontity works this way:

- It tries to match the URL with the `match` of the sites.
- If no site matches, it returns the first site without a `match` field.

That means that the settings declared on Frontity can be:

- Site 1: `match: "/blog"`.
- Site 2: No match (default).

So if the Site 2 `Link` component gets a link that contains `/blog`, like https://domain.com/blog/some-post it will not find a `match` and it will resolve that it is an internal link, which is wrong.

The Frontity core can know about that because it knows about all the matches of all the sites, but a single site can't.

The way to solve this problem is to instruct people with this "folder" configuration to use `match` in both sites:

- Site 1: `match: "\/blog"`.
- Site 2: `match: "(?!\/blog)"`

_The regexps are not accurate, it is just a fast example._

By the way, as always, thanks for the fix @nicholasio!! 😄 👏 

EDIT: I have added an issue in the docs repo to explore that: https://github.com/frontity/docs/issues/258

PR #616, opened 2020-11-09
Avoid a couple of rerenders

luisherranz:
**What**:

Fix a couple of unnecessary rerenders when we transition from one data object to another.

I also switched from `instance of Array` to `Array.isArray` in a couple of places because it's more reliable.

**Why**:

We should try to minimize the number of rerenders to improve the general performance.

**How**:

I made a video explaining the change:

https://www.loom.com/share/eddc33966dcd410b8e931baf1f7d4945

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] End to end tests
- [x] Add a changeset (with link to its [Feature Discussion](https://community.frontity.org/c/33) if it exists)
- [x] Unit tests

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages
- [ ] Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases)
- [ ] Update community discussions

PR #626, opened 2020-11-23
Batch the `actions.source.fetch()` updates of the data object

luisherranz:
**What**:

Use `batch` in the `Object.assign` that modifies the data object to set `isReady: true` in `actions.source.fetch()`.

**Why**:

Last week I introduced a couple of changes in the "Avoid a couple of rerenders" PR (https://github.com/frontity/frontity/pull/616). One of those changes was to start using `Object.assign` instead of the object spread.

- Object spread:

  ```js
  source.data[link] = {
    ...source.data[link],
    ...(isHome && { isHome: true }),
    isFetching: false,
    isReady: true,
  };
  ```

- `Object.assign`:

  ```js
  Object.assign(source.data[link], {
    ...(isHome && { isHome: true }),
    isFetching: false,
    isReady: true,
  });
  ```

That solved a rerender because the object reference is preserved, but `Object.assign` adds the props one by one, which triggers the invalidations before all the object has been mutated, resulting in an "intermediate" state.

**How**:

I have solved it using `react-easy-state`'s `batch` utility:

```js
batch(() =>
  Object.assign(source.data[link], {
    ...(isHome && { isHome: true }),
    isFetching: false,
    isReady: true,
  })
);
```

The problem with that approach is that it only acts as batcher of React components, not `observe()` functions.

This is not ideal, I think `batch` should also batch the `observe` functions, like [MobX transactions](https://mobx.js.org/api.html#transaction). So I am not 100% sure that this will work to solve the problem that I introduced with the change to `Object.assign` in the first place.

@orballo could you please check if this change solves your problem?

- The lines you have to modify to test this are: https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/actions.ts#L99-L104
- You can use this code: https://github.com/frontity/frontity/commit/1d2579f71d9076026158b691b84081549d1b3c30

You can make the modification in `node_modules` directly.

If that doesn't work, we would need to go back to the object spread.

**Tasks**:

- [x] Code
- [x] Wait for @orballo confirmation that this approach solves his problem

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


luisherranz:
@orballo were you able to test if this fixes your issue?

orballo:
@luisherranz sorry I didn't see this issue. I'll try that out, but I think it won't fix my problem completely, because I was using custom handlers and therefore I'd need to add this to each one of them.

**EDIT:** It seems to solve the issue. Now when I `console.log` `data` in the Yoast hook it will run only once and with all the values set.

luisherranz:
Perfect, thanks @orballo! 

This is ready for review then 🙂

SantosGuillamot:
@luisherranz This Pull Request is not a draft anymore right?

luisherranz:
> This Pull Request is not a draft anymore right?

Ups, you're right, sorry. Fixed now.

michalczaplinski (approved):
great 👍

PR #627, opened 2020-11-23
Fix the client/server link mismatch

### LINKED ISSUE #623: state.router.link mismatch when the URL is not the same in the server as in the client
**_Bug report_**

**Expected behavior**

`state.router.link` should have the URL of the client.

**Observed behavior**

`state.router.link` has the URL that was sent to the server.

Add this to the Frontity server:

```js
// Change a query in the server.
if (ctx.URL.searchParams.get("utm")) {
  ctx.URL.searchParams.set("utm", "SERVER");
}
```

And visit http://localhost:3000/?utm=client. `state.router.link` is http://localhost:3000/?utm=SERVER and the URL is also replaced in the browser bar.

**Possible solution**

I think the way to go would be to:

- Solve the problem of the canonical links not pointing to the same data object.
- Restore the `state.router.link` to the client URL in the hydration.

I have explained it in more detail here:

- Video: https://www.loom.com/share/707af27e0c0f4d7db7d4551c98d5f7f8
- Drawing: https://excalidraw.com/#json=5474711555276800,KVqzcQvObG8FHpK0kMujrg

A quick fix would be to remove this line to now replace the client URL with the `state.router.link` value that was sent to the server, but that would break the "Back" button of the browser:

https://github.com/frontity/frontity/blob/dev/packages/tiny-router/src/actions.ts#L90

- You visit `/?utm=client`
- The server receives `/?utm=SERVER`
  - `state.router.link` is `/?utm=SERVER`
  - The data object is for `/?utm=SERVER`
- The client still has `/?utm=client` in the browser bar, but `state.router.link` is `/?utm=SERVER`
- You go forward to some link
- You go back with the browser button to `/?utm=client`. There is no data object for that URL.

We could alleviate the problem by using the autoFetch in the `"pop"` method, here: https://github.com/frontity/frontity/blob/dev/packages/tiny-router/src/actions.ts#L50-L61. That way, at least the page (`/?utm=client`) will be fetched again.

Also, we would need to remove `state.router.link` from the `updateState` action that is going to be introduced with the infinite scroll hooks: https://github.com/frontity/frontity/blob/infinite-scroll-beta/packages/tiny-router/src/actions.ts#L96

DAreRodz:
**What**:

- Change the `init` action of `tiny-router` so it doesn't replace the current URL with `state.router.link`.
- Change the `set` action of `tiny-router` so it calls `actions.source.fetch()` when going back if necessary. 
- Closes #623 

**Why**:

To avoid an issue when the Frontity server returns the state with a `state.router.link` that doesn't match the current URL. 

**How**:

Following the _quick fix_ proposal explained here: https://github.com/frontity/frontity/issues/623 

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] End to end tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


DAreRodz:
Oops, I forgot about the unit test. @michalczaplinski, do not review this yet. :sweat_smile:

michalczaplinski:
Hey @DAreRodz just wanted to check if we're on the same page about the solution :) 

Luis has suggested a solution this which involves modifying `wp-source` so that multiple links in `state.source.data` can point to the same data objects like explained in: https://www.loom.com/share/707af27e0c0f4d7db7d4551c98d5f7f8 (around 26:30) and in [the excalidraw](https://excalidraw.com/#json=5474711555276800,KVqzcQvObG8FHpK0kMujrg)


Are you going to implement the "quick" solution instead for now?

DAreRodz:
> Are you going to implement the "quick" solution instead for now?

Yup, this PR was to implement the "quick" fix (not so quick in the end)... :sweat_smile:

PR #633, opened 2020-11-28
chore: update lint-staged config

jamesgeorge007:
**What**:

Remove redundant instruction from `lint-staged` config.

**Why**:

 Since `v10` modified files are automatically added to the `git commit index`.

**How**:

Remove `git add .` from config.

**Tasks**:

- [x] Code

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

N/A

### DIFF
--- a/lint-staged.js
+++ b/lint-staged.js
@@ -18,10 +18,8 @@ const lintStaged = require("lint-staged");
     await lintStaged({
       allowEmpty: true,
       config: {
-        "*.{js,jsx,ts,tsx}": [
+        "*.{js,jsx,ts,tsx}":
           'replace-in-files --regex="\\/\\*\\*\\s*\n" --replacement="\n/**\n"',
-          "git add",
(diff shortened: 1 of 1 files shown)


DAreRodz (approved):
Good catch, @jamesgeorge007.

In the migration guide for `lint-staged` v10, they say:

> From v10.0.0 onwards any new modifications to originally staged files will be automatically added to the commit. If your task previously contained a **git add** step, please remove this. The automatic behaviour ensures there are less race-conditions, since trying to run multiple git operations at the same time usually results in an error.

It looks that it's safer to remove the `"git add"` commands than leaving them. I also tested it locally and works fine, so I'm going to approve this. :ok_hand:

PR #634, opened 2020-12-01
Fix link mismatch

### LINKED ISSUE #623: state.router.link mismatch when the URL is not the same in the server as in the client

DAreRodz:
**What**:

Closes #623 

**Why**:

To avoid an issue when the Frontity server returns the state with a `state.router.link` that doesn't match the current URL. 

**How**:

Changing the `actions.router.init()` function in `@frontity/tiny-router` so
- it replaces `state.router.link` with the link in the browser if a different one has come from the server.
- it makes the new value of `state.router.link` to point to the correct data object in `state.source.data`.

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] End to end tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


michalczaplinski (commented):
OK, I've reviewed and tested it and everything looks perfectly fine to me which makes me think that maybe I've missed something. That's a comment on my reviewing skill and not your coding, I think you did an awesome job @DAreRodz :) 

@luisherranz Could you also give it a quick look before I approve it?

luisherranz:
> OK, I've reviewed and tested it and everything looks perfectly fine to me which makes me think that maybe I've missed something. That's a comment on my reviewing skill and not your coding, I think you did an awesome job @DAreRodz :)
> 
> @luisherranz Could you also give it a quick look before I approve it?

Hey Michal, no, not really. I am sure the review is fine. And if it is not, we will find out and fix it later.

**None of us make bulletproof reviews and that is absolutely fine**. It is part of the job. Things are going to slip off sometimes and we will discover them later, file bugs, and fix them. There is nothing wrong with that process. The only thing we can do is to try to do our best at the moment, and learn as much as we can when something slips off to improve for the next time. And that's it. The rest is not up to us.

michalczaplinski:
> > OK, I've reviewed and tested it and everything looks perfectly fine to me which makes me think that maybe I've missed something. That's a comment on my reviewing skill and not your coding, I think you did an awesome job @DAreRodz :)
> > @luisherranz Could you also give it a quick look before I approve it?
> 
> Hey Michal, no, not really. I am sure the review is fine. And if it is not, we will find out and fix it later.
> 
> **None of us make bulletproof reviews and that is absolutely fine**. It is part of the job. Things are going to slip off sometimes and we will discover them later, file bugs, and fix them. There is nothing wrong with that process. The only thing we can do is to try to do our best at the moment, and learn as much as we can when something slips off to improve for the next time. And that's it. The rest is not up to us.

Gotcha, I think I wasn't confident that I haven't missed something, like I said. But you're right that we have to just "trust the process" :)

michalczaplinski (approved):
👍

PR #638, opened 2020-12-04
Add Selenium tests for older browsers

### LINKED ISSUE #630: BrowserStack - Selenium tests [21pt]
We are going to use BrowserStack and Selenium to run our tests in older OS and browsers, to ensure that Frontity is reliable in all the scenarios. Not all the current tests will make sense in old browsers, but we have to ensure that Frontity apps don't break.

Branch: https://github.com/frontity/frontity/tree/selenium-tests

SantosGuillamot:
**What**:

Add Selenium tests to the framework.

**Why**:

To ensure the framework works fine in old browsers.

**How**:

We are going to use BrowserStack and Selenium for this.

**Tasks**:

- [x] Code
- [x] End to end tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


SantosGuillamot:
A quick update on the status of this Pull Request. I've been working on it and I already created the different configuration files that were needed and I migrated the tests, dividing them in `module` tests and `ssr` tests.

These are, more or less, the tasks that are left:
* Review in depth the tests, clean them a bit and ensure there are no ESLint errors.
* Analyze better how to run tests in parallel. I tried it but the tests were failing randomly. I am not sure if it's because I was working locally.
* Create different builds per run. Right now I was using just one build for the tests, but we should use a different one each time we run the workflow.
* Review the GitHub action to ensure it works properly.
* Review if it's possible to use localhost in iPhones. This is a [BrowserStack limitation](https://www.browserstack.com/question/759).
* Once the AMP package is ready, I would like to review the tests.

I made a video explaining it with more detail:
https://www.loom.com/share/cfaf7617f97e4b0fbfab62b5b8cdfa73

luisherranz:
Awesome!!! 😄👏

Can't we use the http://bs-local.com URL for iPhone? Maybe we have to run it in a different batch but...

SantosGuillamot:
>Can't we use the http://bs-local.com URL for iPhone? Maybe we have to run it in a different batch but...

Yes, I think so. That's what I wanted to test. I think we won't need a different batch. My idea was to use the [beforeSession function](https://webdriver.io/docs/options/#beforesession) and check for the capabilities before each session. Something like this should work:

```js
beforeSession: (config, capabilities, specs) => {
    capabilities.browserName === "iPhone"
      ? (global.baseUrl = "http://bs-local.com:3000")
      : (global.baseUrl = "http://localhost:3000");
  },
```

And then we can create the urls with that variable.

luisherranz:
Sounds good! 🙂

SantosGuillamot:
I've been able to advanced in this Pull Request. This is what I did since the latest update:
* I moved the `http://localhost:3000` to a global variable and change it to `http://bs-local.com:3000` for iPhones. This way, all the capabilities work fine.
* I fixed all the Eslint errors adding specific rules for the Selenium tests.
* I added a [`localIdentifier` in BrowserStack capabilities](https://www.browserstack.com/docs/app-automate/appium/test-on-internal-network/manage-multiple-connections) to be able to run BrowserStack locally in different devices. For example, running test locally while GitHub is running them.
* I added `BROWSERSTACK_BUILD_NAME` env variable to differentiate the builds. Right now, as it is configured, this is how it would work:
  * If you run the tests locally, it will trigger a build with the name `Selenium-module/proxy/ssr-localhost-${Date}`, and if it's run in GitHub it will trigger a build with the name `Selenium-module/proxy/ssr-github-${ github.sha }`.
* I excluded the `smart-adserver.spec.js` until they fix them or we decide to delete them.
* I fixed the Google Analytics spec to adapt to the new library.

What I still have in mind:
* Take a look at the cypress tests I think I've broken after modifying some packages in the e2e project.
* Review the specs that may be lacking, like the AMP package.
* Review the tests in old browsers because I was having some issues.
* Decide when are we going to run these tests.

I am not sure if we will be able to run tests in parallel for the same build because working locally while the GitHub action was running, there were 4 sessions sometimes. So if we increase the parallel tests in the same build, it could break.

Just so you know, running all the specs for all the browsers with module, took 3 hours. We will have to run also the specs for proxy and ssr, although there are less combinations.

SantosGuillamot:
A new recap of what is left in this Pull Request:
* Add the Selenium specs for the Infinite Scroll.
* I wasn't able yet to prevent the duplicated sessions in BrowserStack.
* Once the[ BrowserStack Cypress Pull Request](https://github.com/frontity/frontity/pull/749) is merged, I'll review everything to make them work together.
* Review some old browser tests and open bugs if necessary.
* Add a callback to n8n when the tests fail.
* Clean the names of the builds and the tests to adapt them to the Cypress ones so they are consistent.

SantosGuillamot:
All the tests for module browsers and ssr browsers are passing:
Screen Shot 2021-03-17 at 16 38 11

Just so you know, they were running both suites at the same time and it took:
* Module suite - 1h 55m
* SSR suite - 33m

I still have to find out how to prevent duplicated session, but if we are able to do that, I think we could run two parallel tests in the module suite to reduce the time.

Apart from that, we still have to add the Proxy suite, which is already created but I didn't include because there is an issue with the browsers where `async` is not supported. I will take a deeper look and probably open an issue tomorrow.

SantosGuillamot:
@luisherranz FYI: According to [GitHub docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run), we could use `workflow_run` to trigger the Selenium BrowserStack workflows after the Cypress ones have finished to prevent too many parallel tests. I wasn't able to test it because, as they say: _"This event will only trigger a workflow run if the workflow file is on the default branch."_. We can check if it works once we merge this Pull Request and if not, I can merge everything into the same workflows and wait on the steps or jobs.

luisherranz:
Sounds good 🙂👍

SantosGuillamot:
FYI: I think I will be able to finish this Pull Request next week. What I still want to do:
* Fix the [`google-analytics` Cypress spec](https://github.com/frontity/frontity/blob/dev/e2e/integration/frontity-01/google-analytics.spec.js). We made some changes to the package to be able to test it in old browsers and the Cypress spy stopped working, so we are going to get rid of it and use the new implementation.
* Fix all the Proxy Selenium tests that aren't working. I have just triggered a suit with all the tests running in all the proxy browsers. I will review them next week to ensure everything works as expected.
* Record a video explaining all the changes included in this Pull Request.

SantosGuillamot:
@luisherranz I made the following video explaining the different changes made in the Pull Request. I wasn't sure how much detail to provide because there are a lot of specs and changes. If you feel you need a more detailed video or jump into a call just let me know. I've also realized while recording it that we may need to add some explanations in the [e2e readme](https://github.com/frontity/frontity/tree/dev/e2e). I'm gonna start working on that but I think you can already start reviewing it.

https://www.loom.com/share/b7259e452f0845adbf15ac1af8516acf

luisherranz (approved):
Everything looks perfect to me, Mario.

Impressive work. Congratulations 🙂

SantosGuillamot:
Thanks Luis :slightly_smiling_face:

PR #639, opened 2020-12-04
Forbidding local npm installs

luisherranz:
**What**:

Forbid running local `npm install`, inside of the packages.

**Why**:

Because it is not straightforward that people should only use `npm install` in the root and the `e2e` folder, and not in the packages or projects.

**How**:

I added a `preinstall` script to all our packages that echos a message and stops the install.

I add to also use lerna's `--ignore-scripts` option and run `lerna run build` after the install to make lerna not to run the `preinstall` commands.

**Tasks**:

- [x] Package.json scripts

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

I am not 100% sure about the approach. Let's just try and see. No problem as this is only for the Frontity Core team 🙂


DAreRodz (approved):
I think is OK so I'll approve this. :+1: 

BTW, did you notice that `lerna run build` is building also the example packages and the e2e project? For me, that takes a quite long time (not so long, anyway).

Maybe we could filter the packages that run the `build` script? The bad news is that there's no way to [filter packages by path](https://github.com/lerna/lerna/issues/1840#issuecomment-469728168) using `lerna`. We would have to look for a different solution.

luisherranz:
You are absolutely right, good catch. I have changed it to `lerna run prepublish`, which was what lerna was doing anyway.

DAreRodz:
Using `prepublish` is definitely better. :blush: Feel free to merge this, @luisherranz.

PR #650, opened 2020-12-20
Fix attachment type

luisherranz:
**What**:

I have fixed the type of `AttachmentEntity` (`state.source.attachment`) because it didn't contain the types of the post type, like `title`.

**Why**:

Because when you use this, TypeScript complains that `title` doesn't exit in `media`.

```ts
const media = state.source.attachment[id];
media.title.rendered;
```

**How**:

Extending `AttachmentEntity` from `PostTypeEntity` instead of just `Entity`.

**Tasks**:

- [x] TypeScript

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

The test was not failing because `AttachmentEntity` extended from `Entity`, which is:

```ts
export interface Entity {
  [k: string]: unknown;
}
```

So the extra things that were missing from `PostTypeEntity`, like `title.rendered` were just fine.

Right now, I don't know how to make a test where that fails. @darerodz, any idea?

---

I have also added an unrelated type test for `useConnect` because I was taking a look and I noticed we forgot to add one.


DAreRodz (approved):
Thanks for the fix! :raised_hands:  BTW, there were other properties missing for that same reason - apart from `title`, I mean.

> Right now, I don't know how to make a test where that fails. @DAreRodz, any idea?

Maybe we could test that by using [`expectType()`](https://github.com/frontity/frontity/blob/19165d366587ba364fa95e85b4afa6f67cb0813c/packages%2Ftypes%2Fhelpers.ts#L7-L9) for every property we expect to be typed. Would it make sense?

luisherranz:
> Maybe we could test that by using expectType() for every property we expect to be typed. Would it make sense?

I'll try that. Thanks!

luisherranz:
I have refactored the entity tests a little bit following what you did for `data`. Although I didn't add tests for every single property. It felt a bit overkill to me.

Merging now!

PR #652, opened 2020-12-21
Release the first stable version of the infinite scroll hooks

### LINKED ISSUE #628: First stable version of the Infinite Scroll hooks [13pt]
Feature: [Infinite Scroll hooks](https://community.frontity.org/t/infinite-scroll-hooks/2055).

Tasks left:
- Add e2e tests.
- Add TSDocs.

DAreRodz:
**What**:

This PR wraps all the work needed to release the first stable version of the infinite scroll hooks. Those hooks are exposed by the `@frontity/hooks` package.

Closes #628 

**Why**:

To release the "official" way to implement that behavior in Frontity.

**How**:

Refactoring the code to avoid some warnings, documenting everything, and adding e2e tests.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests
- [x] TypeScript tests
- [x] Update community discussions (https://community.frontity.org/t/infinite-scroll-hooks/2055/31)

**Unrelated Tasks**

- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


luisherranz:
@darerodz I've finished the merge or tiny-router `actions.ts` in your Live Share session. 

I cannot run Prettier. I'm not sure if that's because there is an error in the code or because of Live Share. Please take a look.

I've also merged the tiny-router tests, although I was not able to run them. The merge seems fine to me but maybe it contains errors so please take a look.

I'll leave everything else to you!

DAreRodz:
I just saw it, thanks @luisherranz!

DAreRodz:
I was able to fix the actions and tests of the `@frontity/tiny-router` package, so those are passing now. There are still some unit tests failing, though, related to the infinite scroll hooks. I'll try to fix them tomorrow.

luisherranz:
Thanks, David. I'll do the review next week 🙂

---

For context: This PR contains the code of the hooks, which was made by @orballo, and a refactoring and some e2e tests made after that by @darerodz.

I already reviewed @orballo's code, so I am going to review only @darerodz's updates: https://github.com/frontity/frontity/pull/652/files/229d9bad488609ae35e7ae0e5b415cfa9dad4f78..14898979742d14ab589aaa0d075a4b1c244c8970?file-filters%5B%5D=

---

@santosguillamot, we should credit both @orballo and @darerodz when we do the release. We can edit the `CHANGELOG.md` file manually before we merge the new version that contains this 🙂

luisherranz (changes requested):
I have a concern about redirections. Imagine this list of events:

1. The user loads `/some-post`.
   - This theme is using infinite scroll for posts.
2. The user scrolls down.
   - The `actions.source.fetch()` is triggered for the next post, `/another-post`.
3. `/another-post` has a 301 redirection to `/redirected-post`.
   - The 301 redirection is stored in `state.source.data`.
4. The user keeps scrolling down.
   - The `actions.router.set("/another-post", { method: "replace", state: { ... } })` is triggered.
   - It includes the proper `state` for `window.history.state`.
5. `actions.router.set` checks for a redirection and sees that there is one in `state.source.data`.
   - It replaces `link` internally from `/another-post` to `/redirected-post`.
   - It continues the `actions.router.set` and saves the `state` in `window.history`.
   - The internal `link` stored in `window.history.state` is still `/another-post`, not `/redirected-post`.
6. The user clicks on a link and goes forward.
7. The user clicks on the back button of the browser.
8. The infinite scroll hook uses `window.history.state` to restore the page.
   - The second link stored in `window.history.state` is `/another-post`, not `/redirected-post.`

My guess is that it won't work because it will restore `/another-post` instead of `/redirected-post`.

If we land in the `/another-post`, [the observer of the `tiny-router`](https://github.com/frontity/frontity/blob/dev/packages/tiny-router/src/actions.ts#L127-L148) will trigger a new `actions.router.set("/redirected-post", { method: "replace" })` that doesn't have the proper state for `window.history.state`.

If we don't land in `/another-post`, but in a previous or next post, I think the `/another-post` will be blank, maybe will even fail because it doesn't have a `data.type` and `data.id`, and the scroll will not be restored properly. And even if it works, whenever the user scrolls into that post, the `tiny-router` observer will kick in and will trigger the `actions.router.set` that will break the infinite scroll.

@DAreRodz I think we should add an e2e for this use case to see if I am right, and if I am, add support for redirections in the infinite scrolls. It shouldn't be hard to do.

DAreRodz:
> @DAreRodz I think we should add an e2e for this use case to see if I am right, and if I am, add support for redirections in the infinite scrolls. It shouldn't be hard to do.

I'll add a test for that case, let's see what happens. Thanks for the review, @luisherranz. :blush:

.changeset/config.json
```diff
@@ -7,7 +7,7 @@
   "commit": false,
   "linked": [],
   "access": "public",
-  "baseBranch": "dev",
+  "baseBranch": "infinite-scroll-beta",
```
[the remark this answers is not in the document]
DAreRodz: Done!

BTW, I wasn't aware of that change. Is it something that changed while entering in "pre" mode? :thinking:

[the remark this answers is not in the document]
luisherranz: Yes :)

luisherranz:
Thanks David!

DAreRodz:
I wrote a simple test and it failed even before, in steps 4. and 5.

The reason it failed in step 4 was the way the e2e test was implemented. The `Wrapper` component is rendered only when the data object of a link is a post type. Redirections are not post types, so the `Wrapper` was missing, and that component is the one resposible of setting the new link. In fact, it was like removing the URL from the list of post links that the hook returns.

This is the relevant code from the `PostType` component used in the e2e tests:

https://github.com/frontity/frontity/blob/305259d627f7646aa6a645dd83c2757cbd216fe7/e2e/packages/use-infinite-scroll/src/components/post-type.tsx#L49-L59

After that, I tried rendering the Wrapper even though `data` was not a post type, and also reactivating the `autoFetch` feature of tiny-router (it was disabled for these tests). Now, for some reason, once the redirected post is fetched, the scroll jumps to the top, changing the link to be the one of the first post.

I'm investigating that.

luisherranz:
Ok, thanks David.

Maybe the infinite scroll hooks need to check if the next link is a redirection (using `isRedirection(data)`) and if it is, change the link (using `data.location`)?

Similar to what `tiny-router` does inside `actions.router.set()`: https://github.com/frontity/frontity/blob/dev/packages/tiny-router/src/actions.ts#L67-L84

DAreRodz:
I'll try to follow that approach and see if it works (I hope so :grin:).

DAreRodz:
I've refactored `useInfiniteScroll()` and modified it to handle redirections. All tests are passing now, including the e2e test to check if redirections are working. :tada: 

However, there are still some cases I want to check (like external redirections) and some unit tests missings I want to add as well. I'm currently working on it.

Once I finish that, I'll record a video explaining the new changes and request a review again. :slightly_smiling_face: 

cc: @luisherranz

luisherranz:
Ok, thanks David 🙂

DAreRodz:
@luisherranz I thought last Thursday I had fixed the problem with redirections, but it turned out to be more complex than expected.

Anyway, don't worry, I think I'm close to solving it. I can post the details once I 'm done with the code or, if you prefer, we can have a meeting to discuss the implementation.

luisherranz:
Sure, we can have a meeting.

DAreRodz:
Hey, so you know, we decided not to implement support for redirections of **posts that exist in an archive** but now point to another post/page/etc.

The `useInfiniteScroll()` hooks should handle them as if those links weren't included in `data.items`.

Here's some reasons behind that decision:

- There is no guarantee for a post type to be redirected to another post type of the same type/archive/etc. In those cases, it doesn't make sense to include the redirected page between the posts the user is reading.
- If a post type is redirected to another post of the same type in the same archive, it will probably appear in that archive before the redirected link, so it' would be enough to remove the old link from the archive.
- External redirections should be ignored anyway. It doesn't make sense ―from a user's point of view― to scroll down and suddenly appear on a different webpage.

I'm working on implementing the logic to remove redirected links from `data.items`. :man_technologist:

DAreRodz:
@luisherranz I've recorded a video in which I explain the latest changes I've made since your previous review. If you have any question just let me know.

BTW, the unit tests are failing again just after merging `dev` into this branch, and it seems is not related to these changes. I'm taking a look to know what's going on.

<a href="https://www.loom.com/share/8f8197e1eceb4750a5f117c5a5eb487f"></a>

DAreRodz:
> BTW, the unit tests are failing again just after merging `dev` into this branch, and it seems is not related to these changes. I'm taking a look to know what's going on.

It looks like it's a problem in the `dev` branch. :eyes:

DAreRodz:
Oops, I forgot to request another review. @luisherranz, just ignore the tests that are failing, that problem will be addresed by #721. You can review this PR again.

packages/hooks/use-archive-infinite-scroll/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  // Aliases to needed state.
+  const last = state.source.get(links[links.length - 1]);
+  const next =
+    isArchive(last) && last.next ? state.source.get(last.next) : null;
+
+  // Infinite scroll booleans.
+  const hasReachedLimit = !!limit && links.length >= limit;
+  const thereIsNext = isArchive(last) && last.next;
+  const isFetching = last.isFetching || !!next?.isFetching;
```
luisherranz, line 123: I have a question here: by using `!!next` you are turning the `next` data object into a boolean.

So this code `!!next?.isFetching` means: 
1. Turn the `next` data object in a boolean.
2. Check if it is truthy. As it is a boolean, that can only be `true`.
3. If it is `true`, check its `isFetching` prop.

So if I am not mistaken, you are doing `true.isFetching`, which doesn't fail, but always returns `undefined`.

luisherranz (commented):
Awesome video. Thanks a lot, David 🙂

I like the new hooks folder organization and the fact that we now have the types in: `@frontity/hooks/use-in-view/types`. It looks clean and follows the same principle as the rest of the packages.

I also like the fact that we are adding links to the browser state only when they are ready. It makes more sense to me as well.

I'm still doing the review but I have a question. It is in a comment below.

packages/hooks/use-post-type-infinite-scroll/index.tsx
```diff
@@ -0,0 +1,375 @@
...
+  const archive: string = (() => {
+    // If `archive` is set in options, use it.
+    if (options.archive) {
+      return options.archive;
+    }
+
+    // If `archive` is already in the state, use it.
+    if (infiniteScroll?.archive) {
+      return infiniteScroll.archive;
+    }
```
luisherranz, line 103: Shouldn't this be the other way around? First, check in the browser state and then in the options?

packages/hooks/use-archive-infinite-scroll/index.tsx
DAreRodz: As far as I know, the Optional Chaining operator (`.?`) takes precedence over the Logical NOT (`!`) operator, same as the Member Access operator (`.`).

Here is a [table from MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table) with all the JS operators listed by precedence, I guess TypeScript maintains the same order.

So, actually, `!!next?.isFetching` means `!! (next?.isFetching)`. :blush:

packages/hooks/use-post-type-infinite-scroll/index.tsx
DAreRodz: I didn't change that part. If you think it makes more sense ―I think that too― I'll change it.

Maybe @orballo has something to say in this regard. :thinking:

packages/hooks/use-archive-infinite-scroll/index.tsx
DAreRodz: I just did a quick test on the TypeScript Playground and, yeah, this is the way the code above is compiled:

```js
!!(next === null || next === void 0 ? void 0 : next.isFetching);
```

luisherranz (dismissed):
I think that the logic of `getLinksFromPages` fails when the initial link (`firstLink`) is not on the first page. It is explained here: https://www.loom.com/share/01153232dfb842bfa9d6de9117d82da0

Also, please add more tests for other edge cases of `getLinksFromPages`.

The commit with the new tests is: https://github.com/frontity/frontity/pull/652/commits/dec292d6d47a2a336e4e3c097e4c87d72a788e59

packages/hooks/use-archive-infinite-scroll/index.tsx
luisherranz: Ok, thanks for the clarification.

By the way, what is the benefit of using `!!next?.isFetching` over `next?.isFetching`? The type?

DAreRodz: Yest, it's just a way to ensure that `isFetching` holds a boolean value. Also, that's the value the hook returns, so I think it makes sense regarding the API users will consume.

On the other hand, we are not using the `strictNullChecks` flag, so there aren't any TS errors if you assign `undefined` to a boolean variable, but I think it's better anyway if we take users into account.

luisherranz: Perfect. Thanks for the explanation David 👍

packages/hooks/use-post-type-infinite-scroll/index.tsx
luisherranz: The question is: if `options.archive` is changed on the fly to a different archive, should the hook also change, or should the hook stick to the initial value of `options.archive`? 

And even before that question: if `options.archive` is changed on the fly to a different archive, will it break the hook?

DAreRodz:
> I think that the logic of `getLinksFromPages` fails when the initial link (`firstLink`) is not on the first page.

I know it could look weird or be confusing, but it works the way it is. In the case you included in your test, as the `itemLinks` that `getLinksFromPages()` returns is used only to get the next link, what happes is the following:

1. `getLinksFromPages()` returns `itemLinks` without `/post-1`.

   ```ts
   const itemLinks = getLinksFromPages({ firstLink: "/post-1", ... });
   // itemLinks =  ["/post-2", "/post-3", "/post-4"]
   ```

2. `usePostTypeInfiniteScroll` gets the index of `currentLink` (which is `/post-1`). It returns `-1` (not found).

   ```ts
   const currentIndex = itemLinks.indexOf("/post-1")
   // currentIndex = -1
   ```

3.  The next link is obtained from adding 1 to the previous index (-1 + 1 = 0). It is `/post-2`.

    ```ts
    const nextLink = itemLinks[currentIndex + 1];
    // nextLink = itemLinks[0] = "/post-2"
    ```

So, it works because the next link you get is the correct one.

---

Anyway, we could follow a different approach and the behavior would be the same. Adding the link in the first position if it doesn't exist in the page 1, and after that, ensuring that posts are not repeated using a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).

Uhm, I think I'm going to do that change. For me, it feels safer as any kind of duplication is avoided, not only with the first link.

DAreRodz:
I simplified the code following the approach I mentioned in my previous comment. I think it's more clear now. :ok_hand:

packages/hooks/use-post-type-infinite-scroll/index.tsx
DAreRodz: Good question.

I think we should add a test for that case and see what happens. My guess is that it should not break, because `archive` is only used to initialize `pages` the first time the hook is executed (when the pages array is not defined yet). Therefore, the pages array is preserved and the hook should work the same.

https://github.com/frontity/frontity/blob/dec292d6d47a2a336e4e3c097e4c87d72a788e59/packages/hooks/use-post-type-infinite-scroll/index.tsx#L192-L195

My concern now is, what happen if you want to change the archive but `pages` already exists? I'll add a test for that as well.

luisherranz:
Perfect, thanks for the new tests. It looks much more solid now 🙂

I will continue the review as soon as I can.

packages/hooks/use-post-type-infinite-scroll/index.tsx
DAreRodz: I've added two e2e tests more to check how the `usePostTypeInfiniteScroll()` hook behaves when the `archive` param changes. Both are passing without changing the code. https://github.com/frontity/frontity/pull/652/commits/60ddde54316f60285cfae65121ed34122a7986d0

DAreRodz:
I've added more e2e tests, as I mentioned here: https://github.com/frontity/frontity/pull/652#discussion_r580159817

- One is for the case where you visit two different archives.
- Another one for the case where you change the archive source while scrolling.

packages/hooks/use-post-type-infinite-scroll/index.tsx
luisherranz: Awesome. Thank you David.

luisherranz:
@darerodz, the unit tests are failing with the last commit. Could you please check them out?

e2e/integration/frontity-01/use-infinite-scroll.spec.ts
```diff
@@ -443,6 +443,83 @@ describe("useArchiveInfiniteScroll", () => {
...
+    // cy.location("href").should(
+    //   "eq",
+    //   "http://localhost:3001/category/two/page/2/"
+    // );
+
+    // // Go back, the link should point to "/page/2/"
+    // cy.go("back");
+    // cy.get("[data-test='fetching']").should("not.exist");
+    // cy.get("[data-test='page-2']").should("exist").should("be.visible");
+    // cy.location("href").should("eq", "http://localhost:3001/page/2/");
```
luisherranz: Is this ok? Does it need to be removed?

packages/hooks/use-post-type-infinite-scroll/types.ts
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * If this post is the last post. Useful to add separators between posts,
+     * but avoid adding it for the last one.
+     */
+    isLast: boolean;
+
+    /**
+     * The Wrapper component that should wrap the real `Post` component.
+     */
+    Wrapper: React.FC;
```
luisherranz: Can we improve the type of the final `Wrapper` components?

I was thinking about at least a `key` prop, but I want to know your opinion: 

```ts
Wrapper: React.FC<{ key: string; }>;
```

packages/hooks/use-archive-infinite-scroll/types.ts
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * If this page is the last page. Useful to add separators between pages,
+     * but avoid adding it for the last one.
+     */
+    isLast: boolean;
+
+    /**
+     * The Wrapper component that should wrap the real `Archive` component.
+     */
+    Wrapper: React.FC;
```
luisherranz: Same here.

packages/hooks/use-infinite-scroll/index.ts
```diff
@@ -0,0 +1,132 @@
...
+      if (isNextReady) {
+        links.push(next.link);
+
+        // Update the browser's history state with the new link once it is
+        // ready.
+        actions.router.updateState({
+          ...state.router.state,
+          infiniteScroll: { ...infiniteScroll, links },
+        });
+      }
```
luisherranz: What do you think about this change? -> Don't update the state until the link is ready.

I've changed it because I always thought it would work that way, but I want to know your opinion. Is it safe? Is it better?

packages/hooks/use-archive-infinite-scroll/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+    isLast:
+      (link === last.link && !!last.isReady && !isError(last)) ||
+      (link === links[links.length - 2] && !last.isReady) ||
+      (link === links[links.length - 2] && !!last.isReady && !!isError(last)),
+    Wrapper: options.active
+      ? MemoizedWrapper(link, {
+          fetchInViewOptions: options.fetchInViewOptions,
+          routeInViewOptions: options.routeInViewOptions,
+        })
+      : ({ children }) => <>{children}</>,
```
luisherranz, line 180: In this hook, if it is not active we don't use the `Wrapper`, which I think makes sense.

But in the other one, we always use the `Wrapper`, even when it is not active.

https://github.com/frontity/frontity/blob/29730058958ab880c2563dce3f135eeb76d72b99/packages/hooks/use-post-type-infinite-scroll/index.tsx#L329-L332

Is there a reason for that?

luisherranz (commented):
I have finished reviewing the code and I think **this is fantastic and ready for prime time** 👏 👏 

_I still want to take a final look at the tests tomorrow morning._

---

I've done a few small changes here and there. Mostly add comments and I refactored the `<Wrapper>` component so both hooks use the same code.

I also want your opinion about some things (asked in the code).

e2e/integration/frontity-01/use-infinite-scroll.spec.ts
DAreRodz: Nope, I just forgot to uncomment those lines of code. :sweat_smile: It's fixed now.

packages/hooks/use-post-type-infinite-scroll/types.ts
DAreRodz: Yes, that would be a good idea. We could use that `Wrapper` type here and also move these props to it.

https://github.com/frontity/frontity/blob/09ec9433b5b74fc6435a30326697971d9a36febc/packages/hooks/use-post-type-infinite-scroll/index.tsx#L31-L36

By the way, I see what you did. You refactored the way `Wrapper` components are generated and moved that logic to some utils inside the `use-infinite-scroll` folder. :raised_hands: 

My only concern is that it could be confusing in some way because the exported `Wrapper` component is in fact a kind of internal wrapper used inside the final `Wrapper` that developers get when calling `useArchiveInfiniteScroll` or `usePostTypeInfiniteScroll`.

So, instead of using `Wrapper` component and `WrapperProps` for those internal things, would it make sense if we reserve those names for the `Wrapper` components returned by the hooks?

packages/hooks/use-archive-infinite-scroll/types.ts
DAreRodz: I answered in the previous comment: https://github.com/frontity/frontity/pull/652#discussion_r583545884

packages/hooks/use-archive-infinite-scroll/index.tsx
DAreRodz: Well, I'm pretty sure I changed that for the `useArchiveInfiniteScroll` because the next page was being fetched even though the hook was disabled, as the `<Wrapper>` is the one in charge of doing that (the wrapper is not disabled).

The same doesn't happened with the `usePostTypeInfiniteScroll` because the next page is fetched when you reach the penultimate post. But, it could happen if you are in the penultimate post and then you disable the hook... :thinking:

I'm going to add a test for that and fix it.

e2e/integration/frontity-01/use-infinite-scroll.spec.ts
luisherranz: Thanks! 👍

packages/hooks/use-post-type-infinite-scroll/types.ts
luisherranz: Absolutely. Good point. Can you rename it?

packages/hooks/use-archive-infinite-scroll/index.tsx
luisherranz: Perfect, thank you! It's not a bid deal but I think it will be more solid that way 🙂

packages/hooks/use-infinite-scroll/index.ts
DAreRodz: The only thing I see is that we're only updating the `infiniteScroll` state when the `next` link is ready, so the first link would not be added until that moment, right?

You would have this in `state.router.state`

```ts
state.router.state = {};
```

and then this

```ts
state.router.state = {
  infiniteScroll: {
    links ["/first-link/", "/second-link/"],
  },
};
```

I guess there's no problem with the high-level hooks because they are doing the initialization, but maybe we should do the same with this one.

luisherranz: Gotcha. I'm going to revert the change 👍

packages/hooks/use-post-type-infinite-scroll/types.ts
DAreRodz: Yup, I'll try to find a good name. :+1:

DAreRodz: I renamed it to `InternalWrapper` and `InternalWrapperProps`, and created `WrapperProps`, which is the type used by the `<Wrapper>` components returned by the hooks.

packages/hooks/use-archive-infinite-scroll/index.tsx
DAreRodz: It turns out I was already doing that test. :joy: I added a missing test anyway to check that the next post wasn't rendered or fetched when the hook is deactivated, and It worked, so I didn't change the code.

packages/tiny-router/src/__tests__/index.test.ts
```diff
@@ -13,10 +13,13 @@ let get: jest.Mock;
...
 
+const spiedPushState = jest.spyOn(window.history, "pushState");
+const spiedReplaceState = jest.spyOn(window.history, "replaceState");
+
 beforeEach(() => {
   normalize = jest.fn().mockImplementation((link) => {
-    const { pathname, hash, search } = new URL(link, "https://dummy.com");
-    return pathname + hash + search;
+    const { pathname, search, hash } = new URL(link, "https://dummy.com");
+    return pathname + search + hash;
```
DAreRodz, line 22: I thought this code was wrong only in this branch but it's in `dev` too! I've added a changeset for these changes.

packages/hooks/use-post-type-infinite-scroll/types.ts
luisherranz: Awesome. Thanks a lot David!

packages/hooks/use-archive-infinite-scroll/index.tsx
luisherranz: Perfect, thank you!

packages/tiny-router/src/__tests__/index.test.ts
luisherranz: Great, good catch!

luisherranz (approved):
I don't have words to say how amazing this feature is guys. This implementation is simply brilliant. And the test coverage is outstanding.

It is the best infinite-scroll feature a frontend framework could dream of: 
- 100% of abstraction of the infinite scroll logic.
- 100% of control of the layout to the developer.

Congratulations to both @orballo and @darerodz for your amazing work 👏👏👏

BRAVO! 🎉😄

---

PS: I have a last question, but it is not important. @darerodz, You can merge it as it if you want.

packages/hooks/use-archive-infinite-scroll/__tests__/use-archive-infinite-scroll.test.tsx
```diff
@@ -0,0 +1,1098 @@
...
+          link: "/page/2/",
+          isLast: true,
+          Wrapper: expect.any(Function),
+        },
+        {
+          key: "/page/3/",
+          link: "/page/3/",
+          isLast: false,
+          Wrapper: expect.any(Function),
+        },
```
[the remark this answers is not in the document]
DAreRodz: I just leaved that code as it is because there is no much of a difference between both approaches. In this case, it wouldn't be probable to have redirected pages – the developer is the one choosing what archive is used as the "source" of posts, and I don't think someone would redirect the `/page/3/` of an archive, for example.

For that reason, as you woudn't have to remove pages from the `pages` array, I didn't change the original implementation.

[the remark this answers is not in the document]
luisherranz: I see. Thanks David. Maybe it is something we can change in future refactorings.

PR #640, opened 2020-12-07
Update Babel Polyfills

luisherranz:
**What**:

I have updated our Babel configuration to make it work with the latest versions of Babel.

**Why**:

We were not injecting the polyfills correctly and some of them were missing, both in the modules and es5 versions.

**How**:

- I have updated `babel-preset-env` to the latest version, which relies on `core-js` v3 instead of `@babel/polyfill`.
- I extracted a list of all the polyfills added to both `es5` and `module`.
- I switched `useBuiltIns` from `"entry"` to `"usage"` in `babel-preset-env` so only polyfills that are actually used in the code are added. That means we don't need to add any entry to our files.
- I had to exclude `webpack`, `core-js` and `regenerator-runtime` from babel itself because if they are transpiled they don't work.
- I had to switch to `sourceType: unambiguous` to make this new configuration work.
- I have updated the target version of node to 10 (previously it was 8).
- Now both `module` and `node` targets use core-js. Previously they didn't.
- I have added the `bugfixes` option because it will be the default in Babel 8.
- I have updated the packages of `@frontity/core`.
- I have updated the packages of `frontity`.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] Add a changeset (with link to its [Feature Discussion](https://community.frontity.org/c/33) if it exists)

**Unrelated Tasks**

- [ ] Update `webpack-dev-middleware` once this bug is fixed: https://github.com/webpack/webpack-dev-middleware/issues/782
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages
- [ ] Open an issue for this feature in the [docs repo](https://github.com/frontity/docs/wiki/Code-Releases)
- [ ] Update community discussions

**Additional Comments**

There is a bug right now in the latest version of `webpack-dev-middleware` that outputs `<i>` in the console: https://github.com/webpack/webpack-dev-middleware/issues/782

Screenshot 2020-12-08 at 13 45 17

As soon as it is fixed, I will update our version.

PR #661, opened 2021-01-12
Initial updates for react@17

### LINKED ISSUE #631: Update React and Emotion and update Babel to use the new JSX transform [13pt]
Feature: [Update to React 17 and Emotion 11](https://community.frontity.org/t/update-to-react-17-and-emotion-11/3225/).

This feature affects the issue to [Move all the CSS to the `<head>`](https://github.com/frontity/frontity/issues/559), so we should take this into account or even address both at the same time.

[Sprint Card](https://github.com/orgs/frontity/projects/12#card-50175547)

cristianbote:
**What**:

This is in regard of #631 and it covers the React update part.

**Why**:

In order to be up to date with React ecosystem and new improvements we need to stay up to date.

**How**:

Updated package version and types.

**Tasks**:

- [x] Update other packages

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes

**Additional Comments**

~Right now there's a warning because `React` is missing, if we remove the import. So the new `jsx-runtime` is not getting captured by the types yet. Looking into it.~ This one got solved by the tsconfig changes.


luisherranz:
@cristianbote oh, could you please rebase this on top of the `babel-update` branch?

- https://github.com/frontity/frontity/tree/update-babel
- https://github.com/frontity/frontity/pull/640

If not, we are going to have conflicts later. I updated many of the `babel` dependencies as well.

_I was working first on the 30X redirections PR but I'll switch to that PR now, so we can merge it before you finish with this one._

luisherranz:
https://github.com/frontity/frontity/pull/640 is merged so you don't need to rebase anymore, just merge `dev` 🙂

cristianbote:
Moving forward, how should we handle React types? 🤔 

For example this:
```tsx
import React from 'react';

const Root: React.FC = ({ state, libraries }) => {
  return <div>root</div>
};
```

should we change it to explicitly named imports for types?
```tsx
import { FC } from 'react';

const Root: FC = ({ state, libraries }) => {
  return <div>root</div>
};
```

Asking cause my type skills are not that strong - yet! 😄 , and not sure what's the status quo on here.

cristianbote:
One other thing does this feel like a minor bump? @luisherranz

luisherranz:
> One other thing does this feel like a minor bump?

Yes, I think it should be minor because the API is backward compatible, but I haven't done a thorough review. Strictly speaking, I guess it should be a major but I would like to avoid that if we can.

Do you see a strong reason to do a major? Have you reviewed the list of breaking changes of React 17? https://reactjs.org/blog/2020/08/10/react-v17-rc.html#other-breaking-changes

cristianbote:
Yup! I did looked over it and all the breaking changes are backward compatible. The `useEffect` timing might be one of the things that it could be breaking but we can't know until we run into an issue. I did scanned the codebase for the use-case they are saying could cause an issue but we are not using it in that way.

I think a minor would suffice as well. Will push the changeset as well. 👍  Thank you!

luisherranz:
> should we change it to explicitly named imports for types?

Yes, I think using named imports to import the types is fine 👍

.changeset/pretty-dingos-travel.md
```diff
@@ -0,0 +1,14 @@
...
+"@frontity/components": minor
+"@frontity/comscore-analytics": minor
+"@frontity/connect": minor
+"@frontity/core": minor
+"frontity": minor
+"@frontity/google-ad-manager": minor
+"@frontity/google-analytics": minor
+"@frontity/google-tag-manager-analytics": minor
+"@frontity/html2react": minor
+"@frontity/smart-adserver": minor
```
luisherranz: I understand that `@frontity/core` needs a minor bump, but why the rest?

luisherranz (commented):
It looks good to me, but as this is a rather big change, let's wait until @darerodz finishes the e2e fix and we can run all the tests in this PR.

.changeset/pretty-dingos-travel.md
cristianbote: The bunch of them have updates to `react-test-renderer` and in changeset I selected all of them. That's all there is. Should I switch to patch instead? Might be better since the updates are not affecting the user-facing code, only the internal. Gonna patch them.

cristianbote:
Uuuh yes! 👍  It would be great to have the e2e suite running on the changes 😄  Will wait on e2e PR then. Thank you!

.changeset/pretty-dingos-travel.md
luisherranz: > Might be better since the updates are not affecting the user-facing code, only the internal

Actually, we don't release a new version when updates don't affect user-facing code.

There is one exception, which is changes to the readme file because if we don't release a new version the readme of NPM is not updated. So I think you can remove them from the changeset and leave `@frontity/core` alone.

luisherranz: I'm going to request the change then 🙂

.changeset/pretty-dingos-travel.md
```diff
@@ -0,0 +1,14 @@
...
+"@frontity/components": patch
+"@frontity/comscore-analytics": patch
+"@frontity/connect": patch
+"@frontity/core": minor
+"frontity": patch
+"@frontity/google-ad-manager": patch
+"@frontity/google-analytics": patch
+"@frontity/google-tag-manager-analytics": patch
+"@frontity/html2react": patch
+"@frontity/smart-adserver": patch
```
luisherranz: You can remove all the packages from the changeset except for `@frontity/core` (discussion: https://github.com/frontity/frontity/pull/661#discussion_r556687124).

```suggestion
"@frontity/core": minor
```

luisherranz (changes requested):

.changeset/pretty-dingos-travel.md
cristianbote: Oh, even better! 😄  Thank you!

cristianbote: Done!

luisherranz (approved):
I have run the e2e tests locally and they pass (except for the AIOSEOP ones, but that's not this PR's fault).

It's approved so you don't depend on me anymore, but I think it's better to not merge it until all the e2e tests pass here as well 👍

PR #663, opened 2021-01-13
Update Jest config so we can use Typescript 4.1

luisherranz:
**What**:

Fix the problem with `ts-jest` and TypeScript 4.1.

**Why**:

`ts-jest` was failing when using TypeScript 4.1 returning a `cannot read property 'length' of undefined` error. It worked fine with TypeScript 4.0.5.

It seems like it is a common problem in TypeScript 4.0/4.1: https://github.com/microsoft/TypeScript/issues?q=is%3Aissue+is%3Aopen+cannot+read+property+%27length%27+of+undefined+

**How**:

I used `babel-jest` to process JS files and it seems like it solved the problem.

Video: https://www.loom.com/share/f6666eb5ca614838973cff478bcd861e

**Tasks**:

- [x] Jest Config

**Unrelated Tasks**

- [ ] Code
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

I fixed a small bug that was making a unit test fail: https://github.com/frontity/frontity/pull/663/commits/e69f6fb6287d48dfcd334c2672222bd2b856d1f7


jest.config.js
```diff
@@ -1,5 +1,11 @@
+// Removes the warning that we are using an unsupported version of TypeScript.
+process.env.TS_JEST_DISABLE_VER_CHECKER = false;
+
 module.exports = {
-  transform: { "^.+\\.(t|j)sx?$": "ts-jest" },
+  transform: {
+    "^.+\\.tsx?$": "ts-jest",
+    "^.+\\.jsx?$": ["babel-jest", { configFile: "../../jest.babel.js" }],
```
cristianbote: How do you feel if we keep all of the config files with `*.config.js` at the end? So this would be `jest.babel.config.js`?

jest.config.js
```diff
@@ -1,5 +1,11 @@
+// Removes the warning that we are using an unsupported version of TypeScript.
+process.env.TS_JEST_DISABLE_VER_CHECKER = false;
```
cristianbote: Oh does this needs to be `true`? 🤔

luisherranz: Yeah, you're right 😅 

https://kulshekhar.github.io/ts-jest/docs/version-checking/

luisherranz: Done.

jest.config.js
luisherranz: I had that at the beginning and felt too long, but it's fine. I'll change it again 👍

luisherranz: Done.

cristianbote (approved):
Awesome work 👍🙏 thank you for looking into it!

luisherranz:
Thanks Cristian. I'll merge it as soon as the unit tests pass so you can use it in your branch.

luisherranz:
Done.

Let's see if David can also fix the e2e tests and get back on all our  ✅ 's again 😄

PR #664, opened 2021-01-14
Migrate Cypress to 6.2.1

### LINKED ISSUE #659: Cypress e2e tests get stuck
**_Bug report_**

**Expected behavior**

The e2e tests should be reliable and pass or fail and not get stuck.

**Observed behavior**

They often get stuck so it prevents and they don't return an error, so we are not able to check if tests are passing or not. It seems it isn't related to any specific test, because it gets stuck on different ones.

**Possible solution**

Not sure if this will be helpful but there are other users complaining about this in the [Cypress repo](https://github.com/cypress-io/cypress/). Some fixes they mention:
- Add `options: --shm-size 2gb` as explained in [this comment](https://github.com/cypress-io/cypress/issues/6695#issuecomment-627363078) and the following ones.
- Use `--disable-dev-shm-usage` as a flag rather than trying to set the `--shm-size` as explained in [this comment](https://github.com/cypress-io/cypress/issues/8206#issuecomment-712458749).
- Pass tests one at a time seems to help as explained in [this comment](https://github.com/cypress-io/cypress/issues/8206#issuecomment-682721485) and [here](https://github.com/cypress-io/cypress/issues/8206#issuecomment-745117472).
- Not using chrome as explained in [this comment](https://github.com/cypress-io/cypress/issues/8206#issuecomment-675663487), although the [next one](https://github.com/cypress-io/cypress/issues/8206#issuecomment-676588332) says it didn't work for them.

DAreRodz:
**What**:

Bump Cypress version and adapt the e2e tests. Fixes #659 

**Why**:

The e2e tests were getting stuck more and more times.

**How**:

Updating Cypress to 6.2.1. That version includes a bugfix for an issue that probably was the problem we had. https://github.com/cypress-io/cypress/issues/9040

**Tasks**:

- [x] End to end tests

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

The e2e tests for the `@frontity/smart-adserver` package were deactivated due to a problem with the ad units that should be addressed by the SmartAdserver team. They are in known of the issue.

Also, the version of the [All In One SEO Pack](https://wordpress.org/plugins/all-in-one-seo-pack/) plugin was fixed to 3.7.1 because the [REST API - Head Tags](https://wordpress.org/plugins/rest-api-head-tags/) plugin doesn't work with newer versions. I'll open an issue in the `wp-plugins` repository and link to this one.


e2e/integration/frontity-01/image.spec.js
```diff
@@ -70,9 +70,8 @@ describe("Image lazy-loading (with native lazy-load)", () => {
 
   it("should render an image without loading if it doesn't have height", () => {
     cy.scrollTo("topLeft");
-    cy.get("img:not([height])")
-      .should("not.have.attr", "loading")
-      .should("not.be.visible");
```
DAreRodz, line 75: Before Cypress 6.0 you could assert that `undefined` is not visible (`.should()` returns `undefined`). That's not allowed anymore.

DAreRodz:
All checks have passed! :tada:

cristianbote:
Amazing job @DAreRodz! 😃

luisherranz:
💪 💪 

dance

luisherranz (approved):
Looks good to me David!

So we can use `cy.intercept()` from now to mock `fetch`, right?

DAreRodz:
> So we can use `cy.intercept()` from now to mock `fetch`, right?

Sure! From the [Cypress docs](https://docs.cypress.io/api/commands/intercept.html#Comparison-to-cy-route):

> Unlike `cy.route()`, `cy.intercept()`:
>
> - can intercept all types of network requests including **Fetch API**, page loads, XMLHttpRequests, resource loads, etc.

DAreRodz:
BTW, I'm taking a look at this before merging the PR: https://github.com/frontity/frontity/issues/659#issuecomment-760061087

DAreRodz:
It seems that the Yoast SEO plugin was updated 2 days ago, and now it's adding a few tags more. I've updated the snapshots.

DAreRodz:
Weird.

Now, if I update the snapshots, the Yoast e2e test for the previous WordPress version fails, and the opposite happens if I revert the changes.

That could mean that the Yoast SEO plugin behaves differently depending on the WordPress version... :sweat:

I'll investigate it tomorrow.

DAreRodz:
Just so you know:

- The _previous_ version of WordPress is currently 5.5, not 5.4, so I've updated it.
- As Yoast SEO requires WordPress ^5.5, updating the WP version fixed the problems with the snapshots.
- I've deactivated the "e2e-tests (Previous WordPress Version, 5.4, --prod --wp-version 5.4)" status check on the `dev` branch and set the new one for WP 5.5 as required.

Everything is working now. :relaxed:

luisherranz:
> I've deactivated the "e2e-tests (Previous WordPress Version, 5.4, --prod --wp-version 5.4)" status check on the dev branch and set the new one for WP 5.5 as required.

Oh, we should change the name of that workflow to something generic so we don't have to change the GH settings each time we update. My fault.

I'm going to do a quick PR for that change.

Apart from that, thanks @DAreRodz! Great work 😄

luisherranz:
> Oh, we should change the name of that workflow to something generic so we don't have to change the GH settings each time we update.

Ok, my bad again, that is not possible because GH also uses the options, not only the name.

Nevermind, it's not a bid deal.

PR #666, opened 2021-01-15
Adopt emotion 11

### LINKED ISSUE #631: Update React and Emotion and update Babel to use the new JSX transform [13pt]

cristianbote:
**What**:

Adopt the new version of emotion 11! 🎉 

**Why**:

The new emotion 11 adds a lot of improvements and besides all of that we can make use of a few nifty features that are brought with it's babel-plugin: source maps, auto labels, css prop that all of them work seamless with the new jsx runtime.

**How**:

I've updated the new dependencies in packages and added the needed configuration.

**Tasks**:

- [x] Update other packages

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes

**Additional Comments**

The changes are based on the react 17 update branch. I suggest that the base should be merged in first to avoid conflicts and messed up changes.


examples/mars-theme-example/package.json
```diff
@@ -24,7 +24,6 @@
     "@frontity/mars-theme": "^1.3.1",
     "@frontity/tiny-router": "^1.1.0",
     "@frontity/wp-source": "^1.5.0",
-    "babel-plugin-frontity": "^1.0.1",
```
cristianbote, line 27: @luisherranz Noticed that our plugin is a dependency in all of the examples. Should I remove the dependency from the packages?

Also how should we deprecate the plugin? Remove it entirely?

luisherranz: > Should I remove the dependency from the packages?

That's weird. That dependency should not be there, it makes no sense.

> Also how should we deprecate the plugin? Remove it entirely?

Nah, I don't think so. Nobody uses it, only `@frontity/core` internally. And I think that if we deprecate it, it will throw a warning for people still on old `@frontity/core` versions.

cristianbote:
Forgot to mention that while I was debugging the `autoLabel` defaults I've stumbled into an issue where the label is kept in production, and it seems emotion has a fix for it, awaiting to be merged https://github.com/emotion-js/emotion/pull/2157/files

cristianbote:
The error that the tests are reporting is a `type` mismatch/unrecognised issue not solved by the new pragma. Have you had these kind of issues before? 🤔 

Looking through emotion's tracker they do seem to have folks report this, but usually the culprit is the TypeScript version. But we are on the new typescript version, so 🤔  wonder why it fails?

cristianbote:
Filled an issue on the emotion side for the `keyframes` error. https://github.com/emotion-js/emotion/issues/2218

luisherranz:
> Filled an issue on the emotion side for the keyframes error. emotion-js/emotion#2218

Is this a blocker? If it is, can we solve it with our babel plugin until it is solved?

luisherranz:
> The error that the tests are reporting is a type mismatch/unrecognised issue not solved by the new pragma. Have you had these kind of issues before? 🤔

Not that I am aware of. Were you able to solve it?

cristianbote:
> > The error that the tests are reporting is a type mismatch/unrecognised issue not solved by the new pragma. Have you had these kind of issues before? 🤔
> 
> Not that I am aware of. Were you able to solve it?

Yup! Was the `jsxImportSource` option that I was not aware of it. 👍  All good now.

cristianbote:
> > Filled an issue on the emotion side for the keyframes error. [emotion-js/emotion#2218](https://github.com/emotion-js/emotion/issues/2218)
> 
> Is this a blocker? If it is, can we solve it with our babel plugin until it is solved?

Not a blocker in my opinion since the source maps are working for the `css` calls. The only result would be the `keyframes` import not being replaced with the `emotion` ones. But the `emotion` team seems open for external contributions on that one and I'm gonna try to add it, asap.

luisherranz:
Awesome, thanks for the answers Cristian.

This test seems to be failing: https://github.com/frontity/frontity/blob/chore/adopt-emotion11/e2e/integration/frontity-01/emotion.spec.js#L98-L104

Could you please take a look?

Once the PR is ready for review, we mark it as ready (I see you already did that 🙂) and assign the supporter 2 as the reviewer, so that person gets a notification. This week the supporter is @darerodz 👍

cristianbote:
@luisherranz @DAreRodz this is ready for review 🎉  I made peace with cypress after all 😄

cristianbote:
@luisherranz @DAreRodz this is ready for review 🎉  I made peace with cypress after all 😄

luisherranz:
> @luisherranz @DAreRodz this is ready for review 🎉 I made peace with cypress after all 😄

Haha, nice 😄

@DAreRodz should be the one reviewing this 🙂

packages/google-tag-manager-analytics/src/components/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -16,19 +16,16 @@ exports[`GoogleTagManager works with a single container id 1`] = `
...
-          visibility: hidden;
-        ",
-        "toString": [Function],
+        "__html": ".css-llz5b9-GtmCode{display:none;visibility:hidden;}",
       }
     }
+    data-emotion="css llz5b9-GtmCode"
+  />
+  <iframe
+    className="css-llz5b9-GtmCode"
```
DAreRodz, line 28: Any idea about why this component has a label in the `className`, even though the `autoLabel` option of `@emotion/babel-plugin` is set to `never` in the `jest.babel.config.json` file?

Just curious. :thinking:

DAreRodz (approved):
Awesome work, @cristianbote!
I liked the way to conditionally test the labels added by Emotion. :blush: 

Just a couple of questions/comments (see below). :point_down:

packages/google-tag-manager-analytics/src/components/__tests__/__snapshots__/index.tsx.snap
cristianbote: Thank you @DAreRodz 🙏  for the review.

> Any idea about why this component has a label in the className, even though the autoLabel option of @emotion/babel-plugin is set to never in the jest.babel.config.json file?

It's because of this issue: https://github.com/emotion-js/emotion/pull/2157

> In that case, I guess we would have to remove the "babel-plugin-frontity" from the other examples as well, right? Not only from the mars theme example.

Honestly I was about to, but then was talking with @luisherranz and if I remember correctly the examples will be moved or maintained by the devrel team. I could easily remove them now. Should I?

DAreRodz: > It's because of this issue: [emotion-js/emotion#2157](https://github.com/emotion-js/emotion/pull/2157)

Good to know. Thanks! :raised_hands: 

> I could easily remove them now. Should I?

Yes, please. It's very simple to fix and that way we don't have to remember to notify the devrel team about this.

cristianbote: Done ✅

examples/mars-theme-example/package.json
luisherranz: Yes, I don't know why it's there. It should only be a dependency of `@frontity/core`.

luisherranz: Ok, I see Cristian has already done so in https://github.com/frontity/frontity/pull/666/commits/979c44edcbe9b6615f1a74236bc2f350e9c7c0d4.

DAreRodz:
@cristianbote, feel free to merge this PR. :rocket:

PR #669, opened 2021-01-19
Link Component: Handling mailto:, sms: and tel: links

### LINKED ISSUE #667: The `Link` component tries to fetch links that start with `mailto:`
**_Bug report_**

**Expected behavior**

When the Link component is rendered with an `href` value that starts with `mailto:`, the browser should open a window to send an email if you click it and nothing else.

**Observed behavior**

If you set `state.theme.autoPrefetch` to `all`, `in-view` or `hover`, an error in the console will appear as soon as Frontity tries to fetch the email link ("no handler was found for the given link").  

**Possible solution**

The check to know if the link should be fetched or not is done in this line:

https://github.com/frontity/frontity/blob/94772c1217b4c48d67bcac742ecd9c93739c73ac/packages/components/link/index.tsx#L81-L83

We should change it to return also for links that start with `mailto`, `tel` and `sms`, e.g. changing the way `isExternal` is computed:

https://github.com/frontity/frontity/blob/94772c1217b4c48d67bcac742ecd9c93739c73ac/packages/components/link/index.tsx#L69

nicholasio:
**What**:

Fixes #667 

**Why**:

The Link component does not handle mailto:, SMS: and tel: links properly.

**How**:

Checking if the links starts with mailto, SMS and tel in a new `isExternalUrl` function.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages
- [ ]  End to end tests

**Additional Comments**


packages/components/link/utils.ts
```diff
@@ -137,3 +139,39 @@ export const removeSourceUrl = (
...
+ * @param link - The link Url.
+ *
+ * @returns True if the link is an external Url.
+ */
+export const isExternalUrl = (link: string) => {
+  return (
+    link.startsWith("http") ||
+    link.startsWith("mailto") ||
+    link.startsWith("tel") ||
+    link.startsWith("sms")
```
cristianbote: Hi @nicholasio thank you so much for taking the time and open this PR. Much appreciate it!

While looking at the changes was thinking if we maybe should approach it the other way around. Meaning, instead of checking that the url starts with a certain protocol, check if the url does not start with `/`?

luisherranz: That's a good idea. I remember that we used links like `whatsapp://send?text=xxx` for the social media apps.

But instead of checking for `/`, why don't we check that there is a protocol? That way it will work for relative links as well.

We could use either a regexp or `new URL` for that:

```js
const isExternal = (link) => {
  try {
    new URL(link);
    return true;
  } catch (e) {
    return false;
  }
};
```

```js
const isExternal = (link) => /\w+:\/\//.test(link);
```

cristianbote: Uh! @luisherranz that's so much better. Love it! 😃 I think the `URL` try/catch will do, but as you said either works. 👍

nicholasio: Definitely makes more sense to check this way, I'll get this updated 👍

nicholasio: Updated!

luisherranz:
@cristianbote can you review this one? 🙂

cristianbote (approved):
> @cristianbote can you review this one? 🙂

More than happy to! 👍  Great work @nicholasio! Thank you for your contribution. Looks great!

cristianbote:
@nicholasio was waiting for the eslint and task completed checker, but I believe these are fine. Gonna look into it. But it's all good 👍

cristianbote:
🎉  awesome job @nicholasio! Thank you!

PR #671, opened 2021-01-20
Enable chromeWebSecurity for e2e and adapt yoast tests to the change

cristianbote:
**What**:

Changed the testing strategy to check to response of the `cy.request` instead of relying on `cy.visit` with chromeWebSecurity disabled.

**Why**:

In order to reliable test the url navigations and redirections we need `chromeWebSecurity` enabled.

**How**:

Adapted the yoast e2e suite to assert the values of `cy.request` by parsing the response with `DOMParser` so the selectors could be reused.

**Tasks**:

- [x] TSDocs
- [x] End to end tests

**Unrelated Tasks**

- [ ] Code
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


DAreRodz (changes requested):
Tests are passing but I think there's a problem with the `async` functions, check it out. :point_down:

e2e/integration/wordpress-01/yoast-package.spec.js
```diff
@@ -13,14 +13,35 @@ describe("Yoast Package", () => {
...
-  const checkTitle = (title) => {
-    it("should render the correct title", () => {
-      cy.get("title").should("contain", title);
+  const checkTitle = (link, title) => {
+    /* eslint-disable-next-line cypress/no-async-tests */
+    it("should render the correct title", async () => {
+      const dom = await visitLinkAndParseDom(link);
+      const el = dom.querySelector("title");
+
+      cy.get(el).should("contain", title);
```
DAreRodz: If I change the value of `title` here to be a random string the test passes as well, so the `async` function is not working as expected.

By the way, I tried to follow the Cypress way and, rewriting the `checkTitle` function like this:

```ts
const checkTitle = (link, title) => {
  it("should render the correct title", () => {
    cy.request(`http://localhost:3001${link}?frontity_name=yoast-package`)
      .then((response) => parseHTML(response.body))
      .then((html) => {
        cy.get(html.querySelector("title")).should("contain", title);
      });
  });
};
```
then I get the following error:

Screenshot from 2021-01-20 13-34-25

cristianbote: wow! Such a great suggestion @DAreRodz! 🙇‍♂️  Will implement it and remove the `async` bits.

cristianbote:
Alrighty, so I believe now that the yoast package is behaving as it should? Right @DAreRodz? That leaves the rest of the tests that are doing SSR assertions and are doing it with disable JavaScript flag. Seems that we might need to generalise this method.

Thoughts?

cristianbote:
By the way, if I'm not gonna update any package, but rather only the tests, do I need to create a changeset?

cristianbote:
The tests will be solved by #672

cristianbote:
wohoo! everything passes 🎉

e2e/support/commands.js
```diff
@@ -48,3 +48,48 @@ Cypress.Commands.overwrite("visit", (visit, url, options = {}) => {
...
+  const parentDocument = cy.state("window").parent.document;
+  const iframe = parentDocument.querySelector(".iframes-container iframe");
+
+  apendChildrenTo(Array.from(doc.head.childNodes), iframe.contentDocument.head);
+  apendChildrenTo(Array.from(doc.body.childNodes), iframe.contentDocument.body);
+
+  return iframe.contentDocument;
+};
+
+Cypress.Commands.add("visitSSR", (url) => {
```
DAreRodz, line 93: I just miss a comment here. :eyes:

DAreRodz (approved):
Everything looks good to me! :+1:

e2e/support/commands.js
cristianbote: @DAreRodz I'll follow-up with a new PR to not block the 3xx redirections anymore.

PR #672, opened 2021-01-22
Add CORS header and Cache-Control to static serve

cristianbote:
**What**:

Add the CORS and Cache-Control headers to the static serve middleware.

**Why**:

In order to be able to load static resources across server domains.

**How**:

`koa-static` has this nifty method `setHeaders` which is designed specifically to add headers at will.

**Tasks**:

- [x] Code
- [x] Update other packages

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes

**Additional Comments**


cristianbote:
The external script test is failing and trying to zero in into it. I'll report back.

luisherranz (approved):
LGTM Cristian! 🙂

I've modified our eslint to avoid that eslint-disables and I've updated the script spec `expect` to use Jest instead of Mocha because we try to always use Jest's one, even in Cypress.

Feel free to merge.

e2e/integration/frontity-01/script.spec.js
```diff
@@ -20,7 +20,7 @@ describe("Script", () => {
       .its("moment")
       .should("exist")
       .then((moment) => {
-        expect(moment()._isAMomentObject).to.equal(true);
+        expect(moment()._isAMomentObject).toBe(true);
```
cristianbote, line 23: 🤔  huh, wondering why do you think we should use Jest's expect? `cypress` has the chai one already 🤷‍♂️

e2e/integration/frontity-01/script.spec.js
```diff
@@ -17,13 +16,11 @@ describe("Script", () => {
...
     cy.window()
-      .then((win) => {
-        isMoment = win.moment()._isAMomentObject;
-      })
-      .then(() => {
-        expect(isMoment).toBe(true);
+      .its("moment")
+      .should("exist")
+      .then((moment) => {
+        expect(moment()._isAMomentObject).toBe(true);
```
DAreRodz, line 23: I think this could be rewritten like this, right? Just to follow the Cypress way of do things and avoid dealing with jest eslint rules. :shrug: 

```ts
cy.window()
  .its("moment")
  .should("exist")
  .then((moment) => moment()._isAMomentObject)
  .should("be.true"); // or .should("equal", true)
```

e2e/integration/frontity-01/script.spec.js
DAreRodz: Wondering the same!

Just my two cents, @luisherranz: https://github.com/frontity/frontity/pull/672#discussion_r563604147

e2e/integration/frontity-01/script.spec.js
cristianbote: That's a good catch @DAreRodz. Like the way it integrates.

e2e/integration/frontity-01/script.spec.js
luisherranz: It was simply a personal preference, to learn the commands of only one `expect` instead of two.

DAreRodz: Gotcha!

PR #676, opened 2021-01-26
Amend getMatch to work with a source object instead

### LINKED ISSUE #596: `actions.source.fetch` fails with links with query params for certain handlers [5pt]

cristianbote:
**What**:

Change the way `getMatch` was handling the `link` to pattern match.

**Why**:

Matching was not happening for urls with querystring params.

**How**:

Different codepath for patterns that are defined using regexp, since the queryparams can only be parsed with regular expressions.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] TypeScript tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


packages/wp-source/src/libraries/get-match.ts
```diff
@@ -56,10 +56,30 @@ interface GetMatchReturn<
...
+ * The `source` object that contains a link and a route.
+ */
+interface Source {
+  /**
+   * The full link with `queryString`.
+   */
+  link?: string;
+
+  /**
+   * The link without the query part. Also known as the pathname.
```
luisherranz: There is no need to change the TSDocs because it's not a big deal, but just to make it clear, the route is not the pathname but the pathname without the pagination: 

- link: `/category/some-category/page/3?a=1`
- pathname: `/category/some-category/page/3`
- route: `/category/some-category/`

We didn't find a better name for that so we used `route` 😆🤷

cristianbote: Totally 😅 . We'll keep this in mind on next source version and we'll name it better.

luisherranz: > We'll keep this in mind on next source version and we'll name it better.

It's open to suggestions 👍

luisherranz (approved):
Great work, thanks Cristian! 👏🙂

PR #680, opened 2021-01-27
Move critical css to head

### LINKED ISSUE #559: Move all the CSS to the <head>. [5pt]

---

The Google AMP framework requires that all the CSS is placed in a `<style>` tag of the head.

For that reason, the `@frontity/amp` package needs to use Emotion's `extractCritical` API and move the extracted CSS to the `<head>`.

This can be done with a `beforeSSR` function. It'll be necessary to create a separate `server.js` file because we don't want to add `create-emotion-server` and `react-dom/server` to the client bundle.

```js
// packages/amp/src/server.js
import { CacheProvider } from "@emotion/core";
import createEmotionServer from "create-emotion-server";
import createCache from "@emotion/cache";

export default {
  actions: {
    amp: {
        // Create an Emotion instance for this SSR.
        const cache = createCache();
        const { extractCritical } = createEmotionServer(cache);

        libraries.frontity.render = (App) => {
          // Add the Provider to the App.
          App = (
            <CacheProvider value={cache}>
              <App />
            </CacheProvider>
          );

          // Run extract critical.
          const { html, css, ids } = extractCritical(
            libraries.frontity.render(App)
          );

          // Replace the template injecting the CSS in the `<head>`.
          libraries.frontity.template = ampTemplate(css, ids);

          return html;
        };
      },
    },
  },
};
```

We also need to be able to add the `CacheProvider` to the client. That can be done with a `beforeCSR` function.

```js
// packages/amp/src/client.js
import createCache from "@emotion/cache";

export default {
  actions: {
    amp: {
      beforeCSR: ({ libraries }) => {
        const cache = createCache();

        libraries.frontity.App = () => (
          <CacheProvider value={cache}>
            <libraries.frontity.App />
          </CacheProvider>
        );
      },
    },
  },
};
```

For context, this is the way Emotion worked in Frontity at first. This is the final commit, including a fix for the `Global` component: https://github.com/frontity/frontity/commit/e8c3430

After that, we removed `extractCritical` because both labels and source maps don't work with that approach. Now it's up to the `@frontity/amp` package add `extractCritical` again for AMP sites. This is where I removed that approach: https://github.com/frontity/frontity/commit/d5af653

Looking at the new docs it seems to me that we don't need to do a separate `hydrate()` on the ids anymore, as they are present in the `data-emotion-css` attribute of the `style` tag:

```js
<style data-emotion-css="${ids.join(' ')}">${css}</style>
```

More info about the `extractCritical` approach on https://emotion.sh/docs/ssr#advanced-approach

cristianbote:
**What**:

Move the critical CSS from emotion to the head section.

**Why**:

In order to build a AMP package this needs to happen.

**How**:

Using emotions server packages and cache. On the server the cache is generated per request, to capture all `Global` calls, and on the client is hydrated via the same cache API.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] End to end tests
- [x] Update other packages

**Unrelated Tasks**

- [ ] TSDocs
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes

**Additional Comments**

This relates to #559.


cristianbote:
Ran the two failing tests locally with the same conditions and they pass. Gonna re-rerun the e2e action.

cristianbote:
Yeah, everything passes, glitches do happen 👍

cristianbote:
I tried to see if this actually will concretely improve our performance metrics studying the Lighthouse results and the Performance Tab, specifically the _hydration_ part.

## Conclusion

There's no objective marker that stand to support that we'll be gaining any better metrics if we move the CSS in the head. I thought this should be the case, but, the main thread is so busy with the hydration that even if there is an improvement it's not clearly visible.

luisherranz:
Ok, thanks for testing it out Cristian 🙂

By the way next week we should agree on how to do this measurements, which we are going to need for the rest of the quarter. I'd like to end the quarter with a table of the things we tried and the metrics they improved (or not).

luisherranz:
I removed myself from the reviewers, add me once this is ready for review again.

cristianbote:
Will do @luisherranz!

Based on what we've discussed yesterday about the AMP package I'm gonna close this PR and the changes needed to move the CSS to head are gonna be part of the package `beforeSSR` hook. As it was initially designed.

cristianbote:
This is gonna be implemented as part of the `amp` package instead. Closing.

PR #686, opened 2021-01-30
Remove `/amp` from the internal source actions

### LINKED ISSUE #586: Remove /amp from link in actions.source.fetch and state.source.get [3pt]

---

There are three standard ways to use AMP. This package needs to support all of them.

1. Using a subdomain, like `https://amp.mydomain.com/some-post`.
2. Using a query parameter, like `https://www.mydomain.com/some-post?amp=true`.
3. Using an extra folder in the path, like `https://www.mydomain.com/some-post/amp`.

1 and 2 are fine because the slug used by `@frontity/wp-source` will be `some-post`, but 3 is not.

We need to make this action and derived state point to `/some-post/`:

```js
actions.source.fecth("/some-post/amp");
state.source.get("/some-post/amp");
```

Ideally, this should be done with hooks, but we don't have hooks yet and they seem too much to make them a dependency of this feature right now.

We've been searching for a workaround but the only thing that seems feasible right now is to hardcode a replacement in the normalize function and change it once we have hooks.

```js
export const normalize = (route: string): string => {
  route = route.replace(/\/amp$/, "");
  return paramsToRoute(routeToParams(route));
};
```

This will need to contain a bit more logic to avoid some known URLs that shouldn't be changed, like:

- A page called amp: `/amp`.
- A category/tag called amp: `/category/amp` or `/tag/amp`.

### LINKED ISSUE #685: Create the `@frontity/amp` package
Just the boilerplate for the AMP package in our monorepo, with a basic index.js, types.js, package.json and so on...

michalczaplinski:
**What**:

Remove the `/amp` suffix from the internal source actions.

**Why**:

Because one way to use AMP with WordPress is to add an extra `/amp` folder in the path, like: https://www.mydomain.com/some-post/amp which serves the AMP pages. Frontity should ignore that suffix in its internal links.

**How**:

Creating a new `@frontity/amp` package. Inside of the package, we override `libraries.source.normalize()`, `libraries.source.parse` and `libraries.source.stringify()`. in the init() action of the package.

Here's a video summary of the changes: https://www.loom.com/share/e939d2fa55924c3a91caac9585a023a9

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [x] End to end tests
- [x] TypeScript tests
- [x] Update starter themes
- [x] Update other packages


luisherranz:
Michal, could you please test if we can do this with the internal wp-source redirections?

We talked briefly about it in this video: https://community.frontity.org/t/amp-package/388/17?u=luisherranz

Thanks! :)

michalczaplinski:
@luisherranz Right, will do. I saw it in the video, too 👍

luisherranz:
Awesome, thanks Michal! 🙂

michalczaplinski:
I've tried and failed with some approaches based on internal frontity redirections.

The redirection should simply look something like this: 

```js
// The redirection for AMP
redirections.push({
  name: "amp",
  priority: 1,
  pattern: ":everything(.*)/amp",
  func: ({ everything }) => everything,
});
```

However, since we want to NOT match links like `/category/amp` or `/tag/amp`, we would have to add [negative lookahead](https://stackoverflow.com/questions/1240275/how-to-negate-specific-word-in-regex) to the regex in `pattern` which makes the regex not very readable...

I think that the best solution is still to hardcode a replacement in the normalize function like mentioned in the [Feature Discussion](https://community.frontity.org/t/amp-package/388/10). Something like:

```ts
export const normalize = (link: string): string => {
  const regex = pathToRegexp("/(category|tag|author)/amp");

  if (!link.match(regex)) {
    link = link.replace(/\/amp$/, "");
  }
  return paramsToLink(linkToParams(link));
};
```

However, this is not going currently not working with Custom Post Types. If a user has a CPT called `amp`, the regex above will **not** match and e.g. a link like `/cpt/amp` will be replaced by `"/cpt/"`.

luisherranz:
@michalczaplinski could you please add a small update here with the final decision? Thanks :)

michalczaplinski:
We've decided to go with the "redirections" approach at the moment 🙂  Here's the reasoning behind the decision.


1. Luis has pointed out that even though the "pattern" in the redirection can only match a regex or a `path-to-regexp`-style pattern, we could do further filtering inside of the body of the `func` like: 

```

// The redirection for AMP
redirections.push({
  name: "amp",
  priority: 1,
  pattern: ":everything(.*)/amp",
  func: ({ everything }) => {
	// filter stuff further.
  },
});
```

2. The "redirections" approach is more portable and encapsulates the change better than hardcoding it inside of the `normalize()`. This way we could even move the code out of wp-source in the future. Which brings me to 3.

3. I think that it makes slightly more sense to add this change inside of the `amp` package rather than to implement it in wp-source. This way the amp-specific functionality is more clearly encapsulated and the final bundle ends up slightly smaller. Of course, the `amp` package is meant to be used together with `wp-source` so both are still needed for this to work.

4. The issue with CPT that I mentioned earlier can be solved by checking if there is a custom post type called `amp` in the `state.source.postTypes` and then if yes, checking that the route matches the post "pattern".

michalczaplinski:
I've realized that we also have to add a tiny fix to how the redirections work.

At the moment, the only thing that the internal redirections do is update the `route` inside of the `actions.source.fetch()` when fetching data for some link: 

https://github.com/frontity/frontity/blob/c139094e956d3119f9fd1ea2342717e29e187f2f/packages/wp-source/src/actions.ts#L67-L69

However, when matching the correct handler inside of the `getMatch()` function we are currently using BOTH `route` and `link`:

https://github.com/frontity/frontity/blob/42eec53f693aeef19df1c93fb94486e52ecc6920/packages/wp-source/src/libraries/get-match.ts#L101-L106

~~So, the `func` passed to the redirection object should return a `{link, route}` tuple, so that the `/amp` link can be stripped from both.~~

So, one solution for this is to [pass the "updated" link to `getMatch()`](https://github.com/frontity/frontity/pull/686/commits/d0bbabf0b279b0889155144c4e2159be8cd8b2f0) that is re-derived from the `route` if there is a redirection.

luisherranz:
@michalczaplinski, I've just realized that differentiating between AMP/not-AMP paths is not needed at all.

The way the `@frontity/amp` package is going to work is with a new site:

```js
export default [
  {
    name: "normal-site",
    packages: [
      // Normal packages.
    ],
  },
  {
    name: "amp-site",
    match: "\\/amp\\/?($|\\?|#)"
    packages: [
      "@frontity/amp",
      // Normal packages.
    ],
  },
];
```

It is going to be done that way because an optimized site should not have additional AMP code added to its bundle. With different sites, we can create independent bundles.

The way to load the AMP site is with a `match`. The usual configurations are:

- An `amp.domain.com` subdomain.
- An `?amp=true` query.
- An ending `/amp` path.

As you can see, the `/amp` path will always load the AMP site, so `/category/amp` will always match.

This means that the `/amp` option doesn't support using `amp` for a term or post slug. When people need that, they have to either:

- Choose another option (subdomain or query).
- Add it manually to the `match` regexp.

As we discovered yesterday that we cannot use redirections, I think adding this to the normalize should be enough:

```js
link = link.replace(/\/amp\/?($|\?|#)/, "");
```

I will update the AMP FD as well to make sure that we don't forget to add a warning about this limitation of the `/amp` option in the documentation.

michalczaplinski:
I have realized another problem in the implementation using `normalize()`, unfortunately 🙁 

### TL;DR

This is the failing test case: 

https://github.com/frontity/frontity/blob/86ba9e13b9ef5078e8ae561265445cc18bc5bb9e/packages/wp-source/src/__tests__/amp.tests.ts#L96-L118

And it fails on this line:
https://github.com/frontity/frontity/blob/86ba9e13b9ef5078e8ae561265445cc18bc5bb9e/packages/wp-source/src/__tests__/amp.tests.ts#L114

### The problem

If we change the normalize function to this:

https://github.com/frontity/frontity/blob/86ba9e13b9ef5078e8ae561265445cc18bc5bb9e/packages/wp-source/src/libraries/route-utils.ts#L194-L197

which strips out the `/amp` in the link, we will not be able to get the correct matches for the handlers in `getMatch()`.

The `getMatch()` is actually usually using the `route` and not the `link` to do the matching unless the pattern is a RegExp: 

https://github.com/frontity/frontity/blob/86ba9e13b9ef5078e8ae561265445cc18bc5bb9e/packages/wp-source/src/libraries/get-match.ts#L101-L106

And because `normalize()` does not update the `route`, **only the `link`**, `getMatch()` matches the `postType` handler instead of the `postArchive` handler when fetching `"/amp"`

### Solution?

I can think of 2 ways to finally handle this:

1. Strip out the `/amp` inside of `getMatch()` function itself. It has the same disadvantage as using internal redirections: the change is hidden in frontity internal code. I have tried it without taking a deep look but it breaks a number of other tests in wp-source so I'm pretty sure this is not the way to go.

2. Hardcode the removal of `"/amp"` inside of `actions.source.fetch()` and `state.source.get()`. This way, we can remove the `/amp` from both `route` and `link`. Not the cleanest solution, of course.

@luisherranz I'd really like some of your input here 🙂

luisherranz:
> Hardcode the removal of "/amp" inside of actions.source.fetch() and state.source.get(). This way, we can remove the /amp from both route and link. Not the cleanest solution, of course.

Yes, let's try modifying the `link` inside the normalize function, but in both `parse` and `stringify` to make sure is consistent and we don't have to hardcode it in the actions. I've tried it out and it seems to work fine.

There is one problem though, we can't add it to wp-source or we will break paths like `/category/amp` for normal sites. We need to add it in the `@frontity/amp` package. We can use the wrapper pattern for that. The only problem is that we need to also fix `normalize`.

```js
/**
 * Remove a final `/amp` from a link.
 *
 * @param link - The link to be processed.
 *
 * @returns The same link but without the final `/amp`.
 */
export const removeAmp = (link: string): string =>
  link.replace(/\/amp\/?($|\?|#)/, "/");

/**
 * ...
 */
export default {
  actions: {
    amp: {
      init: ({ libraries }) => {
        // Wrap libraries.source.parse.
        const parse = libraries.source.parse;
        libraries.source.parse = (link) => {
          const { route, path, ...rest } = parse(link);
          return {
            route: removeAmp(route),
            path: removeAmp(path),
            ...rest,
          };
        };

        // Wrap libraries.source.stringify.
        const stringify = libraries.source.stringify;
        libraries.source.stringify = ({ route, path, ...rest }) => {
          const routeOrPath = route || path || "/";
          return stringify({
            route: removeAmp(routeOrPath),
            path: removeAmp(routeOrPath),
            ...rest,
          });
        };

        // Wrap libraries.source.normalize.
        const normalize = libraries.source.normalize;
        libraries.source.normalize = (link) =>
          libraries.source.parse(libraries.source.stringify(link));
      },
    },
  },
};
```

_We can assume that the `@frontity/amp` package will run before `@frontity/tiny-router`._

I made a video to explain it better: https://www.loom.com/share/8fd2ace832f74eb7afeb02eb5ee476d8

I said in the video that we should close this PR, but it doesn't make sense. We can continue the work here once we have a `@frontity/amp` package (https://github.com/frontity/frontity/issues/685).

This is an example of the type of tests we could add to the `@frontity/amp` package. Maybe it would be nice to have more for `parse` and `stringify`:

```js
test("final `/amp` path is removed", () => {
  expect(normalize("/amp")).toBe("/");
  expect(normalize("/some-post/amp/")).toBe("/some-post/");
  expect(normalize("/some-page/amp/")).toBe("/some-page/");
  expect(normalize("/parent/some-page/amp/")).toBe("/parent/some-page/");
  expect(normalize("/category/some/amp/")).toBe("/category/some/");
  expect(normalize("/tag/some/amp/")).toBe("/tag/some/");
  expect(normalize("/taxonomy/some/amp/")).toBe("/taxonomy/some/");

  expect(normalize("/amp/?a=1&b=1#some-hash")).toBe("/?a=1&b=1#some-hash");
  expect(normalize("/amp/page/3/?a=1&b=1#some-hash")).toBe(
    "/page/3/?a=1&b=1#some-hash"
  );
  expect(normalize("/taxonomy/some/amp/?a=1&b=1#some-hash")).toBe(
    "/taxonomy/some/?a=1&b=1#some-hash"
  );
  expect(normalize("/taxonomy/some/page/3/?a=1&b=1#some-hash")).toBe(
    "/taxonomy/some/page/3/?a=1&b=1#some-hash"
  );

  // These ones are not correct, but it is expected.
  expect(normalize("/category/amp/")).toBe("/category/");
  expect(normalize("/tag/amp/")).toBe("/tag/");
  expect(normalize("/taxonomy/amp/")).toBe("/taxonomy/");
  expect(normalize("/author/amp/")).toBe("/author/");
});
```

luisherranz:
Hey Michal, I've made a couple of videos to explain why my proposed change fixes the problem you explain in https://github.com/frontity/frontity/pull/686#issuecomment-774807432:

- https://www.loom.com/share/f1453dd449194777876521dfbcbfe6cc
- https://www.loom.com/share/a71ef92ca1284c1a80336f70df9e5be0

---

And I think we should also change how `actions.source.fetch` imports `normalize` and `parse`:

```js
import { parse, normalize } from "./libraries/route-utils";

const actions = {
  fetch: ({ state, libraries }) => async (link) => {
    // Get route and route params.
    link = normalize(link);
    const linkParams = parse(route);
    // ...
  },
};
```

We should do this instead:

```js
const actions = {
  fetch: ({ state, libraries }) => async (link) => {
    // Get route and route params.
    link = libraries.source.normalize(link);
    const linkParams = libraries.source.parse(route);
    // ...
  },
};
```

--- 

Like I said in the previous video, I don't know if this is going to work or not but at least the `fetch` will get the correct data. 

I think we will find out if making `state.router.link = "/some-post/"` instead of `/some-post/amp` is the best approach once we have a real AMP application.

packages/amp/src/__tests__/amp.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  test("Compare the output of libraries.source.parse() between the AMP package and the default implementation", () => {
+    // Get the reference to parse() before it's overriden in the init() action.
+    const { parse } = store.libraries.source;
+    store.actions.amp.init();
+
+    // The implementation of parse() from the AMP package.
+    const { parse: parseAmp } = store.libraries.source;
+
+    expect(parse("/")).toEqual(parseAmp("/"));
```
cristianbote, line 99: Hey @michalczaplinski, do you think we should assert that `parse !== parseAmp`? I am thinking that we should make sure that after init with amp package that parse it's not the same anymore. What are your thoughts?

michalczaplinski: Good point, @cristianbote :) I will add that change!

michalczaplinski: ok, done in https://github.com/frontity/frontity/commit/5c66bb90d9b9feb72e4a0384b1185b8b986b1f4b

.changeset/hungry-oranges-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontity/amp": patch
```
luisherranz, line 2: Yeah, I think this is fine. Either a `patch` or a `minor`, it's not very important.

@santosguillamot, we are going to release the `@frontity/amp` package in "beta" first, right? So `0.0.1` is fine I mean.

packages/amp/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+You can should install the package and add it to your `frontity.settings.js` file:
+
+```js
+export default {
+  packages: [
+    "@frontity/amp",
+    // ...your other packages
+  ],
+};
```
luisherranz: Maybe we can say here a bit more about how to normally set up two sites, one for React and another for AMP.

```suggestion
You should add the package to your `frontity.settings.js` file.

Usually, you will want to have a "main site" which doesn't use this package and renders the normal React application and an "amp site" that returns the AMP version of the pages.

The AMP site should have a `match` property which will indicate which are the AMP URLs. Some examples are:

- A subdomain: `"match": "amp\\.domain\\.com"`
- A query: `"match": "(\\?|&)amp=true"`
- An ending path: `"match": "\\/amp\\/?($|\\?|#)"`

You can read more about setting up multiple sites [in our docs](https://docs.frontity.org/learning-frontity/settings#multiple-sites).

```js
export default [
  {
    name: "main-site",
    packages: [
      // ...your normal packages
    ],
  },
  {
    name: "amp-site",
    match: "...",
    packages: [
      "@frontity/amp",
      // ...your other packages
    ],
  },
];
```

packages/amp/types.ts
```diff
@@ -0,0 +1,38 @@
+import { Package, Action, MergePackages } from "frontity/types";
+import WpSource from "@frontity/wp-source/types";
```
luisherranz:
```suggestion
import Source from "@frontity/source/types";
```

packages/amp/types.ts
```diff
@@ -0,0 +1,38 @@
...
+       */
+      init: Action<Packages>;
+    };
+  };
+}
+
+/**
+ * Packages used internally by the AMP package.
+ */
+export type Packages = MergePackages<AMP, WpSource>;
```
luisherranz:
```suggestion
export type Packages = MergePackages<AMP, Source>;
```

packages/amp/src/__tests__/amp.test.ts
```diff
@@ -0,0 +1,193 @@
...
+import WpSource from "@frontity/wp-source/types";
+import AMP from "../../types";
+import { MergePackages } from "frontity/types";
+import Amp from "..";
+
+let store: InitializedStore<MergePackages<WpSource, AMP>>;
+describe("AMP tests", () => {
+  beforeEach(() => {
+    // Initialize the store
+    store = createStore<MergePackages<WpSource, AMP>>(
```
luisherranz: You can use the `Packages` type here.

```suggestion
import AMP, { Packages } from "../../types";
import Amp from "..";

let store: InitializedStore<Packages>;
describe("AMP tests", () => {
  beforeEach(() => {
    // Initialize the store
    store = createStore<Packages>(
```

packages/amp/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "test:inspect": "node --inspect ../../node_modules/.bin/jest --watch --no-cache --runInBand"
+  },
+  "dependencies": {
+    "@frontity/wp-source": "^1.11.1",
+    "frontity": "^1.14.1"
+  },
+  "devDependencies": {
+    "@frontity/connect": "^1.2.1",
+    "clone-deep": "^4.0.1"
+  }
```
luisherranz:
```suggestion
  "dependencies": {
    "@frontity/source": "^1.5.1",
    "frontity": "^1.14.1"
  },
  "devDependencies": {
    "@frontity/connect": "^1.2.1",
    "@frontity/wp-source": "^1.11.1",
    "clone-deep": "^4.0.1"
  }
```

We should use `source` for the types, not `wp-source`. Source types have `parse`, `stringify` and `normalize`: 
 https://github.com/frontity/frontity/blob/dev/packages/source/types/index.ts#L285-L315

luisherranz (approved):
You make a good point about `state.source.get()`. I don't think it's going to be a problem because a link ending in `/amp` will never get to `state.router.link` because it's going to be changed in the `actions.router.init()` function: https://github.com/frontity/frontity/blob/dev/packages/tiny-router/src/actions.ts#L114-L119

So things like `actions.source.fetch()` or `state.source.get()` won't have to deal with those links.

But it's a good point. I think the extensibility we are using here is not good enough. But we already have new plans for this for Source v2 using the upcoming Frontity Hooks/Filters so I am not that worried about it.

I've added a few comments here and there with suggestions that I think you should be able to merge directly (let's see how well that works) but otherwise this is approved 🙂 Thanks for the work Michal.

Let's see if this finally works out fine once we have other pieces of the AMP package working!

.changeset/hungry-oranges-kneel.md
SantosGuillamot: Yes. The idea was to release a beta with the first functionalities and start testing it.

packages/amp/package.json
michalczaplinski: oh, I did't realize this. I thought that they were specific to wp-source :)

michalczaplinski:
Thanks for the review @luisherranz. 

I've added two tiny changes on top of your suggestions, so I ll ask you to review quickly again :) 

- [`77ea55d` (#686)](https://github.com/frontity/frontity/pull/686/commits/77ea55d5e030ce327a4f4a15833e30e97e02ff40)
- [`60a8d47` (#686)](https://github.com/frontity/frontity/pull/686/commits/60a8d476ee637e560477a8e54a72677dbfb3c225)

luisherranz:
> I've added two tiny changes on top of your suggestions, so I ll ask you to review quickly again :)

Oh, no need for those types of small changes, don't worry. You can merge it 🙂👍

PR #693, opened 2021-02-05
Change core-js polyfill strategy to entry

cristianbote:
While exploring the reduction of generated output I've stumbled on [this section on the core-js repo](https://github.com/zloirock/core-js#babelpreset-env). I've quickly tested it out in our codebase and indeed, changing the configuration as noted there it's generating less code. And after careful reading it makes sense.

**What**:

Changing the core-js polyfill strategy to `entry` will generate less code since it's gonna generate only one _entry_ block of polyfills. `usage` would colocate and generate and import per scope/context. So if two files used one feature that needs polyfill the import would be defined twice, in each file.

**Why**:

Less code will be generate, which means faster parsing and less time spending on evaluating the code at load time.

**Tasks**:

- [x] TypeScript
- [x] Unit tests
- [x] Update other packages

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes

**Additional Comments**

On chrome I've notice around `-80kB` shavings and around `-350kB` on Windows on Edge 😲  -- via VirtualBox.

And also the performance score on Lighthouse increased with `+9 points`.

### DIFF
--- a/packages/core/src/config/babel/index.ts
+++ b/packages/core/src/config/babel/index.ts
@@ -60,8 +60,8 @@ const getConfig = (target: Target): TransformOptions => {
       "@babel/preset-env",
       {
         targets: targets[target],
-        useBuiltIns: "usage",
-        corejs: "3",
+        useBuiltIns: "entry",
+        corejs: "3.8",
         modules: false,
         bugfixes: true,
       },
--- a/.changeset/late-phones-sniff.md
+++ b/.changeset/late-phones-sniff.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Change the core-js polyfill strategy to generate less code.
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,6 @@
 			"version": "7.5.5",
 			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
 			"integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
-			"dev": true,
 			"requires": {
 				"@babel/highlight": "^7.0.0"
 			}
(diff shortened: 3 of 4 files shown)


packages/core/src/config/babel/index.ts
```diff
@@ -60,8 +60,8 @@ const getConfig = (target: Target): TransformOptions => {
       "@babel/preset-env",
       {
         targets: targets[target],
-        useBuiltIns: "usage",
-        corejs: "3",
+        useBuiltIns: "entry",
+        corejs: "3.8",
```
luisherranz, line 64: Why `3.8` instead of `3`? Just curious.

luisherranz (changes requested):
Funny thing, I changed it because I thought that `usage` would end up in less code.

---

I don't see the required imports for the `entry` option: 

```js
import "core-js"; 
import "regenerator-runtime/runtime";
```

You can add them in the webpack entry, similar to this, but for both `es5` and `module`: https://github.com/frontity/frontity/commit/9ee75fb263984afb3bb45aa97f2cd55b4571d69f

_By the way, before `babel-preset-env` 7.4 the import was `@babel/polyfill`. It is explained [here](https://babeljs.io/docs/en/babel-preset-env#usebuiltins-entry) in the history section._

---

Could you please also repeat your test once you add those? Maybe your previous test resulted in a smaller file because without those imports no polyfills are added so we better be sure. Thanks! 🙂

packages/core/src/config/babel/index.ts
cristianbote: That's what the docs say 🙂, to use the minor version instead of the major.

luisherranz: Ok then 🙂

cristianbote:
You are totally correct @luisherranz! Adding `core-js/stable` and `runtime-regenerator` -- the last one is only 2.5kB peanuts 😄 , bumps the previous size to 440kB 😬  yikes!

My drive to look into this one, was actually grasping over the coverage reporter in chrome and I was amazed that 56.3% `core-js` it is not used 😲 .

Screenshot 2021-02-10 at 22 46 44

Gonna close this as obviously I failed to take into account the `core-js` dependencies.

luisherranz:
> My drive to look into this one, was actually grasping over the coverage reporter in chrome and I was amazed that 56.3% core-js it is not used

That is a shame.

The only approach that I know of that tries to solve this problem is poyfill.io, a server that inspects the User-Agent of the request and returns only the polyfills required.

The problem with that approach is that it adds a blocking request, even for browsers that don't require any polyfill, so it may add to the TTI time.

But maybe we can explore a `@frontity/polyfill.io` package that does that.

PR #717, opened 2021-02-10
Update the value of `link` in actions.fetch if there's an internal redirection.

michalczaplinski:
**What**:

Update the value of `link` in actions.fetch if there's an internal redirection.

**Why**:

If there is an internal redirection, it updates the value of `route` `in actions.source.fetch()`.
However, the `getMatch()` function uses both `route` and `link` to match a correct handler.

Therefore, if a redirection returns a "new" route, we have to update the value of `link` based on that new route.

**How**:

Update `actions.source.fetch()`.

**Tasks**:

- [x] Code

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


packages/wp-source/src/actions.ts
```diff
@@ -65,8 +65,22 @@ const actions: WpSource["actions"]["source"] = {
...
+      // pass to the `getMatch()` function. If there is a redirection, we update
+      // the value of `route`, so we'll have to later derive the link again
+      // based on the "new" (redirected) route.
+      let linkForHandler = link;
+      if (redirection) {
+        route = redirection.func(redirection.params);
+
+        // Derive the link from the "redirected" route.
+        linkForHandler = normalize(route);
+      }
```
cristianbote: Is there a case were we could capture this within a test? 🤔

DAreRodz: > Is there a case were we could capture this within a test? :thinking:

Yeah, I guess so. In the `3xx-redirections.tests.ts` file, there is a `handler.func` mock that we can use to check what params were passed, and if `link` has the correct value.

@michalczaplinski, could you check the `link` value in those tests?

DAreRodz (changes requested):
Hey, @michalczaplinski, could you add some tests to this patch?

I think it would be simple, just check that `handler.func` was called with the correct `link` param, inside the `3xx-redirections.tests.ts` file.

packages/wp-source/src/actions.ts
michalczaplinski: yup, I've added a test case. Thanks for pointing it out @DAreRodz and @cristianbote 🙂 

I explain in a new video why I did it this way: https://www.loom.com/share/4e265877436547848d6df9cd32cddd18

DAreRodz: Oops, my bad, hahaha :see_no_evil: You're completely right. I don't know why I thought about 3xx redirections. :sweat_smile:

DAreRodz: Wait a second.

If this is about internal redirections, then the value of `link` should not be changed! :worried: Or should it be? I'm confused now. :dizzy_face:

luisherranz: I think Michal is right, the redirections only change `route` so they are not useful for handlers with RegExps.

For example, imagine there is

- A redirection of the posts archive from `/blog/` to `/`.
- A handler that uses this `RegExp:\\/?home_option=.*` which should match the posts archive with the `home_option` query.

The handler won't capture the link `/blog/?home_option=xxx` because only the `route` is changed and the RegExp is using `link` inside `getMatch()`.

What I don't know is if the proposed solution is enough to fix the problem. If I am not mistaken, some of the redirections are removing the queries.

```js
if (postsPage) {
  const pattern = concatLink(subdirectory, postsPage);
  redirections.push({
    name: "posts page",
    priority: 10,
    pattern,
    func: () => "/",
  });
}
```

Those were not important in the `path-to-regexp` patterns, but they are now in the RegExps. So maybe to solve this we also need to take a look at the redirections.

What do you think guys?

DAreRodz: Aaah, gotcha! Sorry again, I missunderstood the code. :see_no_evil: 

I thought the `link` property was passed later to the handler function, but it's only passed to the `getMatch()` call.

michalczaplinski:
I've updated the implementation slightly. Now, it's using:

```js
linkForHandler = addFinalSlash(stringify({ route, page, query, hash }));
```

instead of 

```js 
// yeah, this didn't make sense lol 😝 
linkForHandler = normalize(route);
```

I think it should solve [the problem](https://github.com/frontity/frontity/pull/717#discussion_r576006122) that Luis mentions. I've also added a test case that failed with the previous approach but passes now.

I've mentioned in person that I might have to change the redirections themselves, but after taking another look this morning this doesn't seem to be the case to me.

---

I just wanted to make one final improvement if I can: 

@DAreRodz I see that `libraries.source.parse()` is simply calling `linkToParams` and conversely `stringify()` is calling `paramsToLink`. Could we just get rid of `linkToParams` and `paramsToLink` and use their implementations directly? As far as I can tell paramsToLink and linkToParams are only used internally.

EDIT: After discussing this with David in person, indeed the `paramsToLink` and `linkToParams` can be removed.

DAreRodz (approved):
I reviewed it again, and everything is great. :100:

PR #720, opened 2021-02-15
Extract critical css to head for amp package

### LINKED ISSUE #559: Move all the CSS to the <head>. [5pt]

cristianbote:
**What**:

Move the critical css to head to pave the way for the amp package.
**Why**:

Amp needs to have the css extracted in the head.
**How**:

Using emotion's server package and the nifty render, template and App apis.
**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Update other packages
- [x] Unit tests
- [x] TypeScript tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] Update starter themes

**Additional Comments**


packages/amp/src/server.tsx
```diff
@@ -0,0 +1,56 @@
+import { CacheProvider } from "frontity";
+import createEmotionServer from "@emotion/server/create-instance";
+import createCache from "@emotion/cache";
+import config, { CACHE_KEY } from "./config";
+
+export default {
```
michalczaplinski: Could we add the type here?

```js
const ampServer: AMP = { ... };

export default ampServer;
```

packages/amp/src/client.tsx
```diff
@@ -0,0 +1,26 @@
+import { CacheProvider } from "frontity";
+import createCache from "@emotion/cache";
+import config, { CACHE_KEY } from "./config";
+
+export default {
```
michalczaplinski: Could we add the type here?

```js
const ampClient: AMP = { ... };

export default ampClient;
```

michalczaplinski (changes requested):
Hey @cristianbote I think the code is great! 👍 

I've only suggested some minor changes, mostly to the tests. Just FYI, I guess it's better if you add the "suggested" changes locally, because if you merge them on github, e.g. the snapshots for the tests won't be generated.

I've also noticed that the types in `@frontity/types/src/frontity` should contain the types for `libraries.frontity.App`, `template` & `render`. Could you add them as well? Here's what I mean: 

https://user-images.githubusercontent.com/5417266/108001674-195c9f00-6fbb-11eb-9a5d-f4f03dcad401.mov

Then, you could change the type of `Packages` in https://github.com/frontity/frontity/blob/amp-package-move-the-css-to-head/packages/amp/types.ts#L55 to include those updated Frontity types like

```ts
export type Packages = MergePackages<AMP, Source, Frontity>;
```

packages/amp/src/__tests__/amp.client.test.ts
```diff
@@ -0,0 +1,33 @@
...
+      frontity: {
+        App: jest.fn(),
+      },
+    };
+  });
+
+  test("The initial App is overwritten with the Amp one", () => {
+    const previousApp = store.libraries.frontity.App;
+    store.actions.amp.beforeCSR();
+
```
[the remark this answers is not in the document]
cristianbote: You are correct, I could do better here tests wise. At least making sure the output is gonna stay the same. 👍

packages/amp/src/__tests__/amp.server.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      },
+    };
+  });
+
+  test("The initial values are overwritten with the Amp ones", () => {
+    const previousRender = store.libraries.frontity.render;
+    const previousTemplate = store.libraries.frontity.template;
+
+    store.actions.amp.beforeSSR();
+
```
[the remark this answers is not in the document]
cristianbote: Indeed we can! Will update the tests. Thank you!

packages/amp/src/server.tsx
cristianbote: oh, oops! Yeah 😅

packages/amp/src/client.tsx
cristianbote: Yuppers!

cristianbote:
Hey @michalczaplinski! 😃  Thank you so much for your review! Indeed, the types for `libraries.frontity` are not good enough. Also gonna update the tests to include their output.

cristianbote:
@michalczaplinski did the changes you suggested. Let me know what you think. Thank you!

michalczaplinski:
This is just a comment that I forgot to mention earlier: 

I think that we could add the e2e tests using the `amphtml-validator` later on when we have the rest of the missing pieces of the AMP package. I have the boilerplate for the e2e-amp package ready in my PR: https://github.com/frontity/frontity/pull/718

projects/mars-theme/frontity.settings.js
```diff
@@ -8,6 +8,7 @@ const settings = {
     },
   },
   packages: [
+    "@frontity/amp",
```
michalczaplinski: Hmm, do you want to add the AMP package for some specific reason to mars-theme or was it just something you did for testing?

projects/mars-theme/package.json
```diff
@@ -20,6 +20,7 @@
     "@frontity/mars-theme": "^1.3.1",
     "@frontity/tiny-router": "^1.1.0",
     "@frontity/wp-source": "^1.5.0",
+    "@frontity/amp": "0.0.1",
```
michalczaplinski: Same as above: was it just for testing or did you want to add amp to mars-theme for some specific reason? :)

michalczaplinski (changes requested):
Okay, everything looks great @cristianbote sans those two changes in mars-theme. Just let me know - I've got my finger ready on the "approve" button 😄

projects/mars-theme/frontity.settings.js
cristianbote: Gaah! Totally forgot to remove them sorry! On it right now!

projects/mars-theme/package.json
cristianbote:
```suggestion
```
Does this work?

projects/mars-theme/frontity.settings.js
cristianbote:
```suggestion
```
Remove it!

michalczaplinski (approved):
🚀

cristianbote:
So this one has been reviewed twice over 😄  once as the amp template, render to static markup and once as the extract the critical css. I believe it's ready.

PR #721, opened 2021-02-17
Core: Check if `frontity_name` is an array

DAreRodz:
**What**:

Support the latest version of `@types/koa` (2.11.8) that has changed the type definition of `ctx.query` from

```ts
query: any;
```

to 

```ts
query: ParsedUrlQuery;
```

which is basically a record of strings or an array of strings.

```ts
interface ParsedUrlQuery { [key: string]: string | string[]; }
```

In `koa`, the `ctx.query` is a result of parsing the query string with the [native node `querystring.parse()` function](https://github.com/koajs/koa/blob/master/lib/request.js#L175). 

`querystring.parse()` can return an array of strings if the same key appears multiple times in the querystring, e.g.:

```js
const { parse } = require('querystring');
const querystring = "frontity_name=test&foo=bar&frontity_name=test2"
const query = parse(querystring);
console.log(query);
/*
  { 
    foo: "bar"
    frontity_name: ["test", "test2"]
   }
*/
```

So, what does that mean for Frontity? 

**In case that the user (for some reason) provides two or more `frontity_name` url params, we always use the last one.**

**Why**:

To fix a typescript error that has arised recently inside the `settingsAndStore` middleware.

**How**:

I've just changed the way `frontity_name` is used in the `settingsAndStore` middleware, checking if it's a string or an array. If it's an array, it uses the last value.

**Tasks**:

- [x] Code
- [x] Unit tests

**Unrelated Tasks**

- [x] TSDocs
- [x] TypeScript
- [x] End to end tests
- [x] TypeScript tests
- [x] Update starter themes
- [x] Update other packages

**Additional Comments**

I found this issue after investigating why some unit tests were failing in #652 after mergin `dev` into that branch.

The reason why I decided to use the last value in the case `frontity_name` is an array is that it's the same we are doing in the `initial-state.ts` file, where we are overwriting values with the same key.

https://github.com/frontity/frontity/blob/8b22a6e558269b11c8137d3c265d965896f9023c/packages/core/src/server/utils/initial-state.ts#L36-L43

### DIFF
--- a/.changeset/empty-ties-learn.md
+++ b/.changeset/empty-ties-learn.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Handle the case where the query param `frontity_name` is an array of strings.
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,11 +4,12 @@
 	"lockfileVersion": 1,
 	"dependencies": {
 		"@babel/code-frame": {
-			"version": "7.5.5",
-			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
-			"integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
+			"version": "7.12.11",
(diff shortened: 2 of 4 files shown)


michalczaplinski (approved):
Okay, looks great @DAreRodz ! 

I've just added a slightly longer explanation of why we do it this way, because I didn't understand it myself at first

PR #722, opened 2021-02-17
Amp html template

### LINKED ISSUE #558: Overwrite the HTML template. [5pt]

---

The Google AMP framework needs a specific HTML template because a normal HTML document doesn't validate.

For that reason, the `@frontity/amp` package needs to be able to overwrite the hardcoded HTML template.

This can be done with a `beforeSSR` function but needs the `template` to be exposed in `libraries.frontity`.

```js
// packages/amp/src/index.js

export default {
  actions: {
    amp: {
        libraries.frontity.template = ampTemplate;
      },
    },
  },
};
```

The template should be something like this: https://github.com/frontity/frontity/blob/31d34e4d16042ebc73b4c30a9f53dfae36c633a2/packages/core/src/server/templates/amp.ts

### LINKED ISSUE #560: Change the render to be renderToStaticMarkup. [3pt]

---

The Google AMP framework doesn't need the custom `data-` attributes added by React, so instead of using `renderToString`, the `@frontity/amp` package needs to make Frontity use `renderToStaticMarkup`.

This can be done with a `beforeSSR` function. It'll be necessary to create a separate `server.js` file because we don't want to add `react-dom/server` to the client bundle.

```js
// packages/amp/src/server.js

export default {
  actions: {
    amp: {
        libraries.frontity.render = renderToStaticMarkup;
      },
    },
  },
};
```

It needs the `render` function to be exposed in `libraries.frontity`.

cristianbote:
**What**:

A dedicated AMP template for the `@frontity/amp` package.

**Why**:

In order to be AMP compliant we need to serve a custom html template.

**How**:

Hooking into `beforeSSR` and define the render and the template method.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


cristianbote:
As you might be aware, I've had some troubles figuring out what's happening with the loadable components, since I was getting only the _header_ part rendered. Chasing the issue, it seems that on the server side the chunk loading is sync but it's dependant on an existing chunk extractor context. Relevant code from `@loadable/component` https://github.com/gregberge/loadable-components/blob/main/packages/component/src/createLoadable.js#L112:L127.

So that means even though we are gonna be outputting as static markup, we have to wrap the `<App />` component with the collector.

As you might've noticed, I've merged back the entry point for `@frontity/amp` as we do not need a `client` side handler, since there's not gonna be any script of ours running. So, no need to hydrate the dom and css.

LE: Gonna turn this into a draft cause I forgot to add an e2e test for these changes.

luisherranz:
There is a proposal to add a `bundles` setting to sites which should remove the generation of client bundles and the loadable requirement when it is set to `"server"`.

```js
export default {
  name: "my-amp-site",
  bundles: ["server"],
  state: {
    //...
  },
  packages: [
    //...
  ],
};
```

There is no IP yet, but it is described here: https://community.frontity.org/t/amp-package/388/10

If that would solve this problem, I guess we can simply use a workaround for now.

cristianbote:
Hey @luisherranz, thanks for looking into it. At a first glance it doesn't seem to be the case, since it's about how the loadable components are defined and loaded on the backend. Maybe we could raise this issue in the loadable components repo. It's not something that we are doing but is how the loadable components are being loaded.

The `bundles` property if I understood it completely, will bundle up _only_ that part of the code.

luisherranz:
Oh, ok. So it's not an issue about removing the client bundle but that we actually _cannot remove the client bundle_ or loadable won't work. Is that right?

michalczaplinski:
Hey @cristianbote I've removed myself from reviewers for now since the PR is still in draft and until you add the e2e tests like you've mentioned 🙂

e2e/e2e.js
```diff
@@ -172,16 +172,16 @@ process.env["CYPRESS_FRONTITY_MODE"] =
...
           });
         } else {
           await cypress.run({
             env: { WORDPRESS_VERSION: wpVersion },
             browser,
             spec: spec
-              ? `./integration/**/${spec}.spec.js`
-              : `./integration/${suite}/**/*.spec.js`,
+              ? `./integration/**/${spec}.spec.{j,t}s`
+              : `./integration/${suite}/**/*.spec.{j,t}s`,
```
cristianbote: I've added these since the config was excluding the TS specs.

e2e/integration/frontity-01/amp.spec.ts
```diff
@@ -0,0 +1,23 @@
+const assertTemplate = () => {
+  cy.get("html").should("have.attr", "amp");
+  cy.get("style[amp-boilerplate]").should("exist");
+  cy.get("style[amp-custom]").should("exist");
+  cy.get("style[amp-custom]").should("exist");
+  cy.get("body script").should("not.exist");
```
cristianbote, line 6: This pretty much sums app the needed things for a functional AMP template. The only one that I'm not sure if should be part of this, is the amp script tag. Should we check for that as well?

packages/amp/src/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+ */
+export const removeAmp = (link: string): string =>
+  link.replace(/\/amp\/?($|\?|#)/, "/");
+
+export default {
+  name: "@frontity/amp",
+  actions: {
+    amp: {
+      init: ({ libraries }) => {
+        if (libraries.source) {
```
cristianbote, line 28: @michalczaplinski I've added this check in here, since there can be static website without a _source_. And those kind of website do not need this parsing. Am I assuming correct in here?

cristianbote:
> Oh, ok. So it's not an issue about removing the client bundle but that we actually _cannot remove the client bundle_ or loadable won't work. Is that right?

Yup. 🦊

cristianbote:
It's all ready now @michalczaplinski 😄

packages/amp/src/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+          // Cleanup the head of scripts, but leave only the `amp-` based ones.
+          rest.head = rest.head.filter((tag) => {
+            return /<?script.+?amp-/g.test(tag);
+          });
+
+          // Push the custom css style tag.
+          rest.head.push(`<style amp-custom>${css}</style>`);
+
+          // No scripts allowed.
+          rest.scripts = [];
```
michalczaplinski: Am I seeing it correctly or this line is not actually needed? 

The `ampTemplate` does not accept `scripts` as an argument, so even if `rest.scripts` array was filled, the template would just ignore it, right?

e2e/integration/frontity-01/amp.spec.ts
michalczaplinski: Yes, please :) 

We will also need the `<link rel="amphtml" href="https://domain.com/some-post/amp/" />` element like [Luis mentioned](https://community.frontity.org/t/amp-package/388/32?u=mmczaplinski)

Side note: the tests are absolutely fine, but I think that going forward we should intergate the [amphtml-validator](https://www.npmjs.com/package/amphtml-validator) into the e2e test flow.

I've gotten started on this in my PR to [Add the AMP Processors](https://github.com/frontity/frontity/pull/718) and now that we have the template, I think that I can finish with them once the parent (CSS in the head) PR is merged 🙂

e2e/packages/static-theme/src/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+    margin: 0,
+    padding: 0,
+  },
+  props.p && { padding: props.p + "em" },
+  props.m && { margin: props.m + "em" },
+  props.bg && { background: `var(--${props.bg})` },
+  props.color && { color: `var(--${props.color})` },
+  props.size && { fontSize: props.size + "em" },
+  props.r && { borderRadius: props.r + "em" },
+]);
```
michalczaplinski, line 88: Some of the props like `color` or `size` are ending up in the DOM and it's disallowed in AMP (at least according to https://validator.ampproject.org/). Have a look:

https://www.loom.com/share/ad5fbecd48644b549f2b070820a064b8

michalczaplinski (changes requested):
Hey @cristianbote Everything is great, I just have some comments before we merge this :)

packages/amp/src/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+                  <App />
+                </CacheProvider>
+              )
+            )
+          );
+
+          return result;
+        };
+
+        // Define the emotion style tag with the render result.
```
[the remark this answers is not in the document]
cristianbote: Totally @michalczaplinski, changing it.

packages/amp/src/index.tsx
cristianbote: Indeed! Good catch! 😄

e2e/packages/static-theme/src/index.tsx
cristianbote: Oh, I see. And since we can't control the end-user usages we might at least control our tests and results to make sure their compatible. Is this your line of thought?

packages/amp/src/index.tsx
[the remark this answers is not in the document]
cristianbote: Just as a heads-up, I usually use the `rest` when I wanna mutate something and spread it over to another call 😄  just me being lazy.

```js
const foo = ({param, ...rest}) => {
  
  // No need to pass attr separately
  rest.attr = 1;

  return baz({
    ...rest
  });
};
```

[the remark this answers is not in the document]
cristianbote: Oh actually the `html` needs to be destructed from within `result` cause by default the `html = result` but in this case result is polymorphic in shape, as is the result of `render` fn. So, `html` needs to be destructed.

cristianbote:
Hey @michalczaplinski  and @luisherranz I believe the rel link will have to be part of another change cause this one grows indefinitely  😄 . What are your thoughts?

cristianbote:
@michalczaplinski also about the amp-validator, I would love to get this merged into the css-in-head branch, and that one merged in dev. How does that sounds to you? I want this to happen like this because of the chain of the changes.

e2e/packages/static-theme/src/index.tsx
michalczaplinski: Yup, that was my thinking basically 🙂 I thought that `emotion` should be filtering out the props passed to styled components that are not part of the HTML spec but maybe that's not the case.

michalczaplinski:
> I believe the rel link will have to be part of another change cause this one grows indefinitely

yup, that's absolutely fine. I should have clarified that I was pointing this one out in case you missed it going forward, it wasn't necessary for this PR 🙂

PR #729, opened 2021-02-24
Downgrade webpack-dev-middleware@3.7.0 for faster dev builds

### LINKED ISSUE #723: Webpack builds are now slower than they used to be
**_Bug report_**

**Expected behavior**

Webpack rebuilds are fast.

**Observed behavior**

Webpack rebuilds are slow. Or at least slower than before.

- Start the project in `e2e/project` running `npx frontity dev`.
- Wait until Webpack starts.
- Make a change to a file of that project, for example to `e2e/packages/ads/src/index.tsx`.
- It should take a couple of seconds to rebuild the entire project, but now it takes like 30 seconds.

**Possible solution**

I've been triaging this and the regression was clearly introduced in #640. 

There are many changes in that PR. I am not sure right one which one is causing the problem.

At first, I thought it could be the changes of babel, but reverting those didn't solve the problem. So maybe it's related to an updated package.

I'll keep triaging this until I find the exact problem 🙂

cristianbote:
**What**:

Downgrade `webpack-dev-middleware` to `3.7.2` which is the last known version that didn't cause the slowness.
**Why**:

The dev builds reload times were getting really slow.

**Tasks**:

- [x] Update other packages

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes

**Additional Comments**

I did not investigate on the _why_ is slow. This is a solid gain. If we're gonna adopt `webpack@5` we might need to look into the _why_.

### DIFF
--- a/.changeset/cyan-toes-applaud.md
+++ b/.changeset/cyan-toes-applaud.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Downgrade webpack-dev-middleware@3.7.2 which is the last known version that was not slow.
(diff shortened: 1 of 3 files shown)


luisherranz (approved):
Thanks a lot Cris!! Frontity is fast again 💪

PR #733, opened 2021-03-01
Update @frontity/google-analytics to work with the new GA4 format

### LINKED ISSUE #719: @frontity/google-analytics package doesn't work with Google Analytics 4 [5pt]
**_Bug report_**

**Expected behavior**

Google Analytics released a new way of using the tool, [Google Analytics 4](https://support.google.com/analytics/answer/10089681?hl=en), and `@frontity/google-analytics` package should work with it as it is the new default and recommended way.

**Observed behavior**

The package doesn't work with Google Analytics 4 because it has changed the library used to track pageviews and events.

Create a new property in Google Analytics and try to use `@frontity/google-analytics` package.

**Possible solution**

Google Analytics use now `gtag.js` instead of `analytics.js` so we have to adapt our package following [this guide](https://developers.google.com/analytics/devguides/collection/upgrade/analyticsjs#:~:text=Measure%20pageviews,-analytics.&text=js%20uses%20trackers%20to%20send,specified%20in%20a%20config%20command.). It seems that old properties not using Google Analytics 4 are also compatible with `gtag.js`, so this change should be backwards compatible.

As far as I understood what we have to do is:

**1. Change the `<Head>` script**

Right now -> https://github.com/frontity/frontity/blob/dev/packages/google-analytics/src/components/index.tsx#L29-L38
According to [this article](https://developers.google.com/analytics/devguides/collection/ga4), it should be something like:

```
<script
    async
    src={`https://www.googletagmanager.com/gtag/js?id=MEASUREMENT_ID`}
 />
<script>
    {`
    window.dataLayer = window.dataLayer || [];
    function gtag(){dataLayer.push(arguments);}
    gtag('js', new Date());
   
    gtag('config', 'MEASUREMENT_ID');
    gtag('config', 'MEASUREMENT_ID_2');
    `}
</script>
```

**2. Change the way we send pageviews**
Right now -> https://github.com/frontity/frontity/blob/dev/packages/google-analytics/src/index.ts#L33-L51
According to [this](https://developers.google.com/analytics/devguides/collection/upgrade/analyticsjs#measure_pageviews_with_specified_trackers), we should do something like:

```js
ids.forEach((id) =>
    window.gtag("event", "page_view", { send_to: id })
);
```

**3. Change the way we send events**
Right now -> https://github.com/frontity/frontity/blob/dev/packages/google-analytics/src/index.ts#L52-L77
According to [this](https://developers.google.com/analytics/devguides/collection/upgrade/analyticsjs#measure_pageviews_with_specified_trackers), we should do something like:

```js
window.gtag("event", name, {
    event_action: name,
    event_category: category,
    event_label: label,
    value: value,
    ...rest,
});
```

cristianbote:
This change will require a major version bump for `@frontity/google-analytics` since the ways of handling pageviews and event data has changed.

**What**:

Update @frontity/google-analytics to work with the new GA4.

**Why**:

GA4 introduces a new way of sending events and pageview data.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests
- [x] TypeScript tests
- [x] Update other packages

**Unrelated Tasks**

- [ ] TSDocs
- [ ] End to end tests
- [ ] Update starter themes

**Additional Comments**


packages/google-analytics/src/index.ts
```diff
@@ -59,19 +41,17 @@ const googleAnalytics: GoogleAnalytics = {
...
-              eventValue: value,
-              ...rest,
-            });
+        ids.forEach((id: any) => {
+          const { category, label, value, ...rest } = payload;
+
+          window.gtag("event", name, {
+            send_to: id,
+            event_category: category,
+            event_label: label,
```
luisherranz, line 50: It doesn't seem like GA4 is using action, category and label anymore, is it? https://developers.google.com/analytics/devguides/collection/ga4/events#recommended_events

As this is a major bump, maybe we should change the event to something that makes more sense for the new events.

luisherranz (changes requested):
@cristianbote, the e2e tests are failing. Can you please check it out?

packages/google-analytics/src/index.ts
cristianbote: Oh, hmm 🤔  this is odd. I was using this guide https://developers.google.com/analytics/devguides/collection/upgrade/analyticsjs#measure_events which has a before and after examples.

SantosGuillamot: This new implementation, with the new library, supports also Universal Analytics (previous to GA4). I think that's the reason they keep sending `event_category` and `event_label`. Most of the users are currently using Universal Analytics, so I guess we should keep it working as well.

luisherranz: But what if a user wants to send one of the new events? Shouldn't we at least support both?

~Anyway, let's talk it over https://community.frontity.org/t/google-analytics-package/1084/16.~

luisherranz: I was writing in https://community.frontity.org/t/google-analytics-package/1084/16 to talk about the implementation of this, but I've just realized that the implementation that Cristian did works for both cases (old and new) thanks to the `...rest` spread in the event:

```js
const { category, label, value, ...rest } = payload;

window.gtag("event", name, {
  send_to: id,
  event_category: category,
  event_label: label,
  value,
  ...rest, // <-- adds any other payload
});
```

If an event contains `category` or `label`, it would be translated to `event_category` and `event_label`, and if not, it will be passed down as part of `...rest`.

So sending the new events (things like this: https://support.google.com/analytics/answer/9268036) should work, shouldn't it?

luisherranz (approved):
Ok, sorry for my other message. I think this should work great for both the new events and the category/label ones.

Code is great, and Cristian even added an e2e test that sends a custom payload.

Everything perfect then! Thanks Cris!! 😄👏

juanmaguitar:
@cristianbote or @luisherranz 
Is there a FD for this? 
Are there new features that require some documentation? In that case, Is there a documentation issue created?

As per [this post](https://community.frontity.org/t/code-documentation-workflow/1633/55?u=juanma) I understand there's some documentation to be done here

luisherranz:
Cris and Mario are talking about this in the initial Google Analytics FD: https://community.frontity.org/t/google-analytics-package/1084/19?u=luisherranz

PR #735, opened 2021-03-01
Fix incompatible types of fs-extra with node 12

cristianbote:
Apparently `@types/fs-extra@9.0.8` is based on node v14 where those types exists. More related to the problem in here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51155#issuecomment-787881250.

**Tasks**:

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/e2e/integration/frontity-01/smart-adserver.spec.ts
+++ b/e2e/integration/frontity-01/smart-adserver.spec.ts
@@ -1,7 +1,7 @@
 /**
  * Disable temporary these tests until the ad calls return content.
  */
-describe("Smart Adserver", () => {
+describe.skip("Smart Adserver", () => {
   beforeEach(() => {
     cy.visit("http://localhost:3001?frontity_name=smart-adserver");
   });
--- a/package-lock.json
+++ b/package-lock.json
@@ -4382,6 +4382,12 @@
 				"fs-extra": "^8.1.0"
 			},
 			"dependencies": {
+				"@types/node": {
+					"version": "12.20.4",
+					"resolved": "https://registry.npmjs.org/@types/node/-/node-12.20.4.tgz",
+					"integrity": "sha512-xRCgeE0Q4pT5UZ189TJ3SpYuX/QGl6QIAOAIeDSbAVAd2gX1NxSZup4jNVK7cxIeP8KDSbJgcckun495isP1jQ==",
+					"dev": true
+				},
 				"fs-extra": {
 					"version": "8.1.0",
 					"resolved": "https://registry.npmjs.org/fs-extra/-/fs-extra-8.1.0.tgz",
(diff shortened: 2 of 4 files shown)


DAreRodz:
Hey, Cristian, would it be better if we update the version of `@types/node` instead? I'm not sure tho, just asking because Node 14 is the current LTS. :thinking:

cristianbote:
@DAreRodz good call. Let me try it.

cristianbote:
On a second thought @DAreRodz 😄  I think the other packages do require `@types/node@12.x`. I think bumping `node` should be dealt separately. Thoughts?

luisherranz:
Looking into this 🙂

I think at some point I tried to update to `@types/node@14` but some other types were not compatible and I didn't spend time to fix it, so I went back to `@types/node@12`. Maybe it was `@types/fs-extra`.

luisherranz:
Yep, updating to `@types/node@14` fixes the issue, at least in my local environment. I agree with David that it would be a better fix 👍

luisherranz:
Ok, that fixes the unit tests, but now the smart-ads e2e tests are failing. I'm going to try to figure out what's going on...

cristianbote:
Oh, but that means our current implementations across system, will make use of `node@14.x` type definitions which means we'll inevitable use things that might not work for `node@12.x`. Isn't this a concern? Are we prepared to bump the node version as well?

@luisherranz yeah the smart ad tests were failing with `fs-extra@9.0.7` as well.

> luisherranz removed request for michalczaplinski, DAreRodz and SantosGuillamot 1 hour ago

I've added everyone cause I believe it'll be beneficial for all of us to be aware 😄

luisherranz:
> Oh, but that means our current implementations across system, will make use of node@14.x type definitions which means we'll inevitable use things that might not work for node@12.x. Isn't this a concern? Are we prepared to bump the node version as well?

We've never used the types for that, although is not a bad point... But right now Frontity supports Node10, 12, and 14 and we were using the types of node 12.

What we've done up until now regarding that is to use `babel-preset-env` to make sure the server bundle is transpiled for Node 10 (https://github.com/frontity/frontity/blob/dev/packages/core/src/config/babel/index.ts#L45) although we don't inject any polyfills, so we also run all the unit tests in all versions: 10, 12 and 14.

luisherranz:
The SmartAds e2e are failing because they were deactivated by SmartAds again so I have disabled them.

_This is the last time we do this. Either they activate them forever or we do not test their library._

luisherranz:
> Oh, but that means our current implementations across system, will make use of node@14.x type definitions which means we'll inevitable use things that might not work for node@12.x. Isn't this a concern?

So I guess that what would make more sense taking into account that we have the help of Babel for Node as well would be to target the latest version (Node 14 right now) and let Babel do the transpilation, don't you think?

luisherranz (approved):

luisherranz:
Feel free to merge this if you agree with me. 

Well, if you don't agree maybe we should also merge this and discuss it in a different place. After all, we are already using the types of Node 12 😄🤷

cristianbote:
Oh @luisherranz that's totally right, we are already using `12.x` types with `10.x` code 😄  so my point is moot. Totally in agreement and thanks for indulging my comment. 💪

luisherranz:
Hey, using the types of node 10 it's not a bad point! 

We can try downgrading and taking a look at what happens with the types of other packages, like `fs-extra`. Then figure out if it makes sense or not. 

Do you want to give it a try?

PR #741, opened 2021-03-09
Fix for passing `classname` to the DOM in WebComponents created with Html2React

### LINKED ISSUE #742: Html2React passes the `classname` down to custom web components if the element has a `style` attribute
**Expected behavior**

When using Web Components in React (e.g. `amp-img`, `amp-iframe`) we have to pass a `class` prop to the component instead of `className` like in regular react components. Currently emotion has a known bug where you cannot use the `css` prop on the Web Components because of this. See: https://github.com/emotion-js/emotion/issues/1071 or https://github.com/emotion-js/emotion/pull/1089.

In our the parse function in html2react, if a component has a `style` attribute, we pass the css prop to that component: https://github.com/frontity/frontity/blob/b21c48757b45904c4c13affb35e8dfbc7daf51c8/packages/html2react/src/libraries/parse/index.ts#L43-L45

However, due to the aforementioned emotion bug, this actually passes a classname prop to the DOM if the component is a Web Component.

I've created a failing test case in #741.

**Possible solution**

I think we can check if the component is a Web Component by using the same function that React uses to check it: https://github.com/facebook/react/blob/c954efa70f44a44be9c33c60c57f87bea6f40a10/packages/react-dom/src/shared/isCustomComponent.js

Then, if the component is a Web Component, we can create the class name using `css` from `@emotion/css` and pass it directly to the class prop: 

```js
// src/libraries/parse/index.ts

import { css as cssFunc } from '@emotion/css';

// inside of adaptNode function
if (isWebComponent()) {
  props.class = css(value);
} else {
  props.css = cssFunc(value);
}
```

michalczaplinski:
**What**:

Fix the issue where adding `style` on a custom component like `amp-img` causes passing of `className` property to the DOM.

**Why**:

To fix #742 

**How**:
I think we can change it by updating the `parse` function in `@frontity/html2react` to check if the current node is a custom component.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


michalczaplinski:
@DAreRodz I've realized that using `@emotion/css` is not going to work correctly with SSR so I'm removing you from reviewers until I figure out a better solution.

luisherranz:
I've proposed another approach here: https://github.com/frontity/frontity/issues/742#issuecomment-793577887

packages/html2react/jest.config.js
```diff
@@ -1,2 +1,5 @@
 const rootConfig = require("../../jest.config.js");
-module.exports = rootConfig;
+module.exports = {
+  ...rootConfig,
+  snapshotSerializers: ["@emotion/jest/serializer"],
```
michalczaplinski, line 4: I've added https://emotion.sh/docs/@emotion/jest to improve jest snapshots.

packages/html2react/package.json
```diff
@@ -21,12 +21,14 @@
     "test:inspect": "node --inspect ../../node_modules/.bin/jest --watch --no-cache --runInBand"
   },
   "devDependencies": {
+    "@emotion/jest": "^11.2.1",
     "@frontity/connect": "^1.2.1",
     "@types/react-test-renderer": "^17.0.0",
-    "@testing-library/react": "^9.4.0",
+    "@testing-library/react": "^11.2.5",
```
michalczaplinski, line 27: Updated to the latest version because it was throwing errors when running tests locally.

packages/html2react/src/libraries/__tests__/__snapshots__/component.tests.tsx.snap
```diff
@@ -11,8 +11,12 @@ Array [
...
      for testing purposes.
   </p>,
-  <div
-    className="hello css-12gt2kf"
+  .emotion-0 {
+  width: 180px;
+}
+
+<div
+    className="hello emotion-0"
```
michalczaplinski, line 19: This appears now in the snapshots because I've added the https://emotion.sh/docs/@emotion/jest package

packages/html2react/src/libraries/__tests__/component.tests.tsx
```diff
@@ -1,11 +1,20 @@
+/* eslint-disable @typescript-eslint/no-empty-function */
```
michalczaplinski, line 1: The processors in the first test in the file are empty functions so had to add that to the file to please eslint.

luisherranz (approved):
You are absolutely right.

What about using a `<Wrapper>`, like Mateusz explained here: https://github.com/emotion-js/emotion/issues/1071#issuecomment-444532736?

I have done a quick test and I think it can work: https://www.loom.com/share/3b38039890a2499297959026a2ef9b9e

This is the code of the `Wrapper`:

```js
const Wrapper = ({ Component, children, className, ...props }) => {
  return (
    <Component {...props} class={className}>
      {children}
    </Component>
  );
};
```

```js
if (
  typeof node.component === "string" &&
  isCustomComponent(node.component, node.props)
) {
  return (
    <Wrapper
      key={index}
      css={node.props.css}
      Component={node.component}
      {...node.props}
    >
      {node.children ? handleNodes({ nodes: node.children, ...payload }) : null}
    </Wrapper>
  );
}
```

Also, could you please add more props and children to the tests? That way we can make sure that either solution you finally choose (`Wrapper` or `ClassNames`) will correctly pass down any props and children to the custom component.

I'm going to approve this now so you can merge it once it's done and continue your work with the AMP processors 🙂👍

michalczaplinski:
OK, awesome. Thanks for the tip about the `Wrapper` component @luisherranz. You're right it makes the implementation simpler. 👍 

I've added some extra tests to check for passing children, passing other props and handling nested custom components like you suggested.

I've also realized that we have another small bug in the `parse()` function: If an element in the content has a `className` attribute, it will be passed to the DOM as `class` and also overwrite any `class` attribute on that element. Raised a bug: #747 

Will merge this now 🙂

luisherranz:
Awesome. Great work Michal! 👏😄

PR #745, opened 2021-03-10
Fix `@frontity/tiny-router` when the initial link is not normalized

### LINKED ISSUE #744: Frontity stops working when visiting a link without a trailing slash
**_Bug report_**

This was reported first in the community: https://community.frontity.org/t/error-when-opening-urls-that-don-t-finish-with/4184

**Expected behavior**

If you go to **any link without a trailing** slash (e.g. https://mars.frontity.org/2016/the-beauties-of-gullfoss) everything should work as usual.

**Observed behavior**

SSR works fine, but images are not loaded and links reload the whole document. That happens because Frontity has thrown an exception and stopped working.

This is the exception:

Screenshot from 2021-03-10 18-56-20

**Possible solution**

The issue is located here:

https://github.com/frontity/frontity/blob/3a962aac3f926c43b42713f130694c4455c2ea72/packages/tiny-router/src/actions.ts#L201-L216

In those lines, when `state.frontity.initialLink` doesn't have the trailing slash, `link` is always different than `state.frontity.initialLink` because it's the same link but normalized, i.e. with the trailing slash.

What happens then is that both links (w/ and w/o trailing slash) already point to the same data object, and when the derived `initialDataObject` prop is resolved, it creates an infinite loop.

The problem is, `link` (normalized) is compared with `state.frontity.initialLink` (not normalized), so it should be enough if we compare both before normalizing `link`.

DAreRodz:
**What**:

Prevent Frontity from stop working when the initial link is not normalized

**Why**:

To fix #744 

**How**:

Changing the part where `state.frontity.initialLink` is compared with the browser link.

`state.frontity.initialLink` is never normalized, but `link` is normalized just before comparing both of them.

As both links already point to the same data object, the following code creates an recursive loop that never ends.

https://github.com/frontity/frontity/blob/3a962aac3f926c43b42713f130694c4455c2ea72/packages/tiny-router/src/actions.ts#L209-L211

**Tasks**:

- [x] Code
- [x] End to end tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


michalczaplinski (approved):

michalczaplinski:
Awesome work and well spotted @DAreRodz! How did we ever miss this bug? 😮

I've just added a more expansive comment and a changeset. 

I notice that you ticked the changeset in the task list but there was no changeset so maybe you had some commits that you didn't push? In any case, I ll just approve without merging in case you had any other commits left.

DAreRodz:
> I notice that you ticked the changeset in the task list but there was no changeset so maybe you had some commits that you didn't push?

Oh yeah, that's exaclty what has happened! I only forgot to push the changeset, haha. Thanks for adding it. :blush:

luisherranz:
@SantosGuillamot are you going to release a new version with this?

SantosGuillamot:
Yes, I am planning to do it on Monday. Is there any issue?

luisherranz:
No. I was just curious. Sorry 🙂

PR #746, opened 2021-03-10
Fix overwriting of `class` by `className` attribute in WordPress content

### LINKED ISSUE #747: `className` attribute in WordPress content overwrites the `class`.
**Observed behavior**

When a user has an element with a `className` attribute in their content, this attribute will be passed as `className` to the final react component created with `Html2React`.

Additionally, when a user has an element with both a `className` and `class` attribute in the WordPress content, that `className` will overwrite the `class`.

I've created a test case to show the behavior in #746 

**Possible solution**

I think that we should either a) ignore the `className` attribute in the user content; or b) Merge the content from `class` and `className` attributes.

Personally, I think that a) is a more reasonable solution and leads to less surprising behavior.

michalczaplinski:
**What**:

Create a test case to show the problem in #747 

**Why**:

To fix #747 

- [x] Code
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


luisherranz (approved):
Looks good! Thanks guys 🙂 👍

PR #749, opened 2021-03-11
Run Cypress tests in Browserstack

### LINKED ISSUE #629: BrowserStack - Cypress tests [13pt]
We are going to use BrowserStack to run our tests in different OS and browsers, to ensure that Frontity is reliable in all the scenarios. ~As part of this, we are also updating Babel.~

Branch: https://github.com/frontity/frontity/tree/update-babel-polyfill

luisherranz:
**What**:

Run our non-WordPress e2e tests in BrowserStack.

**Why**:

Because in BrowserStack we can run different versions of those browsers and different operative systems. We are specially interested in running:

- Other browsers: Edge and Firefox
- The oldest versions that Cypress and BrowserStack support: Chrome 66, Firefox 60 and Edge 80.
- Different OS: Win 10, Big Sur (latest macOS) and Mojave (oldest macOS).

This is going to be the combination of browsers/os that we will run:

- Win 10
  - Chrome
    - Latest
    - 66
  - Firefox
    - Latest
    - 60
  - Edge
    - Latest
    - 80
- Big Sur
  - Chrome
    - Latest
  - Firefox
    - Latest
  - Edge
    - Latest
- Mojave
  - Chrome
    - 66
  - Firefox
    - 60
  - Edge
    - 80

We don't run the e2e that require a WordPress instance because the Cypress tasks that we use to install plugins and load databases don't work in BrowserStack.

**How**:

I have updated our `e2e.js` script and now it accepts a new flag for cypress: `--cypress browserstack` which instead of running Cypress in the local machine, it sends the tests to the BrowserStack cloud.

I have also added GH workflows to run tests: 
- The latests Chrome, Edge and Firefox versions on Windows 10 on each PR.
- All the browsers mentioned about on each `dev` push.

**Tasks**:

- [x] Configure BrowserStack for Cypress
- [x] Add BrowserStack commands to `e2e.js` script
- [x] Update e2e Readme
- [x] Add GH workflows
- [x] Add BrowserStackLocal binaries

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


luisherranz:
This is the current status: https://www.loom.com/share/b9f711b55c0d4fcbbc6dace53a50e079

List of remaining things: 

- [x] Change the local identifier so it is unique per commit.
- [x] Find out why e2e tests are failing
- [x] Find out why the `parallel` option in the `dev` workflow is 12 instead of 1.
- [x] Open issue for Infinite Scroll tests failing in Firefox.

luisherranz:
I've opened a ticket to ask the BrowserStack team why the parallel option is not being used.

I'm going to block this until they answer.

EDIT: This is what happens by the way: https://www.loom.com/share/88395715def549bea6de38eea1824eb3

luisherranz:
They answered. It seems that the number of browser combinations needs to be less than the double of our parallel tests, which is 5. So we can only test 10 combinations of browsers at a time.

I am going to try reducing it to 9 to see if the `"parallels"` option works then.

luisherranz:
I fixed the e2e tests https://github.com/frontity/frontity/pull/750, although we will need to review that to update our tests to WP 5.7.

I opened the Infinite Scroll issue https://github.com/frontity/frontity/issues/751

luisherranz:
Ok, the `parallels` option doesn't seem to work to limit the number of machines used by a test. I think it works to increase it only.

I don't think that's a problem for the new action that we want to run in the PRs because we need only two machines, which means that up to 5 PRs can be tested at the same time.

For the tests that we plan to run on the `dev` branch maybe we need to run them one after another instead of in parallel. I'll talk with Mario on Tuesday (I don't work on Monday) to decide how to do this.

By apart from that that we can decide later this is ready for review so I have recorded a video: https://www.loom.com/share/1f83850aa3c2426fb72cd68f93d1d276

e2e/.env-example
```diff
@@ -0,0 +1,4 @@
+# Go to https://automate.browserstack.com/dashboard and copy the username and
+# access key. Then, rename this file to ".env".
```
DAreRodz: I would suggest here copying the file instead of renaming it, or we could end up removing the `.env-example` file by accident.

e2e/plugins/index.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable import/no-unresolved */
+/* eslint-disable import/no-unresolved  */
```
DAreRodz: I guess this change is not necessary. :sweat_smile:

.github/workflows/e2e-bs-dev.yml
```diff
@@ -0,0 +1,56 @@
+name: e2e browserstack (dev)
+
+on:
+  push:
+    branches:
+      - dev
+      - browserstack-cypress
```
DAreRodz: We would have to remove the `browserstack-cypress` branch from here before merging the PR, right?

DAreRodz (approved):
These changes are great so far, good job @luisherranz! :raised_hands:

PS: I just leaved some reminders/suggestions.

e2e/.env-example
luisherranz: Done 👍

.github/workflows/e2e-bs-dev.yml
luisherranz: Yes. I've just done it.

e2e/integration/frontity-01/iframe.spec.js
```diff
@@ -1,24 +1,30 @@
...
-    return cy
-      .window()
-      .its("HTMLIFrameElement")
-      .then((HTMLIframeElement) => {
-        expect("loading" in HTMLIframeElement.prototype).toBe(true);
-      });
-  });
+  if (
+    Cypress.browser.family === "chromium" &&
+    Cypress.browser.majorVersion > 77
```
cristianbote: Was watching the video and notice that the caniuse table was saying that _starting_ with version 77. So, I believe this condition should be:
```suggestion
    Cypress.browser.majorVersion >= 77
```

In any case, how would you feel that this condition should actually be condition we assert inside the `expect` call? That way we don't rely on browser versions but rather on the functionality.

luisherranz: > how would you feel that this condition should actually be condition we assert inside the expect call? 

Can you show an example?

luisherranz: Actually, I don't know why we are testing that the native lazy load should exist in the first place.

We should test that `loading="lazy"` is added to the `` or `<iframe>` tags. The rest is up to the browser.

cristianbote: Exactly my point. And I was thinking to write it in a way that avoids the browser version check. Something in the lines of:

```js
it("lazy attribute should be defined if it's supported", () => {
  const isSupported = "loading" in HTMLIframeElement.prototype;
  
  if (isSupported) {
    cy.scrollTo("topLeft");
    cy.get("iframe").should("have.attr", "loading", "lazy");
  } else {
    this.skip();
  }
});
```

luisherranz: Ohh, I got it now. That's a great point and makes much more sense, thanks Cris 🙂

However, in this specific case, as the `loading="lazy"` is added even if the native lazy load doesn't exist, I think I'm going to just remove the test that checks if native lazy load exists and leave only the one that checks if `loading="lazy"` was added.

luisherranz:
Ok, so all the tests are finally passing, we can finally merge this.

Let's see how reliable the BrowserStack cloud is 🙂

PR #752, opened 2021-03-12
Add e2e tests for head-tags with AIOSEO ^4.0.16

DAreRodz:
**What**:

Adds e2e tests to check the `@frontity/head-tags` package with a WordPress instance using the latest AIOSEO plugin (^4.0.16).

**Why**:

To check these changes are working: https://github.com/frontity/wp-plugins/pull/49

**How**:

Creating a new battery of tests based on the ones we have for the previous AIOSEO version.

**Tasks**:

- [x] End to end tests
- [x] Add a new `.sql` file once the new REST API - Head Tags plugin version is released.

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


DAreRodz:
The tests were created using a local WP instance and a local copy of the https://github.com/frontity/wp-plugins repository.

The only pending task would be to create a new `.sql` file once a new version of the REST API - Head Tags plugin is released, and use it to populate the e2e WP instance.

e2e/integration/wordpress-01/__snapshots__/head-tags-aioseop-3.spec.ts.snap
```diff
@@ -1,117 +1,20 @@
-exports[`Head Tags - All in One SEO Pack > Post > should render the correct canonical link #0`] = `
+exports[`Head Tags - All in One SEO Pack 3 > Post > should render the correct canonical link #0`] = `
```
DAreRodz, line 1: I've modified the title passed to `describe(...)` in the spec file so the name of these snapshots has changed.

e2e/integration/wordpress-01/__snapshots__/head-tags-aioseop-3.spec.ts.snap
```diff
@@ -1,117 +1,20 @@
...
-              "url": "http://my.frontity.site/hello-world/",
-              "name": "Hello world!"
-            }
-          }
-        ]
-      }
-    ]
-  }
-</script>
-`;
```
DAreRodz, line 104: This snapshot stopped being used long time ago, I've removed it.

e2e/integration/wordpress-01/__snapshots__/head-tags-wpseo.spec.ts.snap
```diff
@@ -1,92 +1,92 @@
-exports[`Head Tags - WP SEO 0.13 > Post > should render the correct canonical link #0`] = `
+exports[`Head Tags - WP SEO > Post > should render the correct canonical link #0`] = `
```
DAreRodz, line 1: For some reason, these snapshots were not updated and had the wrong name. Someone forgot to merge them before? :shrug:

DAreRodz (commented):
I've added a new spec file to test the Head Tags plugin with AIOSEO ^4.0.16. It's mostly based on  the previous spec file but checking more tags that weren't added in the previous version of the plugin.

cristianbote (approved):
@DAreRodz this looks totally fine to me!

Do you think would make sense to have a common spec for both versions and then have a separate spec only for the differences?

DAreRodz:
> Do you think would make sense to have a common spec for both versions and then have a separate spec only for the differences?

Not sure. From my point of view, it's better if we have each of them in a separate file, I don't know if moving the common tests to a new spec file would be worth it.

We can discuss it if you want, anyway. :slightly_smiling_face:

cristianbote:
Thanks for your reply @DAreRodz! I was only suggesting that cause the AIOSEO functionality does overlap, between versions, and I thought maybe it could be easier to maintain. But doesn't sound/look like it. We good 👍 .

PR #755, opened 2021-03-17
Remove `form.errors` from wp-comments package

michalczaplinski:
**What**:

As noticed in https://github.com/frontity/api-reference/pull/21#issuecomment-801197220 the `state.comments.forms[].errors` property in the wp-comments package is not used in any way. 

It was originally included to handle potential multiple errors for invalid REST params, but it is not necessary as explained in https://github.com/frontity/api-reference/pull/21#issuecomment-801257123

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


.changeset/stupid-windows-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontity/wp-comments": patch
```
michalczaplinski, line 2: It's technically a breaking change but it doesn't matter because we're in beta. Also the `errors` property is always an empty object so it should not be used by anybody for anything.

e2e/package.json
```diff
@@ -10,7 +10,7 @@
     "@types/execa": "^2.0.0",
     "@types/wait-on": "^4.0.0",
     "amphtml-validator": "^1.0.34",
-    "cypress": "^6.6.0",
+    "cypress": "^6.7.1",
```
michalczaplinski, line 14: Because of https://github.com/cypress-io/cypress/issues/15300#issuecomment-799548183

michalczaplinski:
The issue with the wordpress image for WP 5.7 still persists. I have tried several things in the `docker-compose.yml` file to make the error go away to make the error go away but I'm still not sure what the issue is. 

The diff: https://github.com/frontity/frontity/pull/755/files#diff-6e968fd9eb2b070c0878f0f075eebdec7c319c7b9f90533017e0ceddfd8ee1d2

Perhaps we can drop the support of 5.5 and only test the 5.6 version until the image for 5.7 gets fixed.

.github/workflows/e2e-tests-chrome.yml
```diff
@@ -10,24 +10,24 @@ jobs:
...
+            wp-version: latest
+            args: ""
             mode: "development"
           - name: Previous WordPress Version
-            wp-version: 5.5
-            args: "--prod --wp-version 5.5"
-          - name: Relative Public Path
             wp-version: 5.6
-            args: "--prod --public-path /custom/path/ --wp-version 5.6"
+            args: "--prod --wp-version 5.6"
```
michalczaplinski: Bumping the "Previous WordPress Version" to from 5.5 to 5.6 because Yoast 16 does not support WP 5.5.

e2e/docker-compose.yml
```diff
@@ -26,14 +26,15 @@ services:
     environment:
       WORDPRESS_DB_HOST: db
       WORDPRESS_DB_NAME: wordpress
-      WORDPRESS_DB_USER: root
+      WORDPRESS_DB_USER: test
```
DAreRodz: Hey, why did you have to change the WP database user? The e2e tests are failing because of that:

https://github.com/frontity/frontity/pull/755/checks?check_run_id=2135130303#step:5:1444

DAreRodz (changes requested):
Changes are fine :raised_hands:, but before approving them I would want to ask something. :point_down:

e2e/docker-compose.yml
DAreRodz: If it's better to use a `test` user, I guess we would have to specify it somewhere else in order to make the tests work.

michalczaplinski:
> Hey, why did you have to change the WP database user? 

Sorry @DAreRodz I should have removed you from reviewers until the e2e tests pass. I was trying out a couple of things in the `docker-compose.yml` file to see if they can fix the e2e tests in WP 5.7 docker image. 

For the time being we are just going to test only in WP 5.6 [(Luis has made a fix)](https://github.com/frontity/frontity/pull/749/commits/86355aa89cb37a3de76949f5b42c4b67302aaa3e) until the WP docker image for 5.7 is fixed.

I've removed all the changes related to the e2e tests so it should be good to merge now :)

DAreRodz (approved):
Don't worry. The code is fine so I'll approve this. :slightly_smiling_face:

 Just take a look at the e2e tests because it seems they are failing again. :eyes:

michalczaplinski:
> Just take a look at the e2e tests because it seems they are failing again. 👀

@DAreRodz Yeah, that was the one of the tests of the infinite scroll. It seems to fail randomly every now and then. Let's keep an eye on it.

PR #757, opened 2021-03-19
Run e2e tests in WordPress 5.7

luisherranz:
**What**:

Run our e2e tests with a WordPress 5.7 instance.

**Why**:

Right now we are running them with WordPress 5.6, which is not the latest version.

**How**:

Up until 5.6, we were using a docker container for the `wp-cli` to run certain Cypress tasks in the `wordpress` docker container.

In 5.7, that doesn't seem to work anymore in GH actions. I don't know yet what is causing the problem.

**Tasks**:

- [x] Fix End to end tests

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/.github/workflows/e2e-gh-actions.yml
+++ b/.github/workflows/e2e-gh-actions.yml
@@ -10,17 +10,27 @@ jobs:
       matrix:
         include:
           - name: Default
+            wp: latest
+            mode: "production"
             args: "--prod"
           - name: ES5 Bundle
+            wp: latest
+            mode: "production"
             args: "--prod --target es5"
           - name: Development
+            wp: latest
             mode: "development"
           - name: Previous WordPress Version
             wp: 5.6
+            mode: "production"
             args: "--prod --wp 5.6"
           - name: Relative Public Path
+            wp: latest
+            mode: "production"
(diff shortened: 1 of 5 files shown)


luisherranz:
@michalczaplinski I've done a video with a summary of the changes: https://www.loom.com/share/347fddf549e64c49add484691b1cea7e

The e2e test that failed while I was recording the video was from the infinite scroll suite, but I run it again and it passed. I guess that scrolling up and down is not the most reliable thing in Cypress 🤷

If you don't have any comments, feel free to merge after approval if you want to use the 5.7 version in your PR.

michalczaplinski (approved):
Awesome, thanks Luis! I've spent some time looking into this in the other PR already: https://github.com/frontity/frontity/pull/755 but I haven't managed to fix it.

Also, I totally agree that it makes sense that we hardcode the WP version and when a new version of WP gets released we can make a PR to update the workflow file to 5.8, 5.9, etc.

PR #758, opened 2021-03-19
Fetch data asynchronously in the postType handler

michalczaplinski:
Based on the comments in https://community.frontity.org/t/issue-with-having-2-duplicate-requests-to-posts-and-pages/4270/3 I investigated if we can optimize data fetching in postType handler.

Until now, we were fetching the data serially in the postType handler for each of `post`, `page` and `media` endpoints. With this improvement, the data is loaded and processed in parallel for all 3 endpoints.

**How**:

I have borrowed some ideas for the implementation from https://observablehq.com/@ehouais/multiple-promises-as-an-async-generator

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

### DIFF
--- a/packages/wp-source/src/libraries/handlers/__tests__/post-type.tests.ts
+++ b/packages/wp-source/src/libraries/handlers/__tests__/post-type.tests.ts
@@ -86,7 +86,7 @@ describe("postType", () => {
   });
 });
 
-describe.only("post", () => {
+describe("post", () => {
   test("doesn't exist in source.post", async () => {
     // Mock Api responses
     api.get = jest.fn().mockResolvedValue(mockResponse([post1]));
@@ -225,8 +225,9 @@ describe("page", () => {
     // Mock Api responses
     api.get = jest
       .fn()
-      .mockResolvedValueOnce(mockResponse([]))
-      .mockResolvedValueOnce(mockResponse([page1]));
+      .mockResolvedValueOnce(mockResponse([])) // post
+      .mockResolvedValueOnce(mockResponse([page1])) // page
+      .mockResolvedValueOnce(mockResponse([])); // media
     // Fetch entities
     await store.actions.source.fetch("/page-1/");
     expect(postTypeHandler).toHaveBeenCalled();
@@ -252,8 +253,9 @@ describe("page", () => {
     // Mock Api responses
     api.get = jest
       .fn()
-      .mockResolvedValueOnce(mockResponse([]))
-      .mockResolvedValueOnce(mockResponse([page1]));
+      .mockResolvedValueOnce(mockResponse([])) // post
+      .mockResolvedValueOnce(mockResponse([page1])) // page
+      .mockResolvedValueOnce(mockResponse([])); // media
     // Fetch entities
     await store.actions.source.fetch("/page-1/?some=param");
     expect(postTypeHandler).toHaveBeenCalled();
@@ -279,8 +281,9 @@ describe("page", () => {
     // Mock Api responses
     api.get = jest
       .fn()
-      .mockResolvedValueOnce(mockResponse([]))
-      .mockResolvedValueOnce(mockResponse([page1WithParent]));
+      .mockResolvedValueOnce(mockResponse([])) // post
+      .mockResolvedValueOnce(mockResponse([page1WithParent])) // page
+      .mockResolvedValueOnce(mockResponse([])); // media
     // Fetch entities
     await store.actions.source.fetch("/parent/page-1/");
     expect(postTypeHandler).toHaveBeenCalled();
@@ -293,6 +296,7 @@ describe("attachment", () => {
     // Mock Api responses
     api.get = jest
       .fn()
+      .mockResolvedValueOnce(mockResponse([]))
       .mockResolvedValueOnce(mockResponse([]))
       .mockResolvedValueOnce(mockResponse([attachment1]));
     // Fetch entities
@@ -320,6 +324,7 @@ describe("attachment", () => {
     // Mock Api responses
     api.get = jest
       .fn()
+      .mockResolvedValueOnce(mockResponse([]))
       .mockResolvedValueOnce(mockResponse([]))
       .mockResolvedValueOnce(mockResponse([attachment1]));
     // Fetch entities
--- a/packages/wp-source/src/libraries/handlers/__tests__/post-type.tests.ts
+++ b/packages/wp-source/src/libraries/handlers/__tests__/post-type.tests.ts
@@ -86,10 +86,10 @@ describe("postType", () => {
   });
 });
 
-describe("post", () => {
+describe.only("post", () => {
   test("doesn't exist in source.post", async () => {
     // Mock Api responses
-    api.get = jest.fn().mockResolvedValueOnce(mockResponse([post1]));
+    api.get = jest.fn().mockResolvedValue(mockResponse([post1]));
     // Fetch entities
     await store.actions.source.fetch("/post-1/");
     expect(store.state.source).toMatchSnapshot();
@@ -127,7 +127,7 @@ describe("post", () => {
 
   test("works with query params (doesn't exist in source.post)", async () => {
     // Mock Api responses
-    api.get = jest.fn().mockResolvedValueOnce(mockResponse([post1]));
+    api.get = jest.fn().mockResolvedValue(mockResponse([post1]));
     // Fetch entities
     await store.actions.source.fetch("/post-1/?some=param");
     expect(store.state.source).toMatchSnapshot();
@@ -151,7 +151,7 @@ describe("post", () => {
 
   test("works with types embedded", async () => {
     // Mock Api responses
-    api.get = jest.fn().mockResolvedValueOnce(mockResponse([post1withType]));
+    api.get = jest.fn().mockResolvedValue(mockResponse([post1withType]));
     // Fetch entities
     await store.actions.source.fetch("/post-1/");
     expect(postTypeHandler).toHaveBeenCalled();
@@ -164,7 +164,9 @@ describe("post", () => {
     // Mock Api responses
     api.get = jest
       .fn()
-      .mockResolvedValueOnce(mockResponse([post1]))
+      .mockResolvedValueOnce(mockResponse([post1])) // post
+      .mockResolvedValueOnce(mockResponse([])) // page
+      .mockResolvedValueOnce(mockResponse([])) // media
       .mockResolvedValueOnce(mockResponse([post1Revision]));
     // Fetch entities
     await store.actions.source.fetch("/post-1/?preview=true");
@@ -195,6 +197,8 @@ describe("post", () => {
     api.get = jest
       .fn()
       .mockResolvedValueOnce(mockResponse([post1]))
+      .mockResolvedValueOnce(mockResponse([]))
+      .mockResolvedValueOnce(mockResponse([]))
       .mockRejectedValueOnce(new ServerError("Forbidden", 403, "Forbidden"));
     // Fetch entities
     await store.actions.source.fetch("/post-1/?preview=true");
@@ -206,10 +210,12 @@ describe("post", () => {
     api.get = jest
       .fn()
       .mockResolvedValueOnce(mockResponse([post1]))
+      .mockResolvedValueOnce(mockResponse([]))
+      .mockResolvedValueOnce(mockResponse([]))
       .mockRejectedValueOnce(new ServerError("Forbidden", 403, "Forbidden"));
     // Fetch entities
     await store.actions.source.fetch("/post-1/?preview=true");
-    expect(api.get).toHaveBeenCalledTimes(1);
+    expect(api.get).toHaveBeenCalledTimes(3); // one time each for post, page & media
     expect(store.state.source).toMatchSnapshot();
   });
 });
--- a/packages/wp-source/src/libraries/handlers/postType.ts
+++ b/packages/wp-source/src/libraries/handlers/postType.ts
@@ -66,12 +66,32 @@ const postTypeHandler = ({
     // 1.2 iterate over finalEndpoints array
     let isHandled = false;
     let isMismatched = false;
-    for (const endpoint of finalEndpoints) {
-      const response = await libraries.source.api.get({
+
+    /**
+     * @param promises
+     */
+    // eslint-disable-next-line no-inner-declarations
+    async function* getData(promises: Promise<any>[]) {
+      const map = new Map(
+        promises.map((p, i) => [i, p.then((res) => [i, res])])
+      );
+      while (map.size) {
+        const [key, result] = await Promise.race(map.values());
+        yield result;
+        map.delete(key);
+      }
+    }
+
+    const promises = finalEndpoints.map((endpoint) =>
+      libraries.source.api.get({
         endpoint,
         params: { slug, _embed: true, ...state.source.params },
-      });
+      })
+    );
 
+    const data = getData(promises);
+
+    for await (const response of data) {
       const populated = await libraries.source.populate({
         response,
         state,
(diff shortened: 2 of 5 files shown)


packages/wp-source/src/libraries/handlers/__tests__/post-type.tests.ts
```diff
@@ -89,7 +89,7 @@ describe("postType", () => {
 describe("post", () => {
   test("doesn't exist in source.post", async () => {
     // Mock Api responses
-    api.get = jest.fn().mockResolvedValueOnce(mockResponse([post1]));
+    api.get = jest.fn().mockResolvedValue(mockResponse([post1]));
```
michalczaplinski, line 92: Previously the data was fetched serially for `post`, `page` and `media` (in that order). So, here we previously could assume that after we resolve the first `api.get()` (for post), there will be no more calls to `api.get()` as we already got the data on the first try.

This is not the case anymore as we're fetching data in parallel so `api.get()` will always be called three times no matter what.

The same change had to be done in a couple other places.

michalczaplinski:
@cristianbote Looks like the infinite scroll e2e tests are failing. I will look into this tomorrow but you can leave your comments in the meantime.

luisherranz:
I've run the e2e tests again to see if they pass the infinite scroll now.

It seems like the infinite scroll tests are not very reliable. Maybe the scroll support of Cypress is not good enough?

packages/wp-source/src/libraries/handlers/postType.ts
```diff
@@ -66,12 +67,21 @@ const postTypeHandler = ({
...
+      libraries.source.api.get({
         endpoint,
         params: { slug, _embed: true, ...state.source.params },
-      });
+      }),
+    ]);
 
+    for await (const [endpoint, response] of generateAsyncQueue(
+      promisesAndEndpoints
+    )) {
```
cristianbote, line 84: Hey @michalczaplinski this looks futuristic 😄  so bear with me and my limited knowledge. Since generators are not part of the browser runtime yet, was looking at how does the `generateAsyncQueue` looks after babel transpiles it(https://babeljs.io/repl#?browsers=and_chr%20%3E%3D%2067%2C%20and_ff%20%3E%3D%2018%2C%20and_uc%20%3E%3D%2011.8%2C%20android%20%3E%3D%2067%2Cnot%20android%20%3C%3D%204.4.4%2Cchrome%20%3E%3D%2049%2Cedge%20%3E%3D%2012%2Cfirefox%20%3E%3D%2018%2Cios_saf%20%3E%3D%2010%2Cnot%20op_mini%20all%2Cop_mob%20%3E%3D%2046%2Copera%20%3E%3D%2036%2Csafari%20%3E%3D%2010%2Csamsung%20%3E%3D%205&build=&builtIns=usage&spec=false&loose=false&code_lz=IYZwngdgxgBAZgV2gFwJYHsICoYHMCmE-ATsMvgILjQCKC-9AFAFAwwAOx6AtqiPiGYBKGAG9WMAPSSYAUQgALYNHwxkC1emKpcqCMAA2MYMVJgY6ODAAKXXvxAwAZDEIATduj3IQAOgnSMAAqCMQQMKjIERDI6MbhJmYWVsgI7AYCADQwAO4axKrqqnCoxCBR-BnchFGWrsBQCmppGRGCbIFFxqbA5npu-AAevsEaMPxQmG6uVTUwBSDoBgBuAmpxXanpmlYA2v1D2Vo6eoYASgIIBsgAuhKTEOUcdnwCFD3mALzPPK9-3MB2IxGOxsqgRJ8AHwwfbZdi7ACMN18RQgwIWEOhsI4uwADDdsgsbkJiQBuZgBGQAYQKZFUwBgACJPEtGckbC8HP42A8ngD2DBvkQcjAALKAkGct4fITkykwM4NekGIx1LqcX4OeLTUDjdCYNRjEplWpEeYCJarEDZMCoSrTIryhZXZCE_DcdCrQ2FFqqHKRJpdSamASeCBuPS4H72Yp2b3dJLKNxOpXGFXsroFAF6SPRv4wJBoIxdZn6gxsviubjsZBgbm5BSoVqMfm-ECoABe-BE4jYPMwT12AGt8GBsu5PN43SAXTdBcYcsBIhzNfhfKQoPgW4DfMtDPQQIwhLKJGxbfaYROvDFp7PyX2YK2BhlyIwR2AT2wAL7ML9AA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.12.14&externalPlugins=)

Got scared of the output, lol!

Anyways, my suggestion would be to stick to an `await Promise.all([])`. Would that be suitable for this case? For example something like this looks like it could be fitting:

```js
const results = await Promise.all(
  // Return a map of promises based on each endpoint
  finalEndpoints.map((endpoint) =>
    // The endpoint promise
    libraries.source.api.get({
      endpoint,
      params: { slug, _embed: true, ...state.source.params },
    }).then(res => (
      // Returning the tuple we need
      [endpoint, res]
     ))
  )
);

// Then we can walk over the results to populate
for (const [endpoint, response] of results) {
// ...
```

Again, I might not have the whole context here -- even though I went through the thread and the observablehq link -- but it looks like the `Promise.all` could be enough.

luisherranz (changes requested):
Hey Michal, could you please add a setting to opt-in to this?

Just so you know, we didn't do parallel requests in the first place because it's faster, but it is also less efficient.

The number of pages vs posts on big sites is usually very low. Pages are only for static content, like contact pages, landing pages, privacy, about us, and so on. When sites grow, they usually do it with posts.

A big site can have 100.000-500.000 posts, but it will rarely have more than 1000 pages.

The same happens with media/attachment pages. They exist, but they are rarely used and rarely linked.

Now, imagine we push this change to Frontity. A blog with 100.000 post requests per day will get 200.000 additional daily requests to their REST API to check for pages and media. That's not good. If they have 100 pages, only ~0.05% of those 200.000 extra requests will be useful. That's a lot of wasted bandwidth.

michalczaplinski:
@cristianbote 

> Hey @michalczaplinski this looks futuristic 😄

haha, yeah, that's the first time I have used this feature as well although it was part of ES2018 already! https://2ality.com/2016/10/asynchronous-iteration.html

It's a good point about babel, though! I didn't consider that it would add to the bundle size, although when I checked the difference is not huge: 5.8kb vs 3kb for the version which uses `Promise.all()`: https://babeljs.io/repl#?browsers=and_chr%20%3E%3D%2067%2C%20and_ff%20%3E%3D%2018%2C%20and_uc%20%3E%3D%2011.8%2C%20android%20%3E%3D%2067%2Cnot%20android%20%3C%3D%204.4.4%2Cchrome%20%3E%3D%2049%2Cedge%20%3E%3D%2012%2Cfirefox%20%3E%3D%2018%2Cios_saf%20%3E%3D%2010%2Cnot%20op_mini%20all%2Cop_mob%20%3E%3D%2046%2Copera%20%3E%3D%2036%2Csafari%20%3E%3D%2010%2Csamsung%20%3E%3D%205&build=&builtIns=usage&spec=false&loose=false&code_lz=BQQwzgngdgxgBMAlHAvAPjgbwFBzjAeyjABc4AnAUzAFcAbEsVOEAdxAEsyAFcggWw5hKAOhB06wXHjgB6WXABKlEjXJQWcfiAAOcAgDM4OvoOFMARuEoATfRsogYACziUoNnQQ5QS0vAY-4gCiHl4-jCLaOsDA7p7evsjo_jLycAAqzpRuYYlkJgJClKl4dBwW5CDkHNQiYARqMKK6HCIA5irAODK9uQkRADSlMjrVIPxgAFxYcGB0NO2DcAD6lPwWtjMk5DSUyyKHpCAkog1NomNVk3AAvsN9d4giJNlQwFRM6AgjeOnKqnUPnacFeOVUOjoOVYOSglFsvzgAG14uFfMtPgBdX6IRCpPF4RAAbmwt0QSCAA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.12.14&externalPlugins=

However, I think that the solution based on `Promise.all()` would not work here because of this:

Promise.all([p1, p2, p3]) waits for all of the promises `p1`, `p2` & `p3` to resolve before resolving with an array of those promises. So, in effect it would be making 3 requests in parallel and waiting for all of them to finish before continuing to process the data. 

In the approach that I proposed, as soon as one of the promises resolves, the `for await...of` loop will process the response and wait for the next promise to resolve, process it and then wait for the next one to resolve, etc. [Excalidraw here](https://excalidraw.com/#room=c3f779fa2000e2189c91,oKH7w_bZa6kgYyTTcwl9rA)

image

--- 

But in any case @luisherranz's point kinda makes this whole proposal DOA 😅 . 

@luisherranz Unless you think this is really worth giving a shot, I would close this PR.

luisherranz:
> @luisherranz Unless you think this is really worth giving a shot, I would close this PR.

Your take, Michal 🙂

The only thing that I ask you if you want to go ahead is to add it as an opt-in option to preserve the current behavior, especially for big sites.

packages/wp-source/src/libraries/handlers/postType.ts
cristianbote: https://github.com/frontity/frontity/pull/758#issuecomment-806196882

Continuing here, to keep the discussion on context.

I mean 😄  if we include the iteration part as well the sizes are a bit more night and day:
* [generateAsyncQueue](https://babeljs.io/repl#?browsers=and_chr%20%3E%3D%2067%2C%20and_ff%20%3E%3D%2018%2C%20and_uc%20%3E%3D%2011.8%2C%20android%20%3E%3D%2067%2Cnot%20android%20%3C%3D%204.4.4%2Cchrome%20%3E%3D%2049%2Cedge%20%3E%3D%2012%2Cfirefox%20%3E%3D%2018%2Cios_saf%20%3E%3D%2010%2Cnot%20op_mini%20all%2Cop_mob%20%3E%3D%2046%2Copera%20%3E%3D%2036%2Csafari%20%3E%3D%2010%2Csamsung%20%3E%3D%205&build=&builtIns=usage&spec=false&loose=false&code_lz=IYZwngdgxgBAZgV2gFwJYHsICoYHMCmE-ATsMvgILjQCKC-9AFAFAwwAOx6AtqiPiGYBKGAG9WMAPSSYAUQgALYNHwxkC1emKpcqCMAA2MYMVJgY6ODAAKXXvxAwAZDEIATduj3IQAOgnSMAAqCMQQMKjIERDI6MbhJmYWVsgI7AYCADQwAO4axKrqqnCoxCBR-BnchFGWrsBQCmppGRGCbIFFxqbA5npu-AAevsEaMPxQmG6uVTUwBSDoBgBuAmpxXanpmlYA2v1D2Vo6eoYASgIIBsgAuhKTEOUcdnwCFD3mALzPPK9-3MB2IxGOxsqgRJ8AHwwfbZdi7ACMN18RQgwIWEOhsI4uwADDdsgsbkJiQBuZgBGQAYQKZFUwBgACJPEtGckbC8HP42A8ngD2DBvkQcjAALKAkGct4fITkykwM4NekGIx1LqcX4OeLTUDjdCYNRjEplWpEeYCJarEDZMCoSrTIryhZXZCE_DcdCrQ2FFqqHKRJpdSamASeCBuPS4H72Yp2b3dJLKNxOpXGFXsroFAF6SPRv4wJBoIxdZn6gxsviubjsZBgbm5BSoVqMfm-ECoABe-BE4jYPMwT12AGt8GBsu5PN43SAXTdBcYcsBIhzNfhfKQoPgW4DfMtDPQQIwhLKJGxbfaYROvDFp7PyX2YK2BhlyIwR2AT2wAL7MH_MXlRBqMYgBQ4byB414-POJT6AY4GTjE_wSowV7eJiMISKhN4SAYqAAEakNoAhtugoSbr4gKoL4BDIIwvZ9lhrqnhwJjANwIAAFxiOMBgILg2QAPrunh-BuFxyDEPQ2S-DJ5R0iRZFruwrHsTAX6ZBIX5CBpZIUowoCQLAR6CtC9EAJBwFoC5LlEjAAZe4YIa65ogGG_BznUBBEKQ5BUIZdAMFuQF_KBbjwZBIDHmIP5aaSQA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=false&presets=env&prettier=false&targets=&version=7.12.14&externalPlugins=) 10.3 kB
* [Promise.all](https://babeljs.io/repl#?browsers=and_chr%20%3E%3D%2067%2C%20and_ff%20%3E%3D%2018%2C%20and_uc%20%3E%3D%2011.8%2C%20android%20%3E%3D%2067%2Cnot%20android%20%3C%3D%204.4.4%2Cchrome%20%3E%3D%2049%2Cedge%20%3E%3D%2012%2Cfirefox%20%3E%3D%2018%2Cios_saf%20%3E%3D%2010%2Cnot%20op_mini%20all%2Cop_mob%20%3E%3D%2046%2Copera%20%3E%3D%2036%2Csafari%20%3E%3D%2010%2Csamsung%20%3E%3D%205&build=&builtIns=usage&spec=false&loose=false&code_lz=BQQwzgngdgxgBMAlHAvAPjgbwFBzjAeyjABc4AnAUzAFcAbEsVOEAdxAEsyAFcggWw5hKAOhB06wXHjgB6WXABKlEjXJQWcfiAAOcAgDM4OvoOFMARuEoATfRsogYACziUoNnQQ5QS0vAY-4gCiHl4-jCLaOsDA7p7evsjo_jLycAAqzpRuYYlkJgJClKl4dBwW5CDkHNQiYARqMKK6HCIA5irAODK9uQkRADSlMjrVIPxgAFxYcGB0NO2DcAD6lPwWtjMk5DSUyyKHpCAkog1NomNVk3AAvsN9d4giJNlQwFRM6AgjeOnKqnUPnacFeOVUOjoOVYOSglFsvzgAG14uFfMtPgBdX6IRCpPF4RAAbmk0gMBHICEIxDIKLyQwo1C8xEomP0Rk-9EYyEwt2wt0QSCAA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.12.14&externalPlugins=) 5.9 kB

But, more related to the async queue processing, thank you so much for explaining it.

My follow-up question is: is there a benefit for the end result to use an async queue? More exactly would the execution go beyond the `for async` loop? I don't think so, but I might be wrong. If that's the case, is there a practical reason to asynchronous populate the `.data`?

michalczaplinski: > I mean 😄 if we include the iteration part as well the sizes are a bit more night and day:
>
> generateAsyncQueue 10.3 kB
> Promise.all 5.9 kB

You're right about that, I didn't consider that! 🙈 

> My follow-up question is: is there a benefit for the end result to use an async queue? More exactly would the execution go beyond the for async loop? I don't think so, but I might be wrong. If that's the case, is there a practical reason to asynchronous populate the .data?

I'm not sure I fully understand the question, in particular I don't know what you mean with "execution go beyond the for async loop?". But let me try to answer what I *think* you mean.

Using the async queue we load data from each of the 3 endpoints in parallel: `post`, `page`, `media`. However, if we were to use `Promise.all`, the data would also load in parallel. So what is the difference? 

If we use the async queue, then we don't have to wait for all 3 requests to resolve before continuing to execute the `postType` handler. So, it's not that there is a benefit to populating `.data` asynchronously but rather that there **is** a benefit to being able to break out of the loop early [here](https://github.com/frontity/frontity/blob/0a1bab4ee07de1eb680bc06d1734d00ed15afc25/packages/wp-source/src/libraries/handlers/postType.ts#L91-L95) **as soon as one of the promises resolves** and this `populated.length > 0` condition is true.

michalczaplinski:
Closing this in the light of Luis's comment: https://github.com/frontity/frontity/pull/758#pullrequestreview-619482575. 

While this approach is technically "faster", in practice it is going to be more wasteful (especially for large sites) because most of them have much fewer pages than posts.

PR #764, opened 2021-03-23
Update the call order of `afterSSR()` actions in `frontity/core`

michalczaplinski:
**What**:

Update the call order of `afterSSR()` actions in `frontity/core`

**Why**:

We need to call `afterSSR()` **before** taking the state snapshot so that the action can still modify the state before that snapshot is taken.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] End to end tests

**Unrelated Tasks**

- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


luisherranz:
I've run the e2e tests again to see if they pass the infinite scroll now.

It seems like the infinite scroll tests are not very reliable. Maybe the scroll support of Cypress is not good enough?

packages/core/src/server/middlewares/server-side-rendering.ts
```diff
@@ -67,6 +84,11 @@ export const serverSideRendering = async (
     // https://github.com/smooth-code/loadable-components/pull/239#issuecomment-482501467
     const customExtractor = extractor as Extractor;
 
+    // Run afterSSR actions. It runs at this point because we want to run it
+    // before taking the state snapshot. This gives the user a chance to
+    // modify the state before sending it to the client
+    runAfterSSRActions(store, ctx);
```
cristianbote: Do we have to `await` here? 🤔  I think so? Same thing in the other case.
```suggestion
    await runAfterSSRActions(store, ctx);
```

luisherranz:
> I've run the e2e tests again to see if they pass the infinite scroll now

They passed this time 🤷

packages/core/src/server/middlewares/server-side-rendering.ts
michalczaplinski: Oh, you're right! For some reason I thought it should not be necessary to await here if I await **inside** each of the `afterSSR()` calls but that's not the case of course.

michalczaplinski:
@cristianbote Updated to `await runAfterSSRActions()` like you pointed out. Feel free to merge after review :)

cristianbote:
> > I've run the e2e tests again to see if they pass the infinite scroll now
> 
> They passed this time 🤷

And then failed next time they ran. 🤔  Triggered them again.

cristianbote (approved):

PR #765, opened 2021-03-24
[Html2React] Create an HTML2JSON parser using DOMParser

DAreRodz:
**What**:

Modify the `@frontity/html2react` package to use [`DOMParser`](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) on the browser instead of adding [Himalaya](https://www.npmjs.com/package/himalaya) to the main bundle.

**Why**:

To test if that change improves some performance metrics like TTI.

**How**:

Define a `libraries.html2react.parse` function and change the implementation using client and server entry points.

**Tasks**:

- [x] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Unrelated Tasks**

**Additional Comments**


luisherranz:
@darerodz what is the status of this? Is it worth keeping it open?

DAreRodz:
> What is the status of this? Is it worth keeping it open?

Hey, this was part of a research I worked on a while ago to check if there could be any improvement using the `DOMParser` API instead of bundling the Himalaya library along with the Frontity app. What I wanted to test was:

* How much we could reduce the bundle size
* How much we could gain in performance
 
I don't have the specific numbers (I should have taken some notes/screenshots, and I didn't :see_no_evil:) but if my memory is correct, these are some insights I can share:
 
* The bundle size decreased, but not significantly.
* Metrics like TTI weren't almost affected.
* Execution of the `libraries.html2react.parse` function using `DOMParse` were consistently **slower** than the Himalaya implementation, except for the first call, that took almost the same for both implementations. My hypothesis here is that the first time a function in JS runs, it should be compiled as well, taking longer to execute. Once compiled, it takes less time, and I suppose `DOMParse` has a more complex implementation compared with Himalayas's. In any case, we're talking about a difference of ~20ms in 6x, barely noticeable in the real world.

So, yeah, I think we can close this. Maybe just update the PR's description to include what I've written here, but it's not worth it to  keep working on this research.

luisherranz:
Perfect, thanks for the update David!! 

Closing now.

PR #767, opened 2021-03-25
Do not return a 30x which redirects to the same link

### LINKED ISSUE #756: Do not return a 301 which redirects again to the same link [8pt]
**_Bug report_**

**Expected behavior**

If there is a redirect to the same link, Frontity shouldn't do that redirect to prevent an infinite loop.

**Observed behavior**

This could happen right now if the domains are different and the WordPress domain is redirected to the Frontity domain.

image

**Possible solution**

The logic should be something like this: if the redirection is internal and the `pathname` is the same, don't do the redirection.

```js
const linkParams = new URL(link);
const locationParams = new URL(location);

const linkURI = linkParams.path + linkParams.search;
const locationURI = locationURI.path + locationURI.search;

// Only return the redirection if the location is different.
if (!isExternal && linkURI !== locationURI) {
  return redirection;
}
```

This code hasn't been tested, it's just an example.

michalczaplinski:
**What**:

#756. If there is a redirect to the same link, Frontity shouldn't do that redirect to prevent an infinite loop. This would previously happen if the domains were different and the WordPress domain redirects to the Frontity domain.

**How**:

By implementing a `shouldBailRedirecting()` function which contains the logic to determine whether we want to bail out of redirecting.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] End to end tests
**Unrelated Tasks**

- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


cristianbote (approved):
Great work @michalczaplinski!

PR #770, opened 2021-03-31
Modify <Script> components, to add any other passed props to the script

### LINKED ISSUE #607: Pass `<script>` attributes to the element created by `<Script>`
**_Bug report_**

**Expected behavior**

A `<Script>` component should pass any prop (attribute) to the element created.

This code:

```js
<Script
  async
  data-ad-client="ca-pub-2523307345324148497"
  src="https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js"
></Script>
```

Should create this element:

```js
const script = window.document.createElement("script");
script.src = "https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js";
script.async = true;
script.dataset.adClient = "ca-pub-2523307345324148497";
```

**Observed behavior**

`script.dataset.adClient` is not populated.

Create that `<Script>` and inspect the final `<script>` created by it in the Inspector.

**Possible solution**

Iterate over the `rest` props:

```js
const Script: React.FC<ScriptProps> = ({ src, code, ...rest }) => {
```

and add them to the element created.

We have to take into account that `data-` attributes go to the `element.dataset` object and that they are camelCased: https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset

Koli14:
**What**:

Extended the <Script> component, so its append any other props passed to it to the script.

**Why**:

It is useful, if we can pass e.g. an `onload`, `defer` or `crossOriging` props.
See related [discussion](https://community.frontity.org/t/import-a-javascript-sdk-and-then-use-it/4302/7?u=koli14) discussion too.

**How**:

I extracted other props with the spreed operator, and the used `for-of`with `Object.keys` on these `props`.

**Tasks**:

- [x]  Code
- [x] TSDocs
- [x] TypeScript
- [x] End to end tests

**Unrelated Tasks**

- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**
Sorry, but I was unsure, how to write the change in the TSDocs and TypeScript. I also don't know, how to write tests. And I was also unsure about the changeset.


luisherranz:
Thanks for the PR @koli14! 🙂

@darerodz: can you take a look? Thanks!

DAreRodz:
I've changed the title and linked the issue this PR will close once it's finished.

packages/components/script.tsx
```diff
@@ -64,6 +65,11 @@ const Script: React.FC<ScriptProps> = ({ src, code, id }) => {
       // Add the ID if specified.
       if (id) script.id = id;
 
+      // Add any other props to the internal <script> tag.
+      for (let key of Object.keys(props)) {
```
DAreRodz: Here you can use `for ... in` to iterate over all the keys directly.

```suggestion
      for (const key in props) {
```

packages/components/script.tsx
```diff
@@ -64,6 +65,11 @@ const Script: React.FC<ScriptProps> = ({ src, code, id }) => {
...
+        script[key] = props[key];
```
DAreRodz: This could fail when assigning event handlers and data attributes.

My guess is that using [`Element.setAttribute(key, value)`](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute) here would be a better option, as it would:

- lowercase attributes automatically (i.e `crossOrigin` → `crossorigin`)
- handle data attributes:
  ```js
  script.setAttribute("data-ad-client", "ca-pub-123");
  script.dataset.adClient; // This returns "ca-pub-123".

The only pitfall is that `className` would need to be handled apart.

```suggestion
        script.setAttribute(key === "className" ? "class" : key, props[key]);
```

DAreRodz (changes requested):
Thanks for the PR @Koli14 :raised_hands:

This is actually a bugfix instead of a feature, there's been an issue opened for some time already, hehe (https://github.com/frontity/frontity/issues/607). I requested some changes, check it out.

PS: Don't worry about TSDocs, TypeScript and tests, we can help you with that. :slightly_smiling_face:

packages/components/script.tsx
cristianbote: There might be a bit more to it other than a straight lowercase and `setProperty`. For example functions would need to be assigned. Think of `onLoad` would need to be assigned to `onload` and so on.

There's a way to handle both of them like this:

```js
for (let key in props) {
  const value = props[key];

  // If the current key exists in the `dom` interface
  // we can assign the value.
  if (key in script) {
    // If this is an event handler, lowercase the key
    if (/^on/g.test(key)) {
      key = key.toLowerCase();
    }
    script[key] = value;
  } else if(typeof value !== 'function' && typeof value !== 'object') {
    // Otherwise treat it as an attribute if this is not a function or an object.
    script.setAttribute(key, value);
  }
}
```

For reference this is the exact technique used by any other vdom libraries out there. Fair warning 😅  I wrote it by hand so there might be some execution errors in there.

DAreRodz: Huh, interesting. :thinking:

I thought that using `element.setAttribute()` would handle that case just fine. I mean, if you run

```js
script.setAttribute("onLoad", () => console.log("script loaded!));
```

you're actually setting the `onload` attribute, not the `onLoad` one.

Or is this something that doesn't work the same for all browsers/versions?

cristianbote: Unfortunately `.setAttribute` would stringify the value of the attribute. So the code above would actually become:

```html
<script onload="() => console.log('script loaded!')"></script>
```

Which is not what we want to achieve. We want to add a `onload` listener instead on the element DOM reference, which needs to be set through the DOM properties not attributes.

DAreRodz: Oh, I see, thanks for the clarification. :blush: :+1:

Koli14: Thanks for the helps and suggestions!
Now `className`, `onLoad` and similars are working. Data attributes can be set like `data-value='3'`
An example:
```
<Script
    id='myScript'
    data-value='3'
    onLoad={() => {console.log('Scipt Loaded')}}
    className='coolScript'
    defer=''
    crossOrigin='anonymous'
    src='//some.cdn.com/some/script.js'
  />`
```

luisherranz: Could you please add unit tests for all those props? 😄

- Some random prop.
- The `className` prop.
- Data attributes.
- The `onLoad` prop.
- ...

Those tests are really simple. We use something like this:

```js
test("should work with data attributes", () => {
  const ScriptTest = create(<Script data-value="3" />);
  expect(ScriptTest.toJSON()).toMatchSnapshot();
});
```

You can use this tests of `<Switch>` as a reference to set them up: https://github.com/frontity/frontity/blob/dev/packages/components/__tests__/switch.tests.tsx

Thanks!!

Koli14: I tried to create the test cases, but in the .snap file all the cases are `null`.
I could not figure it out, why.

Koli14: How can I test, if `<script>` tag is added to the body? react-test-renderer `create` will not work, as the `<Script>` component return `null`.

luisherranz: Ohh, yes, that's right.

I've moved them to the e2e tests and they pass 🙂 (https://github.com/frontity/frontity/pull/770/commits/75553ade3d1e052d1443203565bfb7c5180565ee).

luisherranz: Oh, I removed the `onLoad` test. As we are creating the `<script>` tags after hydration I don't think it makes sense. And I guess that if someone wants to use it they can always use `<script>` instead of `<Script>`.

Koli14:
Is there anything, I still need to do? Or it can be merged?

luisherranz:
@darerodz, are you taking care of this?

DAreRodz (approved):
Yup, I've just added a test for event handlers. Everything OK now. :ok_hand:

EDIT: I'll merge the PR once the tests have passed.

DAreRodz:
The tests that have failed are those that don't work with external PRs so I merged it anyway.

Thanks again @Koli14 for your contribution. :blush:

luisherranz:
> The tests that have failed are those that don't work with external PRs so I merged it anyway

That should be fixed from now on (https://github.com/frontity/frontity/pull/787) 🙂

PR #771, opened 2021-04-05
Skip the redirections for embedded mode

### LINKED ISSUE #743: Infinite loop with 404 redirections in Embedded mode in some hostings [13pt]
**_Bug report_**

**Expected behavior**

While using the [redirects feature](https://community.frontity.org/t/301-redirects-stored-in-wordpress-database/3032), with the 404 option enabled and the Embedded mode, it should work as normal.

**Observed behavior**

Some hostings are changing the HEAD methods and use GET instead, and this is causing an infinite loop if you are using the Embedded mode and you visit a page that doesn't exist.

**Possible solution**

Right now, the most appealing solution seems to be to send a query from the Embedded mode (`frontity_embedded=true` for example), that allows Frontity to populate a setting in the state to know if the app is running in Embedded mode and, if it is, don't try the redirections at all in SSR, because using the Embedded mode it will be redirected directly by WordPress.

We discussed this issue and the possible solutions in this video:

https://www.youtube.com/watch?v=ZQekma-77VY

cristianbote:
**What**:

Handle the redirections for embedded mode.

**Why**:

To avoid the infinite loop for redirections handled by wordpress instance.

**How**:

The frontity embedded plugin sends the `frontity_embedded=true` option so we can detect the redirections that are coming from it and skip them.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests
- [x] TypeScript tests
- [x] Update other packages

**Unrelated Tasks**

- [ ] TSDocs
- [ ] Update starter themes

**Additional Comments**

~Opened as draft as I wanna confirm first that this is the right direction.~ Done

### DIFF
--- a/.changeset/early-eyes-add.md
+++ b/.changeset/early-eyes-add.md
@@ -1,0 +1,5 @@
+---
+"@frontity/wp-source": patch
+---
+
+Bail out of redirection if frontity is running into embedded mode.
--- a/e2e/integration/wordpress-01/redirections.spec.ts
+++ b/e2e/integration/wordpress-01/redirections.spec.ts
@@ -62,19 +62,6 @@ describe("Redirections", () => {
     cy.get("#link-counter").should("contain.text", "3");
   });
 
-  it("Should not redirect when navigating on the client in embedded mode", () => {
-    cy.visit(
-      "http://localhost:3001?frontity_name=redirections&frontity_embedded=true"
-    );
-
-    // Go to the "redirected" page.
-    cy.get("#open-post").click();
-
-    cy.location("href").should("eq", "http://localhost:3001/hello-world/");
-    cy.get("#post").should("not.exist");
-    cy.get("#link-counter").should("contain.text", "2");
-  });
-
   it("The back and forward button should work fine when navigating", () => {
     cy.visit("http://localhost:3001?frontity_name=redirections");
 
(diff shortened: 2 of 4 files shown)


luisherranz (changes requested):
Code looks great, although I would add: 
- A unit test for `state.frontity.options.embedded = true` and `state.source.redirections = "all"`.
- A unit test for `state.frontity.options.embedded = true` and `state.source.redirections = "404"`.
- An e2e test for the server using `?frontity_embedded=true` and `state.source.redirections = "all"`.
- An e2e test for the server using `?frontity_embedded=true` and `state.source.redirections = "404"`.
- An e2e test for the client using `?frontity_embedded=true` and `state.source.redirections = "all"`.
- An e2e test for the client using `?frontity_embedded=true` and `state.source.redirections = "404"`.

cristianbote:
> Code looks great, although I would add:
> 
> * A unit test for `state.frontity.options.embedded = true` and `state.source.redirections = "all"`.
> * A unit test for `state.frontity.options.embedded = true` and `state.source.redirections = "404"`.
> * An e2e test for the server using `?frontity_embedded=true` and `state.source.redirections = "all"`.
> * An e2e test for the server using `?frontity_embedded=true` and `state.source.redirections = "404"`.
> * An e2e test for the client using `?frontity_embedded=true` and `state.source.redirections = "all"`.
> * An e2e test for the client using `?frontity_embedded=true` and `state.source.redirections = "404"`.

Oh Luis, really sorry, but apparently my last commit wasn't pushed. And that commit had the unit and e2e tests. But reading your comment above, it seems that I should tackle the `404` codepath as well. So, gonna add a few more tests for `404`. 👍

luisherranz (approved):
Looks perfect! Thanks Cris 👏👏

PR #775, opened 2021-04-09
Use a relative path with non-external redirections

### LINKED ISSUE #728: Redirections don't send you to localhost while running Frontity in dev mode
**_Bug report_**

**Expected behavior**

If you visit a URL that is redirected, and that redirection is not to an external domain, it should maintain the same domain and just change the rest of the URL.
url
For example, the following data object:

```ts
state.source.data["/old-post/"] = {
  isFetching: false,
  isReady: true,
  isRedirection: true,
  is301: true,
  redirectionStatus: 301,
  location: "/new-post/",
}
```

should redirect you from http://localhost:3000/old-post/ to http://localhost:3000/new-post/.

**Observed behavior**

It uses the value of `state.frontity.url`, i.e. the domain of the Frontity site in production.

So, for example, if `state.frontity.url` is `https://my.frontity.site`, then Frontity redirects you from http://localhost:3000/old-post/ to https://my.frontity.site/new-post/.

**Possible solution**

After consulting the [Koa documentation](https://koajs.com/#response-redirect-url-alt-), the [Location header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location#directives) docs from MDN (even the [HTTP](https://tools.ietf.org/html/rfc7231#page-68) standard), I've concluded it is safe to use a relative URL as the Location header.

We could just remove [this line](https://github.com/frontity/frontity/blob/dev/packages/tiny-router/src/actions.ts#L255) and it would work as expected, also in the production environment.

DAreRodz:
**What**:

Turn the absolute URL included in the `location` prop of data objects into a relative URL, just for non external redirections (i.e. links pointing to the same domain).

**Why**:

To fix https://github.com/frontity/frontity/issues/728.

**How**:

Removing the line that was adding `state.frontity.url` to the generated redirection URL.

**Tasks**:

- [x] Code
- [x] Unit tests
- [x] End to end tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/e2e/packages/redirections/src/index.tsx
+++ b/e2e/packages/redirections/src/index.tsx
@@ -169,6 +169,10 @@ const redirections: Redirections = {
           state.source.redirections = query.redirections;
         }
 
+        if (state.frontity.options.url) {
+          state.frontity.url = state.frontity.options.url;
+        }
+
         const handler = {
           pattern: "/urls-with-redirections/:slug",
           func: async ({ state, link }) => {
--- a/.changeset/sour-wolves-boil.md
+++ b/.changeset/sour-wolves-boil.md
@@ -1,0 +1,5 @@
+---
+"@frontity/tiny-router": patch
+---
+
+Make redirection URLs relative, so they redirect to `localhost` in a dev environment.
--- a/e2e/integration/wordpress-01/redirections.spec.ts
+++ b/e2e/integration/wordpress-01/redirections.spec.ts
@@ -20,6 +20,19 @@ describe("Redirections", () => {
     cy.get("#link-counter").should("contain.text", "1");
   });
 
+  it("Should redirect to the requested domain, not to `state.frontity.url`", () => {
+    cy.visit(
+      "http://localhost:3001/hello-world/?frontity_name=redirections&frontity_url=http://my.frontity.site"
+    );
+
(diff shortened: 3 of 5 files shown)


e2e/packages/redirections/src/index.tsx
```diff
@@ -169,6 +169,10 @@ const redirections: Redirections = {
           state.source.redirections = query.redirections;
         }
 
+        if (state.frontity.options.url) {
+          state.frontity.url = state.frontity.options.url;
```
michalczaplinski, line 173: nice trick. I worry sometimes that the Frontity Options can be an attack vector if someone (god forbid) `evals` the contents of a frontity option or passes it as a parameter to a `fetch()` call etc.

I ll ping DevRel to explicitly warn about this in the docs.

michalczaplinski (approved):
Great! 👍 Thanks @DAreRodz! 🙂

e2e/packages/redirections/src/index.tsx
DAreRodz: Good idea :+1:

PR #776, opened 2021-04-09
Fix AMP validation errors

### LINKED ISSUE #772: AMP validation errors
**_Bug report_**

**Expected behavior**

All the pages in test.frontity.org should be AMP valid.

**Observed behavior**

There are some errors in some posts that could affect other websites as we have been discussing in the [community thread](https://community.frontity.org/t/amp-package/388/66?u=santosguillamot).

Install AMP and validate the urls as explained in [this other post](https://community.frontity.org/t/amp-package/388/65?u=santosguillamot).

michalczaplinski:
**What**:

Fix the validation errors for AMP described in https://community.frontity.org/t/amp-package/388/66

**Why**:

To fix #772

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests

**Unrelated Tasks**

- [ ] Update starter themes
- [ ] Update other packages
- [ ] TypeScript tests


packages/amp/src/__tests__/processors.test.tsx
```diff
@@ -65,13 +85,45 @@ test("Validate amp-iframe", async () => {
...
+        processors={processors}
+      />
+    </HelmetProvider>
+  );
+
+  expect(container.firstChild).toMatchInlineSnapshot(`
+    <amp-iframe
+      height="300"
+      layout="fixed-height"
+      sandbox="allow-scripts allow-same-origin allow-scripts allow-popups"
```
michalczaplinski, line 119: We can concatenate the all the `allow-*` flags here and even if some property appears more than once it does not matter because it's set only once.

packages/amp/src/__tests__/processors.test.tsx
```diff
@@ -115,7 +171,6 @@ test("Validate amp-audio", async () => {
   );
   expect(container.firstChild).toMatchInlineSnapshot(`
     <amp-audio
-      controls="true"
```
michalczaplinski, line 118: Previously we always passed a `controls` prop to `amp-audio` but we should not do that. The content should set that attribute.

michalczaplinski:
ok, ready for review, here's a summary:

https://www.loom.com/share/7725ecc239094e808b0975bd85ba7a71

michalczaplinski:
Oh and additional note on validating the html with AMP validator (like described in https://community.frontity.org/t/amp-package/388/65)

https://www.loom.com/share/033dbd36a53342579a571fd101933b4e

SantosGuillamot (approved):
Really nice work Michal! :clap: Regarding the functionality, everything is working now and the AMP errors left are content related, so that's fine. I agree with you that it's fine to not include the way we validated test.frontity.org in the e2e tests because it's a bit complex.

Although I've already taken a look, I think it would be nice if @cristianbote could take a really quick look at the code to ensure there aren't syntax issues.

cristianbote (approved):
The code looks fine to me! 👍 

One small nit: I believe we should warn the user when they are trying to load http instead of https. By automatically handling the http-to-https source transform, we are shielding them from the AMP requirements. Might be more valuable to be informed. What do you think @michalczaplinski? @SantosGuillamot?

michalczaplinski:
@cristianbote Hmm, I think that's a good point. I've initially included warnings but then removed them in [this commit](https://github.com/frontity/frontity/pull/718/commits/9a4f2d82ef18408ac7ca757fc82fcc8435e8eedd) due to [Luis's comment](https://github.com/frontity/frontity/pull/718#issuecomment-793749352). 

However, if we're configuring the AMP site [using the `match` property like described here](https://community.frontity.org/t/amp-package/388/27?u=mmczaplinski), then the AMP processors are not going to run on the non-AMP site so we can add warnings without worrying.

michalczaplinski:
Merging this after having added the warnings per Cris's comment.

luisherranz:
> However, if we're configuring the AMP site using the match property like described here, then the AMP processors are not going to run on the non-AMP site so we can add warnings without worrying.

That's absolutely right. Sorry for the mistake 🙂

PR #777, opened 2021-04-12
Switch to modulepreload hint for module files

### LINKED ISSUE #740: JS modules with `<link rel="preload">` are requested twice
**_Bug report_**

This bug was noticed by @SantosGuillamot.

**Expected behavior**

A preloaded module should be requested once.

**Observed behavior**

It's requested twice: one from the `<link rel="preload">` tag and another from the `<script>` tag.

Screenshot from 2021-03-08 12-04-21

For each script, the following warning is shown in the console:

```
A preload for '/static/mars-theme.module.eaa9243fd3abc5a42c99.js' is found, but is not used because the script type does not match.
```

It should happen with any production build. Tested with https://mars.frontity.org in the following browsers (Ubuntu 20.10):
 - Chrome: 89.0.4389.72
 - Firefox: 86.0

**Possible solution**

It turns out that we should use [`rel="modulepreload"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/modulepreload) instead of `rel="preload"` for js modules.

For doing so, we could add a `.replace()` call to the output of `extractor.getLinkTags()` in this line (`extractor` is an instance of `ChunkExtractor`, from the `@loadable/server` library):

https://github.com/frontity/frontity/blob/a9ee85ffbb851dc819d2c693385bbd8b6f1d4b1b/packages/core/src/server/middlewares/server-side-rendering.ts#L59-L60

It could be something like:

```ts
// Push the link tags.
    output.head.push(
      extractor
        .getLinkTags(crossorigin)
        .replace(
          /( rel=")preload(" as="script" href="\S+\.module\.[\da-f]+\.js")/g,
          "$1modulepreload$2"
        )
    );
```

The regexp is a bit weird because I wanted to ensure that only preloaded modules are matched. I don't know if that's really needed, tho.

This kind of "hackaround" is already proposed by the Loadable team for a different issue: https://github.com/gregberge/loadable-components/issues/653#issuecomment-726667065

cristianbote:
**What**:

In light of recent issues with Chrome warning about using `preload` instead of `modulepreload` we need to switch to it for our compiled modules assets.

**How**:

Instead of a simple replace, I went ahead and extended the `ChunkExtractor` class and overwrite the `getPreAssets` method that returns the assets that need to be preloaded. That way the detection is isolated to that codepath only and does not intervene with the rest of the server side rendering logic.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] TypeScript tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] Update starter themes
- [ ] Update other packages


cristianbote:
FYI: The tests are failing because the `amphtml-validator` project does not pick up the `modulepreload` as a valid value. The project got updated recently to include the `modulepreload` assertion https://github.com/ampproject/amphtml/pull/33372 but the updated version hasn't been published yet.

luisherranz:
> FYI: The tests are failing because the amphtml-validator project does not pick up the modulepreload as a valid value.

Ou! I guess that is a bug because those preloads should not appear in the HTML of an AMP page.

Is there an easy way to turn them off from the `@frontity/amp` package?

packages/core/src/server/utils/custom-chunk-extractor.ts
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * The `getPreAssets` method.
+   *
+   * @returns A list of assets.
+   */
+  getPreAssets() {
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    const assets = (ChunkExtractor as Function).prototype.getPreAssets.call(
+      this
+    );
```
DAreRodz: Hey, couldn't [`super`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super) be used here to call that method in the parent's prototype chain? I think it's possible to write something like this:

```suggestion
    const assets = super.getPreAssets();
```

cristianbote:
> > FYI: The tests are failing because the amphtml-validator project does not pick up the modulepreload as a valid value.
> 
> Ou! I guess that is a bug because those preloads should not appear in the HTML of an AMP page.
> 
> Is there an easy way to turn them off from the `@frontity/amp` package?

Not sure if it's a bug though, cause the scripts that are loaded are the amp related packages that need to include JavaScript.

packages/core/src/server/utils/custom-chunk-extractor.ts
cristianbote: Oh indeed that's possible. Need to check it locally first though. Have you tried it? I'm worried that the `ChunkExtractor` ends up being exported from the package as a function not as a `class` so maybe the `super.` syntax would not work in that case?

cristianbote: Yeah, no need to worry about it it seems it's a `class` [look here](https://unpkg.com/browse/@loadable/server@5.14.2/lib/ChunkExtractor.js)

luisherranz:
> Not sure if it's a bug though, cause the scripts that are loaded are the amp related packages that need to include JavaScript.

From webpack chunks?

packages/core/src/server/utils/custom-chunk-extractor.ts
DAreRodz: I didn't try it but I had a gut feeling (from my old OOP background, I guess :smile:).

cristianbote:
> > Not sure if it's a bug though, cause the scripts that are loaded are the amp related packages that need to include JavaScript.
> 
> From webpack chunks?

You are right. The rule for removing scripts was not matching the preload link tag because the e2e project is named `amp-` which is the same as the convention AMP uses for it's JavaScript includes for the custom elements. Fixed that.

luisherranz:
Great 🙂

packages/core/src/server/utils/custom-chunk-extractor.ts
cristianbote: Alrighty! Fixed it @DAreRodz! Thank you!

cristianbote:
@DAreRodz and @luisherranz this is ready, types have been published so no TS complaints anymore 😄

packages/amp/src/index.tsx
```diff
@@ -101,7 +101,7 @@ const amp: AMP = {
               // Else capture all the script and link with as="script" tags.
               else if (/<?script/g.test(tag)) {
                 // And if they contain the `amp-` specific markers keep them.
-                if (/<?script.+?amp-/g.test(tag)) {
+                if (/<?script.+?="amp-/g.test(tag)) {
```
DAreRodz, line 104: Just to be sure - this would allow only those scripts with `custom-element="amp-..."`, right?

What would happen with the main AMP library?

```html
<script async src="https://cdn.ampproject.org/v0.js"></script>
```

DAreRodz (commented):
Everything looks great, although I have a question :point_down:

packages/amp/src/index.tsx
cristianbote: The main script is part of the template, as far as I remember, not part of the dynamic head scripts.

Here: https://github.com/frontity/frontity/blob/dev/packages/amp/src/template.ts#L19

DAreRodz (approved):
Approved! :heavy_check_mark:

DAreRodz:
Feel free to merge this, @cristianbote :blush:

cristianbote:
Thanks for the reminder @DAreRodz! 😄

PR #778, opened 2021-04-12
Collapse empty ad slots 

VinuRaj:
**What**:

To collapse empty ad slots in Frontity Google Ad manager package.

**Why**:

By default, ad slots which are not filled are left visible, which may result in blank space on page.

Reference - https://developers.google.com/publisher-tag/samples/collapse-empty-ad-slots

**How**:

We could avoid that by calling `slot.current.setCollapseEmptyDiv(true);` on each slot.  

**Tasks**:

- [x] Code
- [x] TypeScript

**Unrelated Tasks**

- [ ] TSDocs
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

### DIFF
--- a/.changeset/thirty-readers-smile.md
+++ b/.changeset/thirty-readers-smile.md
@@ -2,4 +2,5 @@
 "@frontity/google-ad-manager": patch
 ---
 
-Collapse ad slots that are not filled by Google Publisher Tag. Until now, slots would always take up space on the page as defined by its fill even if the ad call fails to fill that slot with an ad.
+The `GooglePublisherTag` component now accepts an optional `collapseEmptyDiv` prop
+which if `true` will collapse ad slots that are not filled by Google Publisher Tag.
--- a/.changeset/thirty-readers-smile.md
+++ b/.changeset/thirty-readers-smile.md
@@ -1,0 +1,5 @@
+---
+"@frontity/google-ad-manager": patch
+---
+
+Collapse ad slots that are not filled by Google Publisher Tag. Until now, slots would always take up space on the page as defined by its fill even if the ad call fails to fill that slot with an ad.
--- a/packages/google-ad-manager/src/components/google-publisher-tag.tsx
+++ b/packages/google-ad-manager/src/components/google-publisher-tag.tsx
@@ -45,6 +45,8 @@ const GooglePublisherTag: React.FC<
       Object.entries(targeting).forEach(([key, value]) =>
         slot.current.setTargeting(key, value)
       );
+      // Collapse empty ad slots
+      slot.current.setCollapseEmptyDiv(true);
 
       // Enables all GPT services.
       window.googletag.enableServices();
--- a/packages/google-ad-manager/src/components/google-publisher-tag.tsx
+++ b/packages/google-ad-manager/src/components/google-publisher-tag.tsx
@@ -14,7 +14,7 @@ import GoogleAdManager, { GooglePublisherTagProps, Size } from "../../types";
  */
 const GooglePublisherTag: React.FC<
   Connect<GoogleAdManager, GooglePublisherTagProps>
-> = ({ unit, size, id, targeting = {}, data }) => {
+> = ({ unit, size, id, targeting = {}, data, collapseEmptyDiv = false }) => {
   /**
    * Append link to the `id` if `data` is specified.
    *
@@ -45,8 +45,8 @@ const GooglePublisherTag: React.FC<
       Object.entries(targeting).forEach(([key, value]) =>
         slot.current.setTargeting(key, value)
       );
-      // Collapse empty ad slots
-      slot.current.setCollapseEmptyDiv(true);
+      // Collapse empty ad slots if
+      slot.current.setCollapseEmptyDiv(collapseEmptyDiv);
 
       // Enables all GPT services.
       window.googletag.enableServices();
--- a/packages/google-ad-manager/types.ts
+++ b/packages/google-ad-manager/types.ts
@@ -78,7 +78,7 @@ export interface GooglePublisherTagProps {
    *
    * @defaultValue false
    */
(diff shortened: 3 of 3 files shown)


michalczaplinski:
Hi @VinuRaj 👋 !

Thanks for sending the PR. However, I'm afraid that this is something that outside the scope of the framework. As the [link that you shared](https://developers.google.com/publisher-tag/samples/collapse-empty-ad-slots) indicates: 

> The optimal configuration of this feature will depend on how often you expect ad slots to be filled:
> 
> 1. If slots will be filled most of the time, use PubAdsService.collapseEmptyDivs(). In this configuration, ad slots are expanded by default and collapse only if they cannot be filled.
> 2. If slots will not be filled most of the time, use PubAdsService.collapseEmptyDivs(true). In this configuration, ad slots are collapsed by default and only expand if they can be filled.
> Because googletag API calls have no effect until GPT begins to run, to minimize reflow you will also need to configure the CSS on your page so that slots which are collapsed by default have zero dimensions.

So, this should be up to the *individual theme user* to decide if they want to "Collapse empty ad slots" or not.

For example: If you're using a theme which has many slots defined but you expect few of them to be filled, if would make sense for you to call `PubAdsService.collapseEmptyDivs(true)` rather than the other option.

Hope this makes sense and I'm sorry to have to close this. I hope it does not detract you from contributing in the future though! 🙂

michalczaplinski:
Looking at the documentation, I realized that your PR is about collapsing the **individual** slots which actually makes sense in this context!

In Frontity by default a slot will be "expanded" if you define a fill for it, so this corresponds to option 1. from the documentation. 

In this context, it makes sense to have the slots collapse if they are not filled because otherwise they will always be filled with empty space.

michalczaplinski:
~~@DAreRodz I'm inclined to just merge it as is (no tests) unless you know of any potential issues.~~

Sorry for flip-flapping on this issue so much @VinuRaj but I think that it does not actually make sense to include this. 

Looking at the `GooglePublisherTag` component, it will always [render a `<div>`](https://github.com/VinuRaj/frontity/blob/2512b86d8a6395fb684ac1b7bf0443aa2f84ed08/packages/google-ad-manager/src/components/google-publisher-tag.tsx#L86-L90)  for a fill with certain `width` and `height` so collapsing the ads does not actually do anything useful here.

VinuRaj:
> ~@DAreRodz I'm inclined to just merge it as is (no tests) unless you know of any potential issues.~
> 
> Sorry for flip-flapping on this issue so much @VinuRaj but I think that it does not actually make sense to include this.
> 
> Looking at the `GooglePublisherTag` component, it will always [render a `<div>`](https://github.com/VinuRaj/frontity/blob/2512b86d8a6395fb684ac1b7bf0443aa2f84ed08/packages/google-ad-manager/src/components/google-publisher-tag.tsx#L86-L90) for a fill with certain `width` and `height` so collapsing the ads does not actually do anything useful here.

But, this would resolve/remove the empty space taken up by the pre-defined ad height? Otherwise it would take up the entire height of the ad, with just blank content.

The `setCollapseEmptyDiv(true)` tag would add a `style="display: none;"` if there is no ads to serve to the div, removing the empty space.

Screenshot 2021-04-13 at 8 55 48 AM

DAreRodz (changes requested):
@VinuRaj, it looks like you already tested it in Frontity and it works, isn't it? :smile:

> Looking at the `GooglePublisherTag` component, it will always [render a `<div>`](https://github.com/VinuRaj/frontity/blob/2512b86d8a6395fb684ac1b7bf0443aa2f84ed08/packages/google-ad-manager/src/components/google-publisher-tag.tsx#L86-L90) for a fill with certain `width` and `height` so collapsing the ads does not actually do anything useful here.

Hey @michalczaplinski, as long as "collapsing" means the `div` is hidden by using `display: none;` it doesn't matter if the `div` also has specified dimensions, we don't have to worry about that :slightly_smiling_face: . My only concern here is that the `div`'s attributes would be modified outside React – I guess that wouldn't be a problem as long as the component is not re-rendered, which in this case is not likely to happen.

In any case, instead of hardcoding the collapse behavior, I would make that option configurable by using a new property, something like `collapseEmptyDiv`, that could be specified in the fill settings as well.

This is what I would do:

1. Add a new boolean prop in the [`GooglePublisherTagProps` interface](https://github.com/frontity/frontity/blob/3709480717760e1d9c716fdfc5dc01a016554b74/packages/google-ad-manager/types.ts#L23-L71) called `collapseEmptyDiv` (you can get the doc text from [here](https://developers.google.com/publisher-tag/reference#googletag.slot-setcollapseemptydivcollapse,-opt_collapsebeforeadfetch)).
2. Use the `collapseEmptyDiv` prop instead of `true` when calling `setCollapseEmtpyDiv` inside the `useEffect` callback.
   ```ts
   // Collapse empty ad slots.
   slot.current.setCollapseEmptyDiv(collapseEmtpyDiv);
   ```

I would match the GPT's behavior here, which is to not collapse the slot unless specified.

michalczaplinski:
> Hey @michalczaplinski, as long as "collapsing" means the div is hidden by using display: none; it doesn't matter if the div also has specified dimensions, we don't have to worry about that

Yeah, I didn't know that GPT adds the `display: none` to the element in which case it does make sense of course :) Thanks for pointing that out @VinuRaj

> Add a new boolean prop in the GooglePublisherTagProps interface called collapseEmptyDiv (you can get the doc text from here).
> Use the collapseEmptyDiv prop instead of true when calling setCollapseEmtpyDiv inside the useEffect callback.

Agreed about the solution 👍

DAreRodz (approved):
Everything OK. Thanks, @VinuRaj and @michalczaplinski! 🙌

PR #779, opened 2021-04-12
Skip the BrowserStack related workflows for externals contributions

cristianbote:
**What**:

In order to avoid a lot of failed workflows we should skip the ones that are coming from externals contributors. Precisely talking about the ones that need to have access to secrets.

**How**:

Add an `if` case for each BrowserStack workflow to run only for the _owners_.

**Tasks**:

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

These are not going to be skipped entirely, at merge time these will run, as the merge is done by the owners.

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
+    if: github.repository_owner == 'frontity'
 
     steps:
       - name: Checkout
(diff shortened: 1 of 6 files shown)


cristianbote:
There might other ones that would need to be skipped maybe. Thoughts?

luisherranz (approved):
Looks great! I've used the same config for the other workflows that shouldn't run on forks either 🙂

cristianbote:
> Looks great! I've used the same config for the other workflows that shouldn't run on forks either 🙂

Awesome! Didn't want to touch those to be _unrelated_ to this change 😛 . Anyways, do you think the `release` action should be under the same if guard? 🤔

luisherranz:
> Anyways, do you think the release action should be under the same if guard?

I thought nobody would want to publish Frontity packages from a fork... 🤷

PR #781, opened 2021-04-13
Remove workspaces from package.json

### LINKED ISSUE #780: `npm install` fails after cloning a repo
**Expected behavior**

`npm install` should succeed

**Observed behavior**

npm fails with the following error:

Screen Shot 2021-04-12 at 11 05 35

```sh
git clone git@github.com:frontity/frontity.git
cd frontity
npm install
```

**Info about your system**

node version 15
npm version 7.6.0

**Possible solution**

not yet clear

luisherranz:
**What**:

Remove the `workspaces` property of our `package.json`.

**Why**:

Fixes https://github.com/frontity/frontity/issues/780.

Because we are still using `lerna` for the monorepo, but `npm` v7 thinks that we are using its workspaces and tries to install them but it fails.

@michalczaplinski reported a bug related to that: https://github.com/npm/cli/issues/3065

We had that property defined because Changesets don't read the workspace from `lerna`, but from the `package.json`. But Changesets only uses that workspace information to populate the package names when running `npx changeset` and without it outputs everything, so it is not required, just nice to have.

**How**:

Simply removing the `workspaces` property.

**Tasks**:

- [x] Change `package.json`

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,6 @@
     "prepare": "lerna bootstrap --hoist",
     "reinstall": "lerna clean --yes && npm run prepare"
   },
-  "workspaces": [
-    "packages/*"
-  ],
   "devDependencies": {
     "@babel/core": "^7.12.10",
     "@babel/preset-env": "^7.12.11",


michalczaplinski (approved):
👍

PR #785, opened 2021-04-14
Fix infinite scroll link race condition

DAreRodz:
**What**:

Prevent a possible race condition in `useInfiniteScroll`. It could happen when a link changes `state.router.link` right before the `useEffect` callback that manages the route ([this one](https://github.com/frontity/frontity/blob/0dc2e0476f9d5efc81ce27ce91a58b9595b15806/packages/hooks/use-infinite-scroll/index.ts#L108-L120)) runs.

**Why**:

It's a very sneaky bug that appears just a few times, but when it does, it makes the e2e tests to fail. It could be affecting UX as well. 

**How**:

Getting the value of `state.router.link` outside the `useEffect` callback and passing it as a callback dependency.

**Tasks**:

- [x] Code

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

I wasn't able to reproduce the error programmatically - that's why I didn't add any test.

### DIFF
--- a/.changeset/sixty-scissors-shave.md
+++ b/.changeset/sixty-scissors-shave.md
@@ -1,0 +1,5 @@
+---
+"@frontity/hooks": patch
+---
+
+Prevent a possible race condition in `useInfiniteScroll`. It could happen when a link changes `state.router.link` right before the `useEffect` callback that manages the route runs.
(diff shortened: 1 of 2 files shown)


michalczaplinski (approved):
LGTM 👍

PR #786, opened 2021-04-14
Fixes behaviour for the `--typescript` flag when using `frontity create`

orballo:
**What**:

Adding needed behaviour on `frontity create --typescript` for a seamless init of a project.

**Why**:

Because it is annoying to manually do these steps every time.

**How**:

Added additional steps to the create command in `frontity`.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] Update starter themes
- [ ] Update other packages
- [ ] End to end tests
- [ ] TypeScript tests

**Additional Comments**

### DIFF
--- a/.changeset/sweet-jars-matter.md
+++ b/.changeset/sweet-jars-matter.md
@@ -1,0 +1,5 @@
+---
+"frontity": patch
+---
+
+Modifies the `frontity create` command so it generates a `tsconfig.json` file and adds dev dependencies to `package.json`.
--- a/.changeset/sweet-jars-matter.md
+++ b/.changeset/sweet-jars-matter.md
@@ -2,4 +2,4 @@
 "frontity": patch
 ---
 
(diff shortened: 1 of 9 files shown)


orballo:
I'd like to have some comments in the actual configuration of the `tsconfig.json`. I just copied the one from the `frontity` repo and added an option that I needed recently to avoid warnings regarding the `css` prop.

Also, I think the `@types/react` and `@types/node-fetch` dependencies are enough, not even sure if the last one is still needed though, so if you think some dependency is missing let me know.

luisherranz (approved):
Looks great to me Edu 🙂

I've fixed the eslint errors so you don't have to. 

> if you think some dependency is missing let me know.

We can add more in the future if needed, don't worry.

The only thing missing is a changeset for the `frontity` package. Could you please add it? After that, you can merge it.

luisherranz:
Awesome. Thanks a lot for your contribution Eduardo!! 😄👏

PR #788, opened 2021-04-15
Create a git repo when running `frontity create` and add a `--no-git` flag to opt out of this behaviour

orballo:
**What**:

Automatically initialize a fresh repository when when running `frontity create`. Also add a `--no-git` flag which opts out of creating the repo.

**Why**:

Because it is annoying to create the git repo manually after creating a project.

**How**:

Creates a `.gitignore` file and runs `git init`, `git add .` and `git commit -m "Initialized with Frontity"`.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests
- [x] Add README for new cli-e2e tests

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


SantosGuillamot:
Thanks for opening the Pull Request :slightly_smiling_face: 

Regarding the content of the default `gitignore` file, what would you prefer: having a complete `gitignore` template and remove the files that don't apply to your project or start with just the most common files in a Frontity project and add the rest yourself? I don't have a strong opinion here.

orballo:
@SantosGuillamot Initially I added the template that Github uses and removed what seemed not necessary for a Frontity project, and added `build` as the only Frontity specific entry needed. I usually don't need more than `build`, `node_modules`, `.DS_Store` and `report.*.json`, but other users might have different things so maybe is better to have some entries from the template.

SantosGuillamot:
> @SantosGuillamot Initially I added the template that Github uses and removed what seemed not necessary for a Frontity project, and added build as the only Frontity specific entry needed. I usually don't need more than build, node_modules, .DS_Store and report.*.json, but other users might have different things so maybe is better to have some entries from the template.

Okay, that's fine to me :slightly_smiling_face: The most common errors we've seen in public Frontity repos is that they sometimes include `node_modules` and `build` folders, and that would be covered.

michalczaplinski:
Great PR @orballo, I suggested the same change like a year ago but never did the actual work 😅 

If I may chime in with a suggestion though - I think that Frontity should always initialize a git repo unless a flag like `--noGit` is passed.

I can't think of a reason why creating a git repo would be cause any problems, the time it takes to init a repo and make the first commit is negligible. Both `create-react-app` and `create-next-app` create a git repo by default as well.

orballo:
@michalczaplinski I agree with you that it should be the default behaviour. What do you think @luisherranz @SantosGuillamot ?

packages/frontity/src/steps/index.ts
```diff
@@ -310,6 +310,25 @@ export const downloadFavicon = async (path: string) => {
...
+    resolvePath(__dirname, "../../templates/gitignore-template"),
+    {
+      encoding: "utf8",
+    }
+  );
+  const filePath = resolvePath(path, ".gitignore");
+  await writeFile(filePath, fileTemplate);
+  await promisify(exec)("git init", { cwd: path });
+  await promisify(exec)("git add .", { cwd: path });
+  await promisify(exec)('git commit -m "initial commit"', { cwd: path });
```
cristianbote: Should we include some `Frontity` related wording in here? Like:
```
Initialised with Frontity
```
😃

SantosGuillamot:
I am not usually in favor of adding by default extra files that some users may not want, but in this case it's widely used and it's really the standard, so I think it makes sense. As Michal mentioned, it is something that `create-react-app`, Next or Gatsby are doing for example.

luisherranz:
I'm fine with whatever default you decide to use, I don't have a strong opinion about it 🙂

luisherranz:
@cristianbote, as you are the S2 this week, can you please do the full review? Thanks! 😄

cristianbote:
Based on the team's feedback and yours as well @orballo 😃  the flag should be to omit the git initialiser all together. So basically the user can opt out. Otherwise the code looks fine 👍

packages/frontity/src/steps/index.ts
orballo: I thought about this while writing it. I think it makes sense, so I'll change the message for the one you proposed.

orballo:
Guys, I have an important exam next week so I won't be able to work on this for a while. Just letting you know :)

luisherranz:
Thanks for the visibility @orballo and good luck with your exam 🙂

michalczaplinski:
Credit also goes to @DAreRodz for pair programming this together! 💪 

Video for reviewer(s) 🙂 

https://www.loom.com/share/fbc3ac0833d249c8b6acaef39a3f1adc

There are some things that occured to me which I mention at the end of the video so I leave them here as suggestions. I don't have a strong opinion on whether we should already implement all of them because they are arguably edge cases:

1.  Check if `git` exists (`git init` should succeed). Maybe we should add a more informative error to the user in case git is not installed.
2. `git commit` can fail if the author [email and name are not](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup) set in the config. In this case we should clean up because a repository would have already been initialized with `git init`.
3. Check if `.gitignore` exists. If it does, we should append to the existing `.gitignore` file instead overwriting it with our own `.gitignore`.
4. Perhaps we should bail out completely if we are already inside of a git repo. E.g if someone has a monorepo setup, we would create a git repo inside of their main git repo and this is obviously wrong.

luisherranz (changes requested):
Thanks for the video Michal, they are really useful 🙂

Regarding your question at the end of the video and https://github.com/frontity/frontity/pull/788#issuecomment-828699445 comment: **Absolutely**.

We cannot release this unless it fails gracefully if `git` fails. If we don't do that, we will be breaking `npx frontity create` for lots of people, which is not acceptable.

So, let's make this right! These are my requested changes:

- [x] Skip git initialization if CWD is inside a git project.

  I've used `simple-git` in the past and it worked great. It has a utility for this: https://github.com/steveukx/git-js#git-rev-parse--repo-properties

- [x] Don't overwrite `.gitignore` if it exists. If it does, just add the strictly necessary modifications: `node_modules` and `build`.

- [x] Fail gracefully and inform if something goes wrong, but continue with the create.

  Instead of doing it inside the step I would add a try/catch in the command to be able to send a message to the console:

  ```js
  if (!noGit) {
    step = initializeGit(path);
    emit("message", `Initializing git.`, step);
    try {
      await step;
    } catch (e) {
      emit("message", `Git initialization failed. Skipping.`);
    }
  }
  ```

  So instead of checking for individiual problems like if `git` is not installed, or the commit failed and so on, I would just catch any kind of error, inform that the step failed, and continue.

  If you want more control inside the step, maybe we could use `simple-git` to run those steps as well. Not sure if that would be an improvement over `exec` though:

  ```js
  const git = simpleGit();
  // ...
  try {
    await git.init();
    await git.add("./*");
    await git.commit("...");
  } catch (e) {
    // Something here? Maybe delete .git folder?

    // Send error to the command.
    throw new Error(e);
  }
  ```

  https://github.com/steveukx/git-js#catching-errors-in-async-code

michalczaplinski:
This is still WIP, although there is not much left. I've taken the opportunity to create actual e2e CLI tests. I've made a video for more details 🙂 

https://www.loom.com/share/644acccc5de64d59a9c7d581ed413f71

luisherranz:
This looks really cool Michal 🙂👏

I only have one question: what do we gain from running the tests inside a docker container vs running them in the actual machine?

I ask this because GH actions can be run in Ubuntu, Windows, and macOS machines, so theoretically we could test that the CLI works on those 3 operative systems, as opposed to only testing that it works on the `node:alpine` image. Other than that I think it looks great!

michalczaplinski:
Thanks Luis! 🙂 I didn't work on integrating the CLI e2e tests with github actions until now, but after taking a look I see that it's not not so straightforward to "just" run them in Ubuntu, Windows, and macOS in github actions.

1. The main reason is that we want to run each test in a separate container and this is going to get messy very quickly. This is because with github actions we have to specify each container "ahead of time" via a matrix. So any parameters that would go into an individual test case would have to be specified directly in the workflow via the `matrix` property. But this is very constraining. Imagine if we want to run additional steps just for one test like installing `git` and then creating a `.gitignore` file with `touch .gitignore`. It's going to be hard to express those in the workflow file and then pass to the test file. 

   This is essentially the same problem that we had in the e2e tests with cypress, where in the end we decided to run all tests in a shared container and only create a new container for a [test run with specific parameters](https://github.com/frontity/frontity/blob/d590755764ba6731c0ecd00afe23df9c7970b8d1/.github/workflows/e2e-gh-actions.yml#L12-L35) like `--wpVersion 5.5`  or `--publicPath /custom/path`.

   Basically I mean this: https://www.loom.com/share/d48a1f5bb1924da0814fba53f950d4d9

2. This is not a major problem perhaps, but in this specific case I wanted to test that the CLI is behaving correctly if `git` is not installed and the Github Actions containers already come with `git` preinstalled.


So all the being said, my current thinking is that the best option would be create a "simple" github action with a simple test or two that just make sure that `frontity create` works fine across all OSes. Then for all the tests that test the actual logic of `frontity create` I would stick with the `alpine` image.

luisherranz:
Perfect, thanks for the explanation Michal 🙂👍

.github/workflows/unit-tests.yml
```diff
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
```
michalczaplinski, line 11: Since node 16 was released a few weeks ago, node 10 is EOL so I've updated our actions to include v16 and remove v10

.github/workflows/unit-tests.yml
```diff
@@ -36,6 +36,3 @@ jobs:
         run: npm run test:ci
         env:
           CI: true
-
-      - name: Upload result to codecov
-        run: npm run codecov
```
michalczaplinski, line 41: I think nobody's been looking at codecov stats for months now so if there's no veto, I would just remove it. Also removing the `codecov.yml` file at the repo root.

michalczaplinski:
ok, so I think this is ready for review: https://www.loom.com/share/66c068b9f6cd465e99e2dd8a356d37b9

There are some failures in the unit tests that appeared just this evening but it seems unrelated to the current change (tests were passing earlier today and the failures are snapshots of webpack configuration in `@frontity/core`). I ll investigate those tomorrow.

.github/workflows/unit-tests.yml
luisherranz: No veto here! 🙂👍

.github/workflows/unit-tests.yml
luisherranz: I'm not so sure about removing Node v10 just yet, because Frontity still supports it.

luisherranz:
> There are some failures in the unit tests that appeared just this evening but it seems unrelated to the current change (tests were passing earlier today and the failures are snapshots of webpack configuration in @frontity/core). I ll investigate those tomorrow.

They were failing due to an update in the `loadable-components` package so I have updated the `package-lock.json` and they pass now 👍

luisherranz:
We've done a pair review and renamed some things to make things more clear. I'll review and push them tomorrow, along with a `README.md` file that will explain how to run the new tests.

luisherranz (changes requested):
Hey Michal, I've pushed the changes we did together yesterday and a video explain a few other things I polished as well: https://www.loom.com/share/6adbaee9a308451281afe083776302bc

I think the only thing missing is writing a `REAMDE.md` file to explain how the new tests work, similar to what we have for the Cypress E2E tests.

Do you mind doing that? Thanks!! 😄

luisherranz:
I went a bit back and forth with the configuration of some of the GH actions, but I think everything is solved now.

All yours Michal! 🙂

michalczaplinski:
Added the new README.md for the e2e tests of the CLI. Back to you @luisherranz if you'd like to review it quickly again :)

luisherranz (approved):
It's perfect, thanks Michal! 👏👏

michalczaplinski:
Will merge it now. You can see that the unit tests have actually passed when you go to https://github.com/frontity/frontity/pull/788/checks but there seems to be a glitch in the UI where they seem stuck:
Screen Shot 2021-05-17 at 12 40 25

luisherranz:
> Will merge it now. You can see that the unit tests have actually passed when you go to https://github.com/frontity/frontity/pull/788/checks but there seems to be a glitch in the UI where they seem stuck

That's because those were removed in this PR. I'll change the global settings now that this is merged 🙂

PR #791, opened 2021-04-16
Pass `libraries` to derived properties

### LINKED ISSUE #716: Derived state does not have access to `actions` or `libraries`, only to `state`
**Expected behavior**

Derived state should have access to `state`, `actions` and `libraries` inside of the callback.

It **seems** that when using derived state, you can access `state` `actions` and `libraries`:

https://github.com/frontity/frontity/blob/a569db247e11195a0ef341ebf0b62ab87df7318d/packages/types/src/derived.ts#L10-L15

The TS types also seem to indicate that this is possible: 

https://user-images.githubusercontent.com/5417266/107553538-deb1cb80-6ba2-11eb-84d6-6b93568b3fc4.mov

**Observed behavior**

Only `state` is passed to the derived state callback in `@fronity/connect`:

https://github.com/frontity/frontity/blob/107d3543ce5463186809b7e6f50ca31ffbdc107d/packages/connect/src/handlers.js#L23-L24

Create any derived state and try to access `libraries` or `actions` in the callback. E.g: 

```js
state = {
  myPackage: {
    test: ({ state, libraries, actions }) => {
      // libraries is undefined !!!
      libraries.source.parse('hello');
      
	  // actions is undefined too.
      actions.amp.init();
    }
  }
}
```

**Possible solution**

We should be clear that `actions` and `libraries` are available in derived state.

OR

We should make `actions` and `libraries` available in derived state.

DAreRodz:
**What**:

Allow derived props to use libraries in their implementation.

**Why**:

To fix #716 

**How**:

Passing `libraries` when the derived props are instantiated inside `createStore`.

https://github.com/frontity/frontity/blob/61f71264db6e865c08065da067632c56384887c5/packages/connect/src/create-store.js#L107-L110

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] TypeScript tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional comments**

`actions` are not passed down because it would break the Flux pattern (see https://github.com/frontity/frontity/issues/716#issuecomment-777340627).

### DIFF
--- a/.changeset/heavy-news-train.md
+++ b/.changeset/heavy-news-train.md
@@ -1,0 +1,6 @@
+---
+"@frontity/connect": patch
+"@frontity/types": patch
+---
+
+Modify derived props so they receive `state` and `libraries` as arguments. Also, fix the `Derived` type definition.
--- a/packages/connect/src/__tests__/create-store.tests.js
+++ b/packages/connect/src/__tests__/create-store.tests.js
@@ -14,6 +14,10 @@ beforeEach(() => {
         prop3: ({ state }) => state.prop1 + state.nested1.prop2,
         prop4: ({ state }) => (num) => state.nested1.prop3 + num,
         prop5: 0,
+        prop6: ({ state, libraries }) => {
+          const { prop3, prop4 } = state.nested1;
+          return libraries.nested1.sum(prop3, prop4(0));
+        },
       },
     },
     actions: {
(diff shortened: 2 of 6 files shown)


michalczaplinski (approved):
Great 👍  I've also made a PR for the docs to add an example with passing `libraries` : https://github.com/frontity/docs/pull/284

PR #792, opened 2021-04-16
Fix `Link` component for multisites 

### LINKED ISSUE #691: [Components] The `Link` component doesn't remove the WordPress subdirectory
**_Bug report_**

- 
I think I've found a bug while triaging this issue: #665 

**Expected behavior**

Using a WordPress multisite setup, the `Link` component should transform links that points to a WordPress site, removing the subdirectory from the link.

**Observed behavior**

The final link maintains the WordPress subdirectory. When a link is clicked, 404 is returned.

- Create a WordPress multisite (subdirectories) with Local: https://www.youtube.com/watch?v=9AZapIsjCTo
- Add a new site.
- Add a link to any page (e.g. in "Sample Page", to "Hello World").
- Create a Frontity project and change `state.source.url` to point to the site added before.
- Visit the page with the link and click it.

**Possible solution**

I'm pretty sure the problem is in this line of code:

https://github.com/frontity/frontity/blob/c139094e956d3119f9fd1ea2342717e29e187f2f/packages/components/link/utils.ts#L139

There, I think we should remove the WordPress subdirectory from `linkUrl.pathname`. We could get the WordPress subdirectory from `state.source.url` easily.

DAreRodz:
**What**:

Fixes #691

**Why**:

Right now, the `Link` component doesn't remove the subirectory of a WordPress multisite (e.g. `/wp-site` in `http://backend.url/wp-site/some-post`), making links in Frontity navigate to the wrong URL.

**How**:

Fixing the `removeSourceUrl` function (WIP):

https://github.com/frontity/frontity/blob/f0b6deedef03b2a713c52cbd88e379eb26d8f333/packages/components/link/utils.ts#L117-L141

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] Unit tests

**Unrelated Tasks**

- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


cristianbote (approved):
👍  awesome @DAreRodz!

PR #795, opened 2021-04-21
Fix category and tag ID collision inside `populate`

### LINKED ISSUE #497: Missing terms from source.tag that are assigned to the current post
**_Bug report_**

**Expected behavior**

Whenever a single post is displayed, all tags/categories in it should be also loaded in `source.tag`

**Observed behavior**

This doesn't seem to be the case. Sometimes (pretty randomly at a first glance at least). There seem to be some sort of, race condition? that makes `source.tag` or `source.category` missing some of the tags that are assigned to the current post in `post.tags` or `post.category`

I think it's difficult to test it but I've left a piece of code here that should help: https://github.com/igmoweb/igmoweb.com/blob/tags-debug/packages/igmoweb-theme/src/components/single/post-tags.js#L22-L33 so you'll need to download my project.

Once, you are using my project and **switched to `tags-debug` branch**, follow these steps:

1. Go to the main page `localhost:3000` and open the Chrome Dev Tools.
2. Most of the times, you need to go fast: Do a hard refresh in the homepage (Cmd + Shift + R), then click on a post that will trigger the error, in my site, this is a good one as it contains several tags:
Captura de pantalla 2020-06-15 a las 13 19 29
3. If any of the tags are `undefined`, the source tab should stop at the `debugger;` line. I've solved this error in the `master` branch so make sure that you are in `tags-debug`.
4. If it doesn't stop, go back to the homepage, hard refresh and click on the same post again, try to do it as fast as you can. After 3-4 attempts you should see the error.

Worths mentioning that it always happens with the same tag and that I haven't installed the dev Frontity version to do a proper debug (I really haven't had time though I'd like to and this seems like a time-consuming issue).

**Info about your system**

```
## System:
 - OS: macOS Mojave 10.14.6
 - CPU: (4) x64 Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
 - Memory: 319.46 MB / 16.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
 - npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
## Browsers:
 - Chrome: 83.0.4103.97
 - Edge: Not Found
 - Firefox: 77.0.1
 - Safari: 12.1.2
## npmPackages:
 - @fortawesome/fontawesome-svg-core: ^1.2.28 => 1.2.28 
 - @fortawesome/free-brands-svg-icons: ^5.13.0 => 5.13.0 
 - @fortawesome/free-regular-svg-icons: ^5.13.0 => 5.13.0 
 - @fortawesome/free-solid-svg-icons: ^5.13.0 => 5.13.0 
 - @fortawesome/react-fontawesome: ^0.1.10 => 0.1.10 
 - @frontity/core: ^1.7.0 => 1.7.0 
 - @frontity/html2react: ^1.3.3 => 1.3.3 
 - @frontity/tiny-router: ^1.2.0 => 1.2.0 
 - @frontity/twentytwenty-theme: ^1.2.0 => 1.2.0 
 - @frontity/wp-source: ^1.7.1 => 1.7.1 
 - @wordpress/components: ^9.8.0 => 9.7.0 
 - @wordpress/date: ^3.9.0 => 3.9.0 
 - @wordpress/prettier-config: ^0.2.0 => 0.2.0 
 - babel-eslint: ^10.1.0 => 10.1.0 
 - eslint: ^7.2.0 => 6.8.0 
 - eslint-config-prettier: ^6.11.0 => 6.11.0 
 - eslint-config-wordpress: ^2.0.0 => 2.0.0 
 - eslint-plugin-prettier: ^3.1.4 => 3.1.4 
 - eslint-plugin-react: ^7.20.0 => 7.20.0 
 - eslint-plugin-react-hooks: ^4.0.4 => 4.0.4 
 - eslint-plugin-sort-destructure-keys: ^1.3.4 => 1.3.4 
 - frontity: ^1.8.0 => 1.8.0 
 - frontity-social-previews: ^1.0.1 => 1.0.1 
 - highlight.js: ^10.1.0 => 10.1.0 
 - highlightjs-line-numbers.js: ^2.8.0 => 2.8.0 
 - igmoweb-theme: ./packages/igmoweb-theme => 1.0.0 
 - lodash: ^4.17.15 => 4.17.15 
 - polished: ^3.6.5 => 3.6.5 
 - pretty-quick: ^2.0.1 => 2.0.1 
 - prismjs: ^1.20.0 => 1.20.0 
 - react-ga: ^3.0.0 => 2.7.0 
 - react-share: ^4.2.0 => 4.2.0 
 - wp-prettier:  1.19.1 
## npmGlobalPackages:
 - frontity: Not Found
 - npx: Not Found
```

**Possible solution**

The problem is that `populate` doesn't recognize different terms with the same ID: https://github.com/frontity/frontity/issues/497#issuecomment-644689264

We didn't expet that situation to happen in `wp.org` but it's something that `wp.com` does for its sites.

We must add different schemas for categories and tags or change all the schemas implementation.

This is the current schema for taxonomies:

- [packages/wp-source/src/libraries/schemas/taxonomies.ts](https://github.com/frontity/frontity/blob/107d3543ce5463186809b7e6f50ca31ffbdc107d/packages/wp-source/src/libraries/schemas/taxonomies.ts)

DAreRodz:
**What**:

Make categories and tags to be considered objects of a different kind, inside `libraries.source.populate()`. That avoids a possible ID collision, ensuring both colliding entities are populated correctly.

**Why**:

To fix #497.

**How**:

Adding two new [`normalizr`](https://github.com/paularmstrong/normalizr) schemas (we use that library internally) for each kind of entity.

**Tasks**:

- [x] Code
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


packages/wp-source/src/libraries/__tests__/__snapshots__/populate.tests.ts.snap
```diff
@@ -61,7 +61,10 @@ Object {
         ],
         "wp:term": Array [
           Array [
-            1,
+            Object {
+              "id": 1,
+              "schema": "taxonomyEntity",
+            },
```
DAreRodz, line 67: `normalizr` replaces IDs with an object containing `id` and `schema`, now that term entities can have different schemas, so I had to update a lot of snapshots. Almost all of the changes inside snapshots are like this one, in the property `_embedded["wp:term"]` of post type entities.

I don't consider that this modifies how entities are consumed in any way because nobody should be reading `_embedded` to get term IDs ― for that, posts already have the `categories` and `tags` properties.

packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/post-type-with-query.tests.ts.snap
```diff
@@ -21,7 +21,16 @@ Object {
...
+  "category": Object {
+    "1": Object {
+      "count": 5,
+      "id": 1,
+      "link": "/category/cat-1/",
+      "parent": 0,
+      "slug": "cat-1",
+      "taxonomy": "category",
+    },
+  },
```
DAreRodz, line 33: In this case, the snapshot was wrong :sweat_smile:. The bugfix now makes the category with id `1` to be populated.

packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/post-type-with-query.tests.ts.snap
```diff
@@ -80,6 +95,16 @@ Object {
...
+    "/category/cat-1/": Object {
+      "id": 1,
+      "isFetching": false,
+      "isReady": false,
+      "link": "/category/cat-1/",
+      "page": 1,
+      "query": Object {},
+      "route": "/category/cat-1/",
+      "taxonomy": "category",
+    },
```
DAreRodz, line 107: Same here. :eyes:

DAreRodz:
Hey @cristianbote, if you need me to explain the changes better let me know. :wink:

cristianbote (approved):
Looks good to me! 👍

cristianbote:
> Hey @cristianbote, if you need me to explain the changes better let me know. 😉

😆  I mean the functional changes are pretty self explanatory. The technical ones the same.

PR #804, opened 2021-04-27
Prevent inifinite scroll from loading the same post twice if the post's link contains a query string.

### LINKED ISSUE #748: Infinite Scroll loads the same post twice when there are queries in the url
**_Bug report_**

**Expected behavior**

When I visit a url with a query, like `?utm_campaign`, in a site where the Infinite Scroll is implemented, it should load the next post the same way it does when a query doesn't exist.

**Observed behavior**

When there is a query in a post and you scroll, it loads the same post but without the query.

* Go to https://frontity.org/blog/monthly-roundup-february-2021/?random_query=1234
* Start scrolling until it fetch the next post.

You will see it is the same one.

**Possible solution**

I guess that, at some point when loading the next post, we will need to check the `route` instead of the `link.

Here is where we are checking if the link is included in the archive or not:
https://github.com/frontity/frontity/blob/dev/packages/hooks/use-post-type-infinite-scroll/utils.ts#L52-L54

That firstLink we are passing as a parameter could be the `route` instead, without the query params.

michalczaplinski:
**What**:

Fix #748.

**How**:

Creating a new data structure that extends a native `Set` and implements an `addLink()` method which exhaustively searches the set to check if a link with an identical `pathname` already exists.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] End to end tests

**Unrelated Tasks**

- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


michalczaplinski:
video for reviewer:

https://www.loom.com/share/a78a2ec309fc4e9393241931d27ac55a

luisherranz (approved):
Hey Michal, thanks for the video 🙂

The implementation looks great to me. 

It is true that this PR will both: 
1. Solve this bug.
1. Limit the ability to use the infinite scroll with posts that are differentiated by queries in the URL.

But even though 2. is not ideal, it is not the fault of this PR, but of the lack of a framework-wide system to configure which queries are important and which queries can be ignored as discussed [here](https://community.frontity.org/t/wpsource-define-a-map-of-allowedqueries-to-only-preserve-them-in-data-references-links/3501).

Great work Michal. Approved! ✅

PR #806, opened 2021-04-28
Fix internal redirection for `homepage` in `wp-source`.

### LINKED ISSUE #805: Frontity returns a 404 in all pages when state.source.homepage is defined
**_Bug report_**

**Expected behavior**

All the pages should work fine even if `state.source.homepage` is defined.

**Observed behavior**

In the latest version of `wp-source` (1.11.3), all the pages except the homepage return a 404 if `state.source.homepage` is defined. It seems the problem is in [this pattern](https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/actions.ts#L275-L277) we introduced in [this Pull Request](https://github.com/frontity/frontity/pull/793). The RegExp is true for all the pages/posts urls and it causes the issue. If I change that piece of code, everything seems to work fine.

* Create a new Frontity project.
* Add the `homepage` setting in the `wp-source` package.
* Navigate to any page or post that is not the homepage.

**Possible solution**

Not sure if it's the best option but I've seen that if we update the RegExp to exclude the urls of the pages/posts and only match the homepage, it seems to work fine. This is the example I used:

```
const pattern = `RegExp:^${concatLink(subdirectory)}(?!(\\?([^&#]+&)*s=)|[a-zA-Z]+)`;
```

michalczaplinski:
**What**:

Fix #805

**Why**:

The regex used to get the internal redirection if a `homepage` is defined is not correct. As described in #805, the regex would match not only for the homepage, but also for other pages.

**How**:

Update the regex used to match the redirection if a homepage is present.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] Unit tests

**Unrelated Tasks**

- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/packages/wp-source/src/actions.ts
+++ b/packages/wp-source/src/actions.ts
@@ -274,7 +274,7 @@ const actions: WpSource["actions"]["source"] = {
     if (homepage) {
       const pattern = `RegExp:^${concatLink(
         subdirectory
-      )}(?!(\\?([^&#]+&)*s=))`;
+      )}(?!(\\?([^&#]+&)*s=))$`;
 
       redirections.push({
         name: "homepage",
--- a/packages/wp-source/src/actions.ts
+++ b/packages/wp-source/src/actions.ts
@@ -16,6 +16,7 @@ import {
   isEagerRedirection,
   fetchRedirection,
   shouldBailRedirecting,
+  verboseRegExp,
 } from "./utils";
 
 const actions: WpSource["actions"]["source"] = {
@@ -272,32 +273,22 @@ const actions: WpSource["actions"]["source"] = {
     } = state.source;
 
     if (homepage) {
-      const regex =
-        // prettier-ignore
-        // This regex is using negative lookahead. It means that the whole regex
-        // matches if the group using the negative lookahead DOES NOT match.
-        "(?!" +         // negative lookahead       
-          "(" +           // group    
-            "\\?" +         // match "?" literally
-            "(" +           // group     
-              "[^&#]+" +      // match any character that is not `&` or `#` 1 or more times 
-              "&" +           // match `&` literally
-            ")*" +          // match the group zero or more times
-            "s=[^&#]*" +    // match a search query param (e.g. `s=something`)
-          ")" +           // end group
-        ")" +           // end negative lookahead
-        "(" +           // group 
-          "\\?" +         // match "?" literally
-          "([^#]+)*" +    // match the query string part of the URL zero or more times
-        ")?" +          // match the group zero or one times
-        "(#.*)?" +      // match the hash part of the URL 
-        "$"; // match line end
+      const homePath = concatLink(subdirectory);
+      const regExp = verboseRegExp`
+        ^                // Beginning of line.
+        ${homePath}      // Path to the homepage.
+        (                // Must be followed by:
+          $|             // 1. End of line or...
+          #|             // 2. Hashtag or...
+          \?             // 3. Query...
+          (?!            //    not containing
+            ([^&#]+&)*   //    ...any parameter (0..n)
+            s=           //    ...and the search param (s=)
+          )
+        )
+      `;
 
-      const pattern = `RegExp:^${concatLink(subdirectory)}${regex}`;
-
-      // For those who are able to read that sort of thing the full pattern is:
-      // RegExp:^/(?!(\?([^&#]+&)*s=[^&#]*))(\?([^#]+)*)?(#.*)?$
-      // (It has to be double-escaped in the string form above though)
+      const pattern = `RegExp:${regExp}`;
 
       redirections.push({
         name: "homepage",
--- a/packages/wp-source/src/actions.ts
+++ b/packages/wp-source/src/actions.ts
@@ -272,9 +272,32 @@ const actions: WpSource["actions"]["source"] = {
     } = state.source;
 
     if (homepage) {
-      const pattern = `RegExp:^${concatLink(
-        subdirectory
-      )}(?!(\\?([^&#]+&)*s=))$`;
+      const regex =
+        // prettier-ignore
+        // This regex is using negative lookahead. It means that the whole regex
+        // matches if the group using the negative lookahead DOES NOT match.
+        "(?!" +         // negative lookahead       
+          "(" +           // group    
+            "\\?" +         // match "?" literally
+            "(" +           // group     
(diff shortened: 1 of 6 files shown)


michalczaplinski:
I'm sorry that the video is slightly rambling (I was a bit tired already) and I got interrupted twice 🤦 

https://www.loom.com/share/57a5e094855e413989222596cf5cd478

luisherranz (changes requested):
Hey Michal, the new regexp seems fine at a first glance, but we must confirm that with a proper unit test to make sure that it works in all the scenarios we know about so far.

This is the unit test that David created in that PR. I see he added some URLs and matched them against the RegExp, but none of them had a page homepage. We need to make sure that we are covering that case.

https://github.com/frontity/frontity/pull/793/files#diff-d78f73d7107a2c5d6d85eb0b870213faa478e4ff9f27f78ef363d8adafe1ec3aR253-R282

packages/wp-source/src/actions.ts
```diff
@@ -272,9 +272,32 @@ const actions: WpSource["actions"]["source"] = {
...
+          "([^#]+)*" +    // match the query string part of the URL zero or more times
+        ")?" +          // match the group zero or one times
+        "(#.*)?" +      // match the hash part of the URL 
+        "$"; // match line end
+
+      const pattern = `RegExp:^${concatLink(subdirectory)}${regex}`;
+
+      // For those who are able to read that sort of thing the full pattern is:
+      // RegExp:^/(?!(\?([^&#]+&)*s=[^&#]*))(\?([^#]+)*)?(#.*)?$
+      // (It has to be double-escaped in the string form above though)
```
DAreRodz: Okay, this just blowed my mind, hahaha :exploding_head:

Writing comments inside regexps is actually a good idea, so I did a quick search, and I found [this article](https://keleshev.com/verbose-regular-expressions-in-javascript) explaining how this could be implemented for JavaScript using template literals. I've modified the code a bit –just to allow interpolation– and, voilà, I was able to write something like this: :smile:

```ts
    if (homepage) {
      const homePath = concatLink(subdirectory);
      const regExp = verboseRegExp`
        ^                // Beginning of line.
        ${homePath}      // Path to the homepage.
        (                // Must be followed by:
          $|             // 1. End of line or...
          #|             // 2. Hashtag or...
          \?             // 3. Query...
          (?!            //    not containing
            ([^&#]+&)*   //    ...any parameter (0..n)
            s=           //    ...and the search param (s=)
          )
        )
      `;

      const pattern = `RegExp:${regExp}`;
```

DAreRodz: BTW, this is the function I wrote:

```ts
/**
 * Write verbose RegExp strings. It supports:
 * - Expressions in multiple lines.
 * - Single-line and multiple-line comments.
 * - Interpolation.
 *
 * @remarks Purely based on
 * https://keleshev.com/verbose-regular-expressions-in-javascript
 *
 * @param input - Input.
 * @param args - Interpolated args.
 * @returns Regexp in string format.
 */
export const verboseRegExp = (
  input: TemplateStringsArray,
  ...args: string[]
) => {
  const source = input.raw.reduce(
    (source, input, index) =>
      source.concat(input, String.raw`${args[index] || ""}`),
    ""
  );

  const regexp = /(?<!\\)\s|[/][/].*|[/][*][\s\S]*[*][/]/g;
  const result = source.replace(regexp, "");

  return result;
};
```

DAreRodz:
@michalczaplinski, @luisherranz, I've added the missing unit tests, in case you want to take a look.

packages/wp-source/src/actions.ts
michalczaplinski: Awesome, thanks for digging into this @DAreRodz 🙂 I've seen a similar approach in https://stackoverflow.com/a/53925033/2638310 where I got the idea for concatenating the regex to make it more readable.

> I've modified the code a bit –just to allow interpolation– and, voilà, I was able to write something like this: 😄

Did you mean that you modified the code in this PR (because you didn't push any changes) or were you just testing locally?

michalczaplinski:
Thanks for adding the test @DAreRodz and for the review @luisherranz. 
It slipped my mind but I was going to mention in the video that I ll add a test but I wanted to let you review already. (Shouldn't grab videos late in the evening I guess 🙂 )

luisherranz (approved):
Thanks guys!

I have approved this as it is, but @darerodz if you want, go ahead and add your function for commented regexps 🙂👍

packages/wp-source/src/actions.ts
DAreRodz: > Did you mean that you modified the code in this PR (because you didn't push any changes) or were you just testing locally?

I meant [this code](https://gist.github.com/keleshev/c49465caed1f114b2bb3f2b730e221ca) (from the [article](https://keleshev.com/verbose-regular-expressions-in-javascript) I mentioned). :smile: 

An yes, I just tested it locally. I don't think it would be good idea to add a new function that we would have to test and maintain, just to be able to write RegExps that way. But I thought it was worth sharing it anyway. :grin:

DAreRodz: On the other hand, we could use that function to comment other regexps we have, like this one:

https://github.com/frontity/frontity/blob/ad55f6d854e59d2d2939e03f14de9eca690aa25a/packages/wp-source/src/libraries/route-utils.ts#L105

It could be a good idea in the end. :thinking:

DAreRodz:
Okay, I'll add the function for commented RegExps. :blush:

luisherranz:
Thanks David! 👍

@santosguillamot this solves a critical bug, do you want to do a release or do you want me to do it?

SantosGuillamot:
> @SantosGuillamot this solves a critical bug, do you want to do a release or do you want me to do it?

I will do it don't worry.

packages/wp-source/src/actions.ts
```diff
@@ -272,9 +273,22 @@ const actions: WpSource["actions"]["source"] = {
...
+        (                // Must be followed by:
+          $|             // 1. End of line or...
+          #|             // 2. Hashtag or...
+          \?             // 3. Query...
+          (?!            //    not containing
+            ([^&#]+&)*   //    ...any parameter (0..n)
+            s=           //    ...and the search param (s=)
+          )
+        )
+      `;
```
michalczaplinski, line 289: oh, that's a way better regex than my previous one, nice @DAreRodz ! 😄

PR #810, opened 2021-04-29
Remove negative lookbehind

### LINKED ISSUE #809: Remove negative lookbehind in regular expressions
**_Bug report_**

**Observed behavior**

Our Cypress tests fail on Firefox 70 because it doesn't support lookbehind in regular expressions: https://caniuse.com/?search=Lookbehind%20in%20JS%20regular%20expressions

Safari doesn't seem to support them either.

**Possible solution**

It seems like the problem was introduced here: https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/utils.ts#L286

We simply have to find a way to do the same, but without using lookbehind 😄

DAreRodz:
**What**:

Remove a negative lookbehind in the `verboseRegExp` function.

**Why**:

To make tests pass again on Firefox 70. Fixes #809

**Tasks**:

- [x] Code
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

### DIFF
--- a/.changeset/brave-kiwis-fry.md
+++ b/.changeset/brave-kiwis-fry.md
@@ -1,0 +1,5 @@
+---
+"@frontity/wp-source": patch
+---
+
+Remove a negative lookbehind in the `verboseRegExp` function.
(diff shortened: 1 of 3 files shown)


luisherranz (approved):
Awesome. Thanks David 🙂

PR #823, opened 2021-05-19
Script component should not define the children prop

### LINKED ISSUE #821: Issue with embedded tweets in Frontity's blog posts
**_Bug report_**

**Expected behavior**

Blog posts with embedded tweets should be properly displayed.

**Observed behavior**

When a I visit a Frontity blog post that have a embedded tweet the screen goes blank. Here's some examples:

- https://frontity.org/blog/frontity-2020-a-year-in-review/
- https://frontity.org/blog/frontity-2019-a-year-in-review/
- https://frontity.org/blog/recap-wordcamp-sevilla-2019-first-wordcamp-developers-spain/
- https://frontity.org/blog/introducing-twenty-nineteen-frontity-theme/

Cristian took a look at this issue and says the error is coming from the embedded twitter code, as it’s creating a `Script` element which usually does not have `children` and is the same in this case.

cristianbote:
**What**:

`Script` component should not try to define the `children` prop. The `children` prop is falsy, as in `undefined`, `null`, `''` etc.

**Why**:

The incoming prop is not the same type as the DOM `.children`, and because we are bridging the virtual DOM world with the real DOM they are mismatching and throwing.

**How**:

Adding an explicit case where we do skip setting the `children` entirely.

**Tasks**:

- [x] Code
- [x] TypeScript
- [x] Unit tests
- [x] TypeScript tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] Update starter themes
- [ ] Update other packages
- [ ] TSDocs

**Additional Comments**

### DIFF
--- a/packages/components/__tests__/__snapshots__/script.tests.tsx.snap
+++ b/packages/components/__tests__/__snapshots__/script.tests.tsx.snap
@@ -1,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Script should execute the code 1`] = `null`;
+
+exports[`Script should not render a script tag without code or src 1`] = `null`;
+
+exports[`Script should not render the passed children with src 1`] = `null`;
+
+exports[`Script should remove the script tag at unmount 1`] = `null`;
+
+exports[`Script should render a script tag with src 1`] = `null`;
--- a/packages/components/__tests__/__snapshots__/script.tests.tsx.snap
+++ b/packages/components/__tests__/__snapshots__/script.tests.tsx.snap
@@ -1,11 +1,3 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Script should execute the code 1`] = `null`;
-
 exports[`Script should not render a script tag without code or src 1`] = `null`;
-
-exports[`Script should not render the passed children with src 1`] = `null`;
-
-exports[`Script should remove the script tag at unmount 1`] = `null`;
-
-exports[`Script should render a script tag with src 1`] = `null`;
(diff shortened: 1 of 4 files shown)


packages/components/__tests__/script.tests.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+describe("Script", () => {
+  test("should not render a script tag without code or src", () => {
+    let root;
+    act(() => {
+      root = create(<Script id="script-1">{"let index = 0;"}</Script>);
+    });
+
+    // This is `null`. This is here to make sure the virtual dom tree does not return anything.
+    expect(root.toJSON()).toMatchSnapshot();
```
DAreRodz: I would use an inline snapshot here. That way, we don't need to say "this is `null`", haha. :smile:

```suggestion
    // This is here to make sure the virtual dom tree does not return anything.
    expect(root.toJSON()).toMatchInlineSnapshot(`null`);
```

packages/components/__tests__/__snapshots__/script.tests.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Script should not render a script tag without code or src 1`] = `null`;
```
DAreRodz: This file wouldn't be needed then.

DAreRodz (approved):
Code is fine and tests are great. Thanks @cristianbote!

I suggested just a minor change, check it out. :point_up:

PS: The tests seemed strange to me at first because you are accessing the DOM and using `react-test-renderer` ― which doesn't modify the DOM. Then I realize it's the `Script` component the one who does it, so :ok_hand:.

packages/components/__tests__/script.tests.tsx
cristianbote: Oh, such a good idea! 😄  Thank you!

packages/components/__tests__/__snapshots__/script.tests.tsx.snap
cristianbote: Yup! Removing it.

cristianbote:
> which doesn't modify the DOM

I mean, not the _real_ DOM indeed, but the underlaying, globally defined DOM via jsdom.

PR #825, opened 2021-05-20
Make mars-theme AMP-aware

### LINKED ISSUE #708: Make mars-theme AMP-aware [13pt]
This is needed for [AMP feature](https://community.frontity.org/t/amp-package/388/). For full context please check out the [final Implementation proposal](https://community.frontity.org/t/amp-package/388/10?u=santosguillamot).

---

The HTML that is inside of `post.content.rendered` is outside of the scope of the theme because it's managed by Html2React, but from the theme perspective, it should contain AMP-aware components. We could make the `mars-theme` starter them AMP-aware.

michalczaplinski:
**What**:

Make mars-theme AMP-aware. Fixes #708 .

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] End to end tests
- [x] Update starter themes

**Unrelated Tasks**

- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update other packages

**Additional Comments**


SantosGuillamot:
I was trying to make the `twentytwenty-theme` AMP-aware and I came up with some questions regarding this Pull Request.

1. I think [the way we are adding the canonical link](https://github.com/frontity/frontity/blob/facd9b2e682effaa9d023378f7798afcaa18b615/packages/mars-theme/src/components/index.js#L28-L30) isn't safe. We are assuming that the canonical link will be `{state.frontity.url + data.link}`, but that isn't always the case. Users could have a post/page pointing to an external domain or to a different url in their same website. Is there a way to check if a `<link rel="canonical" />` exists in the html and, only if it doesn't, add it ourselves for AMP.

2. As JavaScript doesn't work in AMP, the toggle menu button (used in mobile) doesn't work. Maybe we shouldn't show the button if `mode === amp` and we could show the navigation links in the footer. What do you think?

3. This isn't 100% related to AMP. If the WordPress image has a `width` and `height` defined, shouldn't we pass them in [this component](https://github.com/frontity/frontity/blob/facd9b2e682effaa9d023378f7798afcaa18b615/packages/mars-theme/src/components/featured-media.js#L33-L37)? Something like this:
```jsx
<StyledImage
        alt={media.title.rendered}
        src={media.source_url}
        srcSet={srcset}
        width={media.media_details.width}
        height={media.media_details.height}
/>
```

luisherranz:
> As JavaScript doesn't work in AMP, the toggle menu button (used in mobile) doesn't work. Maybe we shouldn't show the button if mode === amp and we could show the navigation links in the footer. What do you think?

I think we used the `amp-sidebar` in the past (https://amp.dev/documentation/components/amp-sidebar/).

We could also use `amp-bind` to toggle the menu (https://amp.dev/documentation/examples/components/amp-bind/). Maybe it's simpler.

luisherranz:
There you go 🙂

```html
<button
  [text]="isMenuOpen ? 'Close Menu' : 'Open Menu'"
  on="tap:AMP.setState({ isMenuOpen: !isMenuOpen })"
>
  Open Menu
</button>

<ul hidden [hidden]="!isMenuOpen">
  <li>Item 1</li>
  <li>Item 2</li>
  <li>Item 3</li>
  <li>Item 4</li>
</ul>
```

[Playground](https://playground.amp.dev/#share=PCFkb2N0eXBlIGh0bWw+CjxodG1sIOKaoSBsYW5nPSJlbiI+Cgo8aGVhZD4KICA8bWV0YSBjaGFyc2V0PSJ1dGYtOCI+CiAgPHRpdGxlPiBIZWxsbyBXb3JsZDwvdGl0bGU+CiAgPHNjcmlwdCBhc3luYyBzcmM9Imh0dHBzOi8vY2RuLmFtcHByb2plY3Qub3JnL3YwLmpzIj48L3NjcmlwdD4KICA8c2NyaXB0IGFzeW5jIGN1c3RvbS1lbGVtZW50PSJhbXAtYmluZCIgc3JjPSJodHRwczovL2Nkbi5hbXBwcm9qZWN0Lm9yZy92MC9hbXAtYmluZC0wLjEuanMiPjwvc2NyaXB0PgoKCiAgPGxpbmsgcmVsPSJjYW5vbmljYWwiIGhyZWY9Imh0dHBzOi8vYW1wLmRldi9kb2N1bWVudGF0aW9uL2V4YW1wbGVzL2ludHJvZHVjdGlvbi9oZWxsb193b3JsZC9pbmRleC5odG1sIj4KCiAgPG1ldGEgbmFtZT0idmlld3BvcnQiIGNvbnRlbnQ9IndpZHRoPWRldmljZS13aWR0aCI+CgogIDxzdHlsZSBhbXAtYm9pbGVycGxhdGU+Ym9keXstd2Via2l0LWFuaW1hdGlvbjotYW1wLXN0YXJ0IDhzIHN0ZXBzKDEsZW5kKSAwcyAxIG5vcm1hbCBib3RoOy1tb3otYW5pbWF0aW9uOi1hbXAtc3RhcnQgOHMgc3RlcHMoMSxlbmQpIDBzIDEgbm9ybWFsIGJvdGg7LW1zLWFuaW1hdGlvbjotYW1wLXN0YXJ0IDhzIHN0ZXBzKDEsZW5kKSAwcyAxIG5vcm1hbCBib3RoO2FuaW1hdGlvbjotYW1wLXN0YXJ0IDhzIHN0ZXBzKDEsZW5kKSAwcyAxIG5vcm1hbCBib3RofUAtd2Via2l0LWtleWZyYW1lcyAtYW1wLXN0YXJ0e2Zyb217dmlzaWJpbGl0eTpoaWRkZW59dG97dmlzaWJpbGl0eTp2aXNpYmxlfX1ALW1vei1rZXlmcmFtZXMgLWFtcC1zdGFydHtmcm9te3Zpc2liaWxpdHk6aGlkZGVufXRve3Zpc2liaWxpdHk6dmlzaWJsZX19QC1tcy1rZXlmcmFtZXMgLWFtcC1zdGFydHtmcm9te3Zpc2liaWxpdHk6aGlkZGVufXRve3Zpc2liaWxpdHk6dmlzaWJsZX19QC1vLWtleWZyYW1lcyAtYW1wLXN0YXJ0e2Zyb217dmlzaWJpbGl0eTpoaWRkZW59dG97dmlzaWJpbGl0eTp2aXNpYmxlfX1Aa2V5ZnJhbWVzIC1hbXAtc3RhcnR7ZnJvbXt2aXNpYmlsaXR5OmhpZGRlbn10b3t2aXNpYmlsaXR5OnZpc2libGV9fTwvc3R5bGU+PG5vc2NyaXB0PjxzdHlsZSBhbXAtYm9pbGVycGxhdGU+Ym9keXstd2Via2l0LWFuaW1hdGlvbjpub25lOy1tb3otYW5pbWF0aW9uOm5vbmU7LW1zLWFuaW1hdGlvbjpub25lO2FuaW1hdGlvbjpub25lfTwvc3R5bGU+PC9ub3NjcmlwdD4KPC9oZWFkPgoKPGJvZHk+CgogIDxidXR0b24gW3RleHRdPSJpc01lbnVPcGVuID8gJ0Nsb3NlIE1lbnUnIDogJ09wZW4gTWVudSciIG9uPSJ0YXA6QU1QLnNldFN0YXRlKHsgaXNNZW51T3BlbjogIWlzTWVudU9wZW4gfSkiPgogICAgT3BlbiBNZW51CiAgPC9idXR0b24+CgogIDx1bCBoaWRkZW4gW2hpZGRlbl09IiFpc01lbnVPcGVuIj4KICAgIDxsaT5JdGVtIDE8L2xpPgogICAgPGxpPkl0ZW0gMjwvbGk+CiAgICA8bGk+SXRlbSAzPC9saT4KICAgIDxsaT5JdGVtIDQ8L2xpPgogIDwvdWw+Cgo8L2JvZHk+CjwvaHRtbD4=)

packages/mars-theme/src/components/post.js
```diff
@@ -146,8 +146,7 @@ const Content = styled.div`
 
   figure {
     margin: 24px auto;
-    /* next line overrides an inline style of the figure element. */
-    width: 100% !important;
```
DAreRodz, line 150: I hope nothing breaks removing that `!important`... :sweat_smile:

packages/mars-theme/src/components/featured-media.js
```diff
@@ -35,6 +44,11 @@ export default connect(FeaturedMedia);
 const Container = styled.div`
   margin-top: 16px;
   height: 300px;
+  ${({ isAmp }) =>
+    isAmp &&
+    css`
```
DAreRodz: I tought you could use a string here instead of `css`. :thinking:

DAreRodz (approved):
Everything great! :100:

michalczaplinski:
> I think the way we are adding the canonical link isn't safe. We are assuming that the canonical link will be {state.frontity.url + data.link}, but that isn't always the case. Users could have a post/page pointing to an external domain or to a different url in their same website. 

@SantosGuillamot You're right, `state.frontity.url + data.link` might not be correct link in all cases.

> Is there a way to check if a <link rel="canonical" /> exists in the html and, only if it doesn't, add it ourselves for AMP. 

I'm not sure I understand what you mean by the "in the html"? Do you mean in the `post.content` ? I don't think there is any way that a `post.content` can contain a `<head>` tag. Or did you mean "elsewhere in the Frontity theme"? In that case, I don't think there is such a mechanism. Normally, with `react-helmet-async` you can render a `<Head>` anywhere in your component tree. And whatever you pass in there will end up in the document `<head>`. In case that there are conflicts, the element that is [lower in the component tree takes priority](https://css-tricks.com/its-all-in-the-head-managing-the-document-head-of-a-react-powered-site-with-react-helmet/#the-escape-hatch).

All that said, I think that perhaps the most reasonable thing to do would be not to add the canonical link if we cannot provide a reasonable default. If someone is developing for AMP, they will get a warning in the console and will have to deliberately set the canonical link in their theme. Which is better than having a default which would suppress the warning but mask a potential problem.

SantosGuillamot:
> I'm not sure I understand what you mean by the "in the html"? Do you mean in the post.content ?

No, I meant somewhere else in the Frontity application. If you are using our head-tags package for example.

> All that said, I think that perhaps the most reasonable thing to do would be not to add the canonical link if we cannot provide a reasonable default. 

Yes, I agree. If we can't ensure we are going to add the correct link I think it's better to not add it and leave it to the user.

michalczaplinski:
> This isn't 100% related to AMP. If the WordPress image has a width and height defined, shouldn't we pass them in this component? Something like this:

Oh, nice catch! This is gonna help with AMP as well actually as well. Because if the image has width and height and we pass it as props then in the AMP mode the image will render using the `responsive` layout which is the best option.

michalczaplinski:
> As JavaScript doesn't work in AMP, the toggle menu button (used in mobile) doesn't work. Maybe we shouldn't show the button if mode === amp and we could show the navigation links in the footer. What do you think?

I've tried the approach suggested by Luis and I think it could work nicely but it has a slight drawback which is that we'd have to include the [amp-bind](https://amp.dev/documentation/components/amp-bind/) script either in the mars-theme (which is wasteful if the user is not using AMP) or in the AMP package (which is wasteful if the user does not need the functionality of `amp-bind`).

I think the best way is to go with the old-school "hidden checkbox" technique. You make the hamburger menu icon an `<input type="checkbox">` and then style it with CSS to look like a hamburger icon. And then in the CSS you can do sth like: 

```css

.mobile-menu:checked ~ .menu-modal { 
  position: absolute;
  display: block;
  /* ...other styles */
}

```

so if the checkbox is "checked", you show the menu and hide if when it's unchecked. The only drawback of doing that is that we cannot set the:

```js
{/* Add some style to the body when menu is open to prevent body scroll */}
<Global styles={{ body: { overflowY: "hidden" } }} />
```

to prevent scrolling of the document body, but that's a minor detail.

michalczaplinski:
https://www.loom.com/share/684f996b73ff4244911e7b6efddaffd0

luisherranz:
> we'd have to include the amp-bind script either in the mars-theme (which is wasteful if the user is not using AMP) or in the AMP package (which is wasteful if the user does not need the functionality of amp-bind)

It's not that I dislike your alternative solution or something, but what is wrong with conditionally loading the AMP script? Isn't that what we do for all the AMP-aware components?

```jsx
const MobileMenu = () =>
  state.frontity.mode === "amp" ? (
    <>
      <Head>
        <script
          async
          custom-element="amp-bind"
          src="https://cdn.ampproject.org/v0/amp-bind-0.1.js"
        ></script>
      </Head>

      // ...
    </>
  ) : (
    // ...
  );
```

Regarding the other question, as this code is going to be part of _how we teach AMP in Frontity to people_, I think the goal here is to **reuse as much HTML and CSS as possible** between the AMP version and the React version. So I would choose whatever solution let us reuse more code 🙂

michalczaplinski:
> what is wrong with conditionally loading the AMP script?

oh, I'm not sure why I didn't do it this way in the first place 🙈. That's obviously the better solution and more idiomatic Frontity.

A small caveat: You have to use the `data-amp-bind-property="value"` syntax instead of just `[property]="value"` because [it's not valid JSX](https://amp.dev/documentation/components/amp-bind/#react-and-xml-compatibility)

michalczaplinski:
https://www.loom.com/share/88fbcf38b4994c9bbef1188cc0adb2ef

packages/mars-theme/src/components/post.js
michalczaplinski: I could not see notice anything breaking after removing it and as @SantosGuillamot has pointed out almost nobody uses unmodified mars-theme so I don't think this would break anything.

To clarify, this is being removed as AMP prohibits the use of `!important` in CSS.

michalczaplinski:
Would you like to review this again @luisherranz? 🙂 

I know that @DAreRodz has already approved, but I've addressed the issues you pointed out specifically.

packages/mars-theme/src/components/menu.js
```diff
@@ -1,10 +1,51 @@
...
+          role="button"
+          tabindex="0"
+          data-amp-bind-hidden="isMenuOpen"
+          on="tap:AMP.setState({ isMenuOpen: true })"
+        />
+        <CloseIcon
+          color="white"
+          size="20px"
+          role="button"
+          tabindex="0"
```
michalczaplinski, line 40: AMP validator also checks accessibility as of late, so we need to add those.

luisherranz (approved):
Sorry for being so late to this.

I've taken a quick look and everything seems great to me 🙂

@santosguillamot, I guess that we could add a second site to our mars.frontity.org (`projects/mars-theme`) to add support for `/amp` URLs so we can show case it. Or do you have any other plans for that?

PR #826, opened 2021-05-21
Make the `Image` component AMP-aware

michalczaplinski:
**What**:

Make the `@frontity/components/Image` AMP-aware and remove the Intersection Observer fallback which used to lazily load images in the absence of native lazy loading.

**Why**:

To complete making AMP a first-class citizen in Frontity: https://community.frontity.org/t/amp-package/388/

**How**:

Check if the `state.frontity.mode === "amp"`

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests

**Unrelated Tasks**

- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

Not linking the issue because I don't want to close it yet when this PR is merged.


michalczaplinski:
Now, that the `mode` of the Frontity app can be set to a new value (`"amp"`) I've noticed that in some places the default value was `"default"` but in others it was referred to as `"html"`.

- it's `"html"` in [`merge-packages.ts`](https://github.com/frontity/frontity/blob/bcb5c04dc19270fd99d9c1fcfb376ee64c2d5fc2/packages/core/src/utils/__tests__/merge-packages.tests.tsx#L12-L12) and [`file-settings/types.ts`](https://github.com/frontity/frontity/blob/31d34e4d16042ebc73b4c30a9f53dfae36c633a2/packages/file-settings/types.ts#L31-L31) and some other tests files.
- it's `"default"` pretty much everywhere else.

In order to avoid bikeshedding about it, I went ahead and changed the references to `"html"` mode to `"default"`

michalczaplinski:
https://www.loom.com/share/34fda69daca84616a02a1ac36969bc30

packages/components/image.tsx
```diff
@@ -1,176 +1,69 @@
...
-  srcSet,
-  sizes,
-  className,
+/**
+ * The Image component.
+ *
+ * @param props - Defined in {@link ImageProps}.
+ * @returns A React component.
+ */
+const Image: React.FC<Connect<Frontity, ImageProps>> = ({
```
luisherranz: When using `connect` with `injectProps: false` you don't need to use `Connect` to type the component, because it doesn't receive the store 🙂

```suggestion
const Image: React.FC<ImageProps> = ({ 
```

packages/components/image.tsx
```diff
@@ -1,176 +1,69 @@
-/* eslint-disable jsx-a11y/alt-text */
-
 import * as React from "react";
```
luisherranz, line 1: Is that type of wildcard import safe for the tree-shaking/unused-code-elimination algorithms?

Maybe we could change it for named imports...

luisherranz (approved):
Looks great to me! I love when we remove code 😄

Thanks Michal!

packages/components/image.tsx
michalczaplinski: I had to double check this because I was not 100% sure myself. It appears that whether you use the namespace import or the named import makes no difference. Both will bring the whole module into your code. Which makes sense, because even if you just import one function from a library, like `import { someFunc } from "library";` there is no guarantee that this function is not relying on some side-effect inside of `"library"`. 

In any case, it's up to the creator of the `"library"` to make sure that it's tree shakeable using the [`sideEffects`](https://webpack.js.org/configuration/optimization/#optimizationsideeffects)  or [`/*#__PURE__*/` annotation](https://webpack.js.org/guides/tree-shaking/#mark-a-function-call-as-side-effect-free).

Everything above only holds true for ES Modules, however. If using CommonJS I think the tree-shaking story is more complicated - I haven't looked into it in detail. React is still not shipping the ES Modules build by the way until at least React 18: https://github.com/facebook/react/issues/11503#issuecomment-662647468

- https://webpack.js.org/guides/tree-shaking

- https://stackoverflow.com/questions/58741044/why-webpack-doesnt-tree-shake-the-lodash-when-using-import-as

- https://github.com/webpack/webpack/issues/2713

luisherranz: I see. Thanks for the explanation Michal 🙂

PR #830, opened 2021-05-25
Transpile packages that contain webpack in the name

### LINKED ISSUE #829: Packages that contain `webpack` in the name are not transpiled
**_Bug report_**

**Expected behavior**

A package that contains `webpack` in the name, like for example `my-webpack-package` should be transpiled.

**Observed behavior**

They are not transpiled.

- Create a package with `webpack` in the name.
- Try to use it.

**Possible solution**

The problem is that we exclude `webpack` itself from being transpiled in this line: https://github.com/frontity/frontity/blob/dev/packages/core/src/config/webpack/modules.ts#L47. The current regular expression used `/\bwebpack\b` doesn't differentiate between webpack and Frontity packages that contain the `webpack` word.

The solution is to simply improve the regular expression.

luisherranz:
**What**:

Transpile packages that contain `webpack` in their name.

**Why**:

Because right now, due to the regular expresion that prevents the `webpack` package from being transpiled, all the packages that contain `webpack` are also excluded.

**How**:

Improve the regular expresion to exclude only the `/webpack/` package.

**Tasks**:

- [x] Code
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages


nicholasio:
should it also exclude webpack's plugin from transpilation?

luisherranz:
> should it also exclude webpack's plugin from transpilation?

Webpack's plugins are not imported in the Frontity app, so we don't need to exclude them.

The reason we need to exclude `webpack` is because it is actually imported by Webpack itself during the bundling: https://github.com/zloirock/core-js/issues/743#issuecomment-572074215 (it is a dependency of `core-js`).

nicholasio (approved):

luisherranz:
I'm not sure what happened here. I merged the PR (https://github.com/frontity/frontity/commit/4adb767c2b39af28c22c3312ddec10c8d76d37af) but the GH UI doesn't show it as merged 😕

luisherranz:
Solved now 🤷

PR #832, opened 2021-05-27
Update the default value of `mode` to be consistently `"default"` instead of `"html"`

michalczaplinski:
**What**:

The mode of a Frontity app can now be set to a new value ("amp") I've noticed that in some places the default value was `"default"` but in others it was referred to as `"html"`.

For example: 

- it's `"html"` in [`merge-packages.ts`](https://github.com/frontity/frontity/blob/bcb5c04dc19270fd99d9c1fcfb376ee64c2d5fc2/packages/core/src/utils/__tests__/merge-packages.tests.tsx#L12-L12) and [`file-settings/types.ts`](https://github.com/frontity/frontity/blob/31d34e4d16042ebc73b4c30a9f53dfae36c633a2/packages/file-settings/types.ts#L31-L31) and some other tests files.
- it's `"default"` pretty much everywhere else.

I've updated the codebase so that the default value is consistently `"default"`.

**Why**:

Consistency.

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


michalczaplinski:
https://www.loom.com/share/55cc16c32a20417080ba56f3c30b47de

luisherranz (changes requested):
Actually, I am sorry to say this because it's going to involve some undoing, but could please make the `mode` type a `string` instead of `"default" | "amp"`? 

The reason is that, when I added that type in Frontity 1.0, the "amp" mode was going to be hardcoded. But now AMP is going to be a package, so I think it makes sense to make `mode` simply a string so if someone wants to create another package with another mode they can.

Thanks Michal! 😅

michalczaplinski:
oh, okay, I see. Yup, no problem 🙂

michalczaplinski:
I've removed the changeset since the only changes after all are test files and minor refactoring of types (but without actually changing the types)

luisherranz (approved):
I have done a more generalistic explanation of the `mode` because it is not only for AMP, but other than that it's great Michal 🙂👍

PR #834, opened 2021-05-28
Fix with-preact to be defined as commonjs

cristianbote:
**What**:

Change the format for `@frontity/with-preact` config file to commonjs since it's not getting transpiled if installed and use from `node_modules`.

**Why**:

It breaks compilation otherwise.

**Tasks**:

- [x] Code
- [x] Update other packages

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes

**Additional Comments**

### DIFF
--- a/.changeset/shaggy-laws-sniff.md
+++ b/.changeset/shaggy-laws-sniff.md
@@ -1,5 +1,5 @@
 ---
-"@frontity/with-preact": minor
+"@frontity/with-preact": patch
 ---
 
-with-preact is not usually transpiled when is part of the node_modules so the config file should be defined as commonjs
+Use `commonjs` in the `frontity.config.js` file until we add support for modules.
--- a/.changeset/shaggy-laws-sniff.md
+++ b/.changeset/shaggy-laws-sniff.md
@@ -1,0 +1,5 @@
+---
+"@frontity/with-preact": minor
+---
+
+with-preact is not usually transpiled when is part of the node_modules so the config file should be defined as commonjs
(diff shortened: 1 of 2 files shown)


.changeset/shaggy-laws-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontity/with-preact": minor
+---
+
+with-preact is not usually transpiled when is part of the node_modules so the config file should be defined as commonjs
```
luisherranz:
```suggestion
"@frontity/with-preact": patch
---

Use `commonjs` in the `frontity.config.js` file until we add support for modules.
```

luisherranz (approved):
Hey Cris, I have been talking about this with @nicholasio and I think we should find a way to transpile those config files, shouldn't we?

I'm going to open an issue.

We can merge this PR until we fix that so the package is usable 🙂

cristianbote:
Hey @luisherranz thanks for the review. Yeah, it'll be best to define a rule frontity.config files.

luisherranz:
Issue opened: https://github.com/frontity/frontity/issues/835

nicholasio:
> Hey Cris, I have been talking about this with @nicholasio and I think we should find a way to transpile those config files, shouldn't we?
> 
> I'm going to open an issue.
> 
> We can merge this PR until we fix that so the package is usable 🙂

Definitely think we should fix this. The TS defs are really nice for modifying the webpack config.

PR #836, opened 2021-05-28
Typescript version of Mars Theme

luisherranz:
**What**:

This adds a new theme called `mars-theme-typescript` that is basically mars theme, but written in Typescript.

We should also modify the `npx frontity create` template so if the user chooses typescript, this theme is installed.

**Why**:

We don't have an example of a theme made with Typescript yet.

**How**:

- Adding the new theme as a new package.
- Modifying the `create` CLI command.

**Tasks**:

- [x] Add `mars-theme-typescript` package
- [x] TSDocs
- [x] Update `create` CLI command
- [x] CLI tests
- [x] Update community discussions (https://community.frontity.org/t/full-typescript-support-in-frontity-create/880)

**Unrelated Tasks**

- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

These initial changes were done by @darerodz and me, and this is still a work in progress 🙂


luisherranz:
@michalczaplinski this is ready for review 🎉 

We recorded a video for you:  https://www.youtube.com/watch?v=WlqJb_CdrWA

michalczaplinski (approved):
Okay, looks good to me! I've just reformatted the TSDocs in b129476246c5e38d3d34437b4a118b12ef988c57 because they were still wonky.

I'm not sure why but on first try, husky formatted it back to the "wrong" format before commiting. Tried again and it seems to have worked. Just FYI.

michalczaplinski:
Do we then still need the [`types.ts`](https://github.com/frontity/frontity/blob/dev/packages/mars-theme/types.ts) in the (non-typescript) mars-theme?

luisherranz:
> I'm not sure why but on first try, husky formatted it back to the "wrong" format before commiting. Tried again and it seems to have worked. Just FYI.

Yeah, I'm not sure what's going on there. Did we update `lint-staged` or `husky`?

> Do we then still need the types.ts in the (non-typescript) mars-theme?

I don't know, to be honest. Maybe it's not bad to have expose a type definition for your package, even if the package is done in JS, because that means that other packages made in TS can include those types in their `Packages` export 🤷

I'd vote for leaving it there as an example of that 🙂

michalczaplinski:
> Did we update `lint-staged` or `husky`?

Nope, last update was a year ago according to `git blame`. Maybe it was a fluke, but if it happens again, we could try updating them like I did in https://github.com/frontity/frontity.org/pull/192#issuecomment-875965953


> I'd vote for leaving it there as an example of that 🙂

ok 👍

PR #837, opened 2021-05-31
Transpile all frontity.settings/config imports

### LINKED ISSUE #835: Transpile `frontity.config.js` files
**_Bug report_**

**Expected behavior**

The `frontity.config.js` files should be transpiled on the fly.

**Observed behavior**

Right now they are transpiled only if the package is in the local folder, but as soon as you publish the package in npm and install it in `node_modules`, you are forced to use `commonjs` for them:

```js
module.exports = {
  webpack: ({ config }) => {
    // ...
  },
};
```

Instead of:

```js
export const webpack = ({ config }) => {
  // ...
};
```

Create a package in the `node_modules` folder (using `npm link` for example) and add a `frontity.config.js` file.

**Possible solution**

I'm not sure why they are not transpiled if they are in `node_modules`. I'll try to triage the problem and add more information.

luisherranz:
**What**:

Change the files that the `ts-node` that we run in the `dev` and `build` commands transpile.

**Why**:

Because it was not transpiling files that were inside the `node_modules` folder, like `frontity.config.js` files in external packages.

**How**:

I created a regular expression that is passed to the `ignores` setting of `ts-node` and makes it transpile only:

- Packages from the `@frontity` org
- Packages with `frontity` in the name (for example `some-frontity-package`).
- `frontity.settings` files
- `frontity.config` files
- Any TypeScript files that people import.

By default this setting was `node_modules`.

_This `ts-node` transpilation is only used outside of Webpack, which means mostly things people import in their `frontity.settings` or `frontity.config` files. Once Webpack runs, all the files are transpiled._

**Tasks**:

- [x] Code
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

### DIFF
--- a/.changeset/modern-hounds-tell.md
+++ b/.changeset/modern-hounds-tell.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Fix transpilation of `frontity.config.js` files in external packages.
--- a/packages/core/src/scripts/build.ts
+++ b/packages/core/src/scripts/build.ts
@@ -20,6 +20,7 @@ import * as tsNode from "ts-node";
  */
 tsNode.register({
   transpileOnly: true,
+  ignore: [],
   compilerOptions: {
     // Target latest version of ECMAScript.
     target: "es2017",
--- a/packages/core/src/scripts/build.ts
+++ b/packages/core/src/scripts/build.ts
@@ -1,3 +1,4 @@
+import ignore from "./utils/ts-node-ignore";
 import * as tsNode from "ts-node";
 
 /**
@@ -20,7 +21,7 @@ import * as tsNode from "ts-node";
  */
 tsNode.register({
   transpileOnly: true,
-  ignore: [],
+  ignore,
   compilerOptions: {
     // Target latest version of ECMAScript.
     target: "es2017",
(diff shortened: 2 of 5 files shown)


michalczaplinski:
I'm not sure why the CLI tests keep failing occasionally and only on MacOS. I thought this should have been fixed in https://github.com/frontity/frontity/pull/824.

luisherranz:
True. It seems like this time was a timeout. We'll keep an eye on them 🙂

I have rerun them and they passed this time.

DAreRodz (approved):
Seems just fine to me. :+1: 

Although for me it's not clear the part of not ignoring packages with `frontity`. Would it mean that some frontity packages (those imported by frontity settings/config files) should contain `frontity` in their name?

luisherranz:
It is a compromise between transpiling everything (which is pretty slow) and not transpiling anything, which could lead to problems like the one this PR is trying to solve. I tried to come up with a set of scenarios where transpiling seems more right than wrong.

EDIT: I think I didn't address your question: 

> Would it mean that some frontity packages (those imported by frontity settings/config files) should contain frontity in their name?

The answer is no, they don't need to contain `frontity` in their names for a normal use 🙂

PR #841, opened 2021-06-02
Fix CLI-e2e tests

michalczaplinski:
**What**:

Fix the issue of intermittently failing e2e CLI tests

**How**:

Add the `.DS_Store` files to the list of files that a directory where you want to run `npx frontity create` can contain.

**Tasks**:

- [x] Code
- [x] Add a changeset (with link to its [Feature Discussion]

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] End to end tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


michalczaplinski:
https://www.loom.com/share/d101f853747c4ce981233614d28cfc1e

luisherranz:
Thanks for the PR Michal :)

As you said, running a new container for each individual test doesn't scale well. Why don't we use a different temporary folder for each test? We could use something like https://github.com/sindresorhus/tempy.

```js
import tempy from "tempy";

it("should create the initial files", async () => {
  const tempFolder = tempy.directory();

  await execa.command(
    `cd ${tempFolder} && ts-node src/cli/index.ts create --no-prompt --theme @frontity/mars-theme --cwd`,
    { stdio: "inherit" }
  );

  expect(await readdir(tempFolder))...
});
```

luisherranz:
How did it go? Do you think it's a better approach? 😄

By the way, I think we can maintain the increased timeout. I've seen them fail because of timeouts as well, don't you think?

michalczaplinski:
> How did it go? Do you think it's a better approach? 😄

I'm really not sure at this point. In particular, I think that on windows `cd /some/path && some_command` does not work the same as on unix: https://github.com/frontity/frontity/pull/841/checks?check_run_id=2750546338#step:6:42

It seems that the temp directory is on `C:/` whereas the repo is on `D:/` 


> By the way, I think we can maintain the increased timeout. I've seen them fail because of timeouts as well, don't you think?

yup, agreed 👍

michalczaplinski:
It occurred to me that perhaps the e2e-cli failures were caused because sometimes the `test-fronity-app` folder in the github actions container contains the `.DS_Store` files. Which would explain why the failures were only occurring in the MacOS containers and also explain why they didn't always happen!

luisherranz (approved):
Ok, let's see if this helps! 🙂

PR #843, opened 2021-06-03
Refactor `createSymlinks` function in `@frontity/core`

DAreRodz:
**What**:

Replace the [`fs-extra`](https://www.npmjs.com/package/fs-extra)'s `readFile` function by an in-house implementation.

**Why**:

That function seems to fail when using [`mock-fs`](https://www.npmjs.com/package/mock-fs) and Node 16.

**How**:

Use the standard `fs` library to implement a new `readFile` function that works with `mock-fs`.

**Tasks**:

- [x] Code

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

I'm not sure if there is a problem in the `mock-fs` library, in the `fs-extra`, or it is just that they stopped to be compatible when Node 16 is used. Making some tests to isolate the problem I discovered that it was in the `readFile` function of `fs-extra`, which was throwing this error:

```
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[Error: EINVAL: invalid argument, read] {
  errno: -22,
  code: 'EINVAL',
  syscall: 'read'
}
```

The easiest solution I've found was to replace that function by a new one, using `fs/promise`. That seems to work fine (tested in node 10 and 16).

### DIFF
--- a/.changeset/wise-shrimps-explode.md
+++ b/.changeset/wise-shrimps-explode.md
@@ -1,0 +1,5 @@
+---
+"@frontity/core": patch
+---
+
+Refactor `createSymlinks` to use a custom implementation of `fs-extra`'s `readFile` function.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -55,7 +55,7 @@
     "deepmerge": "^4.2.2",
     "express": "^4.17.1",
     "file-loader": "^6.2.0",
-    "fs-extra": "^9.0.1",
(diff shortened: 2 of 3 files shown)


michalczaplinski (approved):
LGTM 👍

PR #850, opened 2021-06-18
Infinite loop when client/server URL queries are sorted in the server

luisherranz:
**What**:

Fix an infinite loop created when the URL in the browser and the URL that was received in the server is not the same.

**Why**:

It seems like Vercel started sorting the queries in their cache, causing this problem.

**How**:

I have created a failing e2e test that reproduces the issue. I'll take a look at how to prevent the problem.

**Tasks**:

- [x] End to end tests
- [x] Code
- [x] Update community discussions (https://community.frontity.org/t/problem-with-links-with-utm-campaings/3903/11)
- [x] Add a changeset

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


luisherranz:
@Michalczaplinski, this is the explanation of the PR https://www.loom.com/share/bb3201c04a9649ebbaaf0169b0bef4d8. If you have questions let me know 🙂

michalczaplinski (approved):
LGTM 👍

luisherranz:
Thanks Michal!

Issue #695, opened 2021-02-08
Create a clean Koa application per request


SantosGuillamot:
This is part of the [Server Extensibility feature](https://community.frontity.org/t/server-extensibility/2810). For full context please check out the [final Implementation proposal](https://community.frontity.org/t/server-extensibility/2810/23).

---

To make sure that we don't leak middleware and app configuration from one request to another, my proposal is to create a new Koa application in each request.

Right now this is our `@frontity/core/src/server` export:

```js
const server = ({ packages }) => {
  // Create new app.
  const app = new Koa();

  // Configure the app...
  app.use(/* ... */);

  // Return the req/res function.
  return app.callback();
};
```

We should switch to something like this:

```js
const server = ({ packages }) => (req, res) => {
  // Create a new app for each request.
  const app = new Koa();

  // Configure the app...
  app.use(/* ... */);

  // Return the final response.
  return app.callback()(req, res);
};
```

**Dependencies**: none.

**Relevant code**:

https://github.com/frontity/frontity/blob/dev/packages/core/src/server/index.tsx

nicholasio:
Assigning this first task over to me, planning on working on this in the next few days.

luisherranz:
Awesome 👏 👏 😄

PR #852, opened 2021-06-18
Add `server` middlewares defined in packages to the Koa app

### LINKED ISSUE #696: Make sure the middleware (server) is properly merged in the creation of the store
This is part of the [Server Extensibility feature](https://community.frontity.org/t/server-extensibility/2810). For full context please check out the [final Implementation proposal](https://community.frontity.org/t/server-extensibility/2810/23).

---

We think this should work almost out of the box, but we have to make sure that the `server` middleware is populated in the store.

[This is](https://github.com/frontity/frontity/blob/dev/packages/core/src/utils/merge-packages.ts) where the merge is happening right now. At least we need to add the `server` prop to the initialization:

```ts
let config: Package = {
  roots: {},
  state: {},
  actions: {},
  libraries: {},
  server: {},
};
```

Also, we need to make sure that if a package wants to overwrite the middleware of another package, it can do so by exporting the new version with the same namespace and name, in the same way that people can overwrite `state` and `actions` now.

```js
// frontity.settings.js
const settings = {
  name: "my-site",
  packages: ["some-package", "my-custom-package"],
};
```

```js
// some-package
export default {
  server: {
    somePackage: {
      someMiddleware: ({ state, ctx }) => {
        // Original implementation.
      },
    },
  },
};
```

```js
// my-custom-package
export default {
  server: {
    somePackage: {
      someMiddleware: ({ ctx }) => {
        // My custom overwrite!
      },
    },
  },
};
```

Again, that should work out of the box.

**Dependencies**: None

**Relevant code**:

https://github.com/frontity/frontity/blob/dev/packages/core/src/utils/merge-packages.ts

DAreRodz:
**What**:

Get middlewares exposed by Frontity packages in the `server` property and register them into the Koa server.

Co-authored-by: @cristianbote

**Why**:

To close #696 

**Tasks**:

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests

**Unrelated Tasks**

- [ ] Update starter themes
- [ ] Update other packages

### DIFF
--- a/packages/core/src/server/index.ts
+++ b/packages/core/src/server/index.ts
@@ -10,6 +10,7 @@ import { scriptsStats } from "./middlewares/scripts-stats";
 import { settingsAndStore } from "./middlewares/settings-and-store";
 import { serverSideRendering } from "./middlewares/server-side-rendering";
 import { errorHandling } from "./middlewares/error-handling";
+import { storeMiddlewares } from "./middlewares/store-middlewares";
 
 /**
  * Options for {@link server}.
@@ -99,6 +100,9 @@ const server = ({ packages }: ServerOptions): ReturnType<Koa["callback"]> => {
 
   // Setup the settings and store.
   app.use(settingsAndStore(packages));
+
+  // Add the server middlewares exposed by packages.
+  app.use(storeMiddlewares(app));
 
   // Sever Side Rendering with the defined
   // template and render method.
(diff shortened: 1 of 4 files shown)


DAreRodz:
After doing some minimal testing, I've noticed that middlewares are registered for each request, again and again. That would be fixed though once we implement https://github.com/frontity/frontity/issues/695.

@SantosGuillamot should I mark the issue as blocked?

SantosGuillamot:
Sure, I have just done it :slightly_smiling_face:

PR #853, opened 2021-06-23
Frontity Update command

luisherranz:
**What**:

I have been working with @santosguillamot on a new `npx frontity update` command.

**Why**:

Because now users need to follow several steps to update their Frontity project. It'd be nice if they could do it just by running a single command.

https://docs.frontity.org/guides/keep-frontity-updated

**How**:

By adding a new `npx frontity update` command that basically runs `npx npm-check-updates -u -t minor` on the root and all the local packages.

**Tasks**:

- [ ] Code
- [ ] End to end tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

This is still a work in progress.


luisherranz:
At first, we thought it'd be better to run `npm-check-updates` programmatically, but then we thought that for the first version it will be simpler to run it directly using `exec`.

PR #858, opened 2021-07-02
Another fix for the problem when browser/server links are different

luisherranz:
**What**:

Another fix for the patch that we do to check if the browser link is different than the link that was received in the server.

**Why**:

Because servers/services sitting in between the browser and the Frontity server can modify the URL. That happens with: Pantheon CDN, Vercel CDN, Stackblitz dev environment...

**How**:

We were not comparing the normalized links so they could end up causing an infinite loop. This PR adds an explicit check of the normalized server link.

https://www.loom.com/share/047154edea5f41f5843ee3ef4205d91d

**Tasks**:

- [x] Code
- [x] End to end tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

In general, this is a very weak patch. In the future, this should be replaced by adding support for canonical links in Source, so multiple links point to the same data object.


DAreRodz (approved):
Approved :+1:

PS: yeah, you're right. Tiny Router is getting uglier and doesn't seem _tiny_ anymore either. :sweat_smile:

It should be great to have that logic handled by `source` packages. which are the ones that should know if two different URLs are actually the same page.

luisherranz:
Thanks for the approval David 🙂

PR #866, opened 2021-09-14
Check for all returned posts in the post type handler instead of just the first one

### LINKED ISSUE #865: Child pages with same slug as top-level pages can result in 404.
**_Bug report_**

If you create a top-level page and a child page with the same slug (which WP allows) frontity won't be able to find and render both pages. Only the most recent page with the same slug will render.

**Expected behavior**

Both the top-level page and child page should render even if they share the same slug.

**Observed behavior**

Only the most recent page with the same slug renders. What happens is that when frontity finds a child page it only uses the slug to fetch the child page (which is expected) but WP will return both the top level page and the child page as both matches the slug. However, [frontity is always assuming](https://github.com/frontity/frontity/blob/dev/packages/wp-source/src/libraries/handlers/postType.ts#L85) that the desired page is the first one returned, which is not correct when there are multiple pages with the same slug. 

Steps to reproduce:
1 - Create a top-level page with `contact-us` as the slug The link for this page will be `domain.com/contact-us`
2 - Create a child page of any page (e.g `leadership` with `contact-us` as the slug. The link for this page will be `domain.com/leadership/contact-us`
3 - Vising `domain.com/contact-us` will result in a 404
4 - Visiting `domain.com/leadership/contact-us` will work.
5 - Delete and recreate `domain.com/contact-us`, now `domain.com/contact-us` renders and `domain.com/leadership/contact-us` is a 404.

**Possible solution**

Check if any of the posts returned from the rest API matches the route.

```
--- a/node_modules/@frontity/wp-source/src/libraries/handlers/postType.ts
+++ b/node_modules/@frontity/wp-source/src/libraries/handlers/postType.ts
@@ -82,7 +82,7 @@ const postTypeHandler = ({
       if (populated.length > 0) {
         // We have to check if the link property in the data that we
         // populated is the same as the current route.
-        if (populated[0].link === route) {
+        if (populated.some((post) => post.link === route)) {
           isHandled = true;
           isMismatched = false;
           matchedEndpoint = endpoint;
```

nicholasio:
**What**:

Fixes #865 

**Why**:

It's standard WP behavior

**How**:

By checking all of the returned posts instead of just the first one.

**Tasks**:

- [x] Code
- [x] Unit tests

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/.changeset/chilly-socks-care.md
+++ b/.changeset/chilly-socks-care.md
@@ -1,0 +1,5 @@
+---
+"@frontity/wp-source": patch
+---
+
+Fix a bug where frontity could not fetch a page or child page with the same slug as the parent/child.
--- a/packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/post-type.tests.ts.snap
+++ b/packages/wp-source/src/libraries/handlers/__tests__/__snapshots__/post-type.tests.ts.snap
@@ -1493,6 +1493,119 @@ Object {
 }
 `;
 
+exports[`page can have a parent with the same slug 1`] = `
+Object {
+  "api": "https://test.frontity.org/wp-json/",
+  "attachment": Object {},
+  "author": Object {
+    "1": Object {
+      "id": 1,
+      "link": "/author/author-1/",
+      "name": "Author 1",
+      "slug": "author-1",
+    },
+  },
+  "authorBase": "",
+  "category": Object {},
+  "categoryBase": "",
+  "data": Object {
+    "/author/author-1/": Object {
+      "id": 1,
+      "isFetching": false,
+      "isReady": false,
+      "link": "/author/author-1/",
+      "page": 1,
+      "query": Object {},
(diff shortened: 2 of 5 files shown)


luisherranz (changes requested):
Hey, nice catch @nicholasio 🙂

The code looks great to me, but could you also add a unit test and a changelog?

Thanks! 😄

nicholasio:
> Hey, nice catch @nicholasio 🙂
> 
> The code looks great to me, but could you also add a unit test and a changelog?
> 
> Thanks! 😄

Sure I'll get this updated with a changelog and a couple of tests.

luisherranz (approved):
Looks perfect to me! Thanks a bunch, Nícholas 🙂👏

PR #867, opened 2021-09-16
Purge expired artifacts

luisherranz:
**What**:

Purge the expired artifacts.

**Why**:

Because the Eslint GH action is failing a lot lately and it seems like expired artifacts can be the cause, acording to this issue: https://github.com/bradennapier/eslint-plus-action/issues/45.

**How**:

By adding a previous step in the Eslint GH action that purges the expired artifacts before eslint runs.

**Tasks**:

- [x] GH action

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -33,7 +33,8 @@ jobs:
         env:
           CI: true
           
-      - uses: LKP-RnD/purge-expired-artifacts-action@v4
+      - name: Delete expired artifacts
+        uses: LKP-RnD/purge-expired-artifacts-action@v4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           repo_to_purge: "frontity/frontity"
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -33,7 +33,7 @@ jobs:
         env:
           CI: true
           
-      - name: Delete expired artifacts
+      - name: Delete expired artifacts 
         uses: LKP-RnD/purge-expired-artifacts-action@v4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -32,6 +32,12 @@ jobs:
         run: npm ci
         env:
           CI: true
+          
+      - uses: LKP-RnD/purge-expired-artifacts-action@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          repo_to_purge: "frontity/frontity"
+          dry_run: "false"
 
       - name: Run Eslint
         uses: bradennapier/eslint-plus-action@v3.4.2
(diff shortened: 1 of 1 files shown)


luisherranz:
I keep hitting the rate limit. I'll wait a while and retrigger the purge action again.

luisherranz:
It seems like there are many artifacts, so maybe it'd be better to merge this and let it run each 15 minutes until they are all gone.

@santosguillamot, would mind taking a look and approving this PR? Thanks.

SantosGuillamot:
I've seen that in the GitHub action you added, [LKP-RnD/purge-expired-artifacts-action@v4](https://github.com/LKP-RnD/purge-expired-artifacts-action), they mention as a [common issue](https://github.com/LKP-RnD/purge-expired-artifacts-action#common-issues) this:

_"If you have thousands of expired artifacts the action won't be able to delete everything in one go as githubs API has a rate limit. However, if you run it on a schedule as in the example every other hour you will eventually come to a point where the script will have cleared most unused artifacts and this will no longer cause problems."_

Checking [the repo in GitHub artifacts API](https://api.github.com/repos/frontity/frontity/actions/artifacts), it seems that there are 4583. So I guess this could be the problem.

I guess we have to remove the artifacts manually or, as they suggest, implement a CRON workflow triggering the action you added.

luisherranz:
> I guess we have to remove the artifacts manually or, as they suggest, implement a CRON workflow triggering the action you added.

The CRON is already set up for this action. That is the reason I thought we may just merge this and let it fun each 15 minutes until they are all gone 🙂

SantosGuillamot:
> It seems like there are many artifacts, so maybe it'd be better to merge this and let it run each 15 minutes until they are all gone.

Oh, I was just writing my answer 🙂 . Triggering it removed like 300 artifacts, so we'll need it to run a few times until all of them are deleted.

> @SantosGuillamot, would mind taking a look and approving this PR? Thanks.

Sure, I was already on it. Approving it 👍

SantosGuillamot (approved):
LGTM 👍

luisherranz:
Haha, ok, we were talking about the same at the same time 😄

Thanks Mario, I'll merge it now and I'll keep an eye on the action runs after that.

luisherranz:
BTW, these are the action run by the workflow each 15 min: https://github.com/frontity/frontity/actions/workflows/eslint.yml

SantosGuillamot:
> BTW, these are the action run by the workflow each 15 min: https://github.com/frontity/frontity/actions/workflows/eslint.yml

I saw it in the changed files, thanks Luis 🙂 

Btw, I think I can trigger it manually until the artifacts are gone. There are 3884 left -> https://api.github.com/repos/frontity/frontity/actions/artifacts

luisherranz:
578 now! https://api.github.com/repos/frontity/frontity/actions/artifacts 😄 👏

PR #869, opened 2021-09-21
Change Eslint action

luisherranz:
**What**:

I've changed the GH action that we use for Eslint annotations.

**Why**:

The Eslint action we were using is definitely broken. I tried to fix it [here](https://github.com/frontity/frontity/pull/867), but it's not working.

**How**:

Just start using another GH action.

**Tasks**:

- [x] GH action

**Unrelated Tasks**

- [ ] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -1,17 +1,6 @@
 name: eslint
 
-on:
-  # By adding a schedule task to this workflow we will automatically
-  # begin serializing read-only runs and handling them. The cron job
-  # below is set to run every 15 minutes, GitHub will ignore anything
-  # under 10 minutes and run every 10 minutes anyway.
-  schedule:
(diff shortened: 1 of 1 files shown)


luisherranz:
Let's use a different Eslint action that actually works, even if it doesn't have support for external forks 🙂

Mario, would you mind reviewing this? Thanks.

SantosGuillamot (approved):
> Mario, would you mind reviewing this? Thanks.

Sure. It seems to be working now, although in this PR it didn't lint anything because there are no files with [.mjs, .js, .ts] extensions added or modified. It looks good to me! Thanks Luis 🙂

PR #877, opened 2021-10-05
Keeping original iframe props when using Frontity iframe component

### LINKED ISSUE #875: Iframe processors removes props from iframe.
**_Bug report_**

**Expected behavior**

The iframe processor should not strip out valid iframe props (such as custom data attrs, id, and class names for instance).

**Observed behavior**

The iframe processor will strip all of the props originally present in the iframe and will only add the following
```javascript
    "data-src": src,
    className: "frontity-lazy-iframe".concat(className ? ` ${className}` : ""),
    loading,
    style: { visibility: "hidden" },
    height,
    width,
```

1 - Add a iframe with an id to your post content.
2 - See that id is gone when the iframe processor is enabled.

**Info about your system**

**Possible solution**

```javascript
 const Iframe: Component = ({
-  state,
   src,
   title,
   width,
@@ -77,7 +76,9 @@ const Iframe: Component = ({
   className,
   loading = "lazy",
   rootMargin,
+  ...rest
 }) => {
+  const { state } = useConnect();
   const lazyAttributes: Attributes = {
     "data-src": src,
     className: "frontity-lazy-iframe".concat(className ? ` ${className}` : ""),
@@ -86,6 +87,7 @@ const Iframe: Component = ({
     height,
     width,
     title,
+    ...rest
   };
 
   const eagerAttributes = changeAttributes(lazyAttributes);
@@ -144,4 +146,4 @@ const Iframe: Component = ({
   );
 };
 
-export default connect(Iframe);
+export default connect(Iframe, { injectProps: false });
```

nicholasio:
**What**:

https://github.com/frontity/frontity/issues/875

**Why**:

Some code on the page might rely on arbitrary iframe props (like id) 

**How**:

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests

**Unrelated Tasks**

- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/packages/components/iframe.tsx
+++ b/packages/components/iframe.tsx
@@ -1,8 +1,9 @@
 /* eslint-disable jsx-a11y/iframe-has-title */
 import * as React from "react";
-import { connect, Head } from "frontity";
+import { useConnect, connect, Head } from "frontity";
 import { Connect, Package } from "frontity/types";
 import useInView from "@frontity/hooks/use-in-view";
+import { Packages } from "../amp/types";
 
 const noJsStyles = `
   :non(noscript) > .frontity-lazy-iframe {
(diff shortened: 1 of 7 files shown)


packages/components/iframe.tsx
```diff
@@ -24,29 +25,90 @@ const noProxyScript = `
   }
 `;
 
+/**
+ * The iframe type definitions.
+ */
 interface Props {
+  /**
```
luisherranz, line 32: What about extending the types of an iframe here as we do with the [Image props](https://github.com/frontity/frontity/blob/379a44761423b4e47466359bbfe598615a869dcb/packages/components/image.tsx#L8):

```suggestion
 */
interface Props extends React.IframeHTMLAttributes<HTMLIframeElement> {
  /**
```

(I haven't tested it)

luisherranz (approved):
It looks great to me Nícholas! Feel free to merge.

----

In the future, I think it'd be nice to add support for AMP and remove lazy-loading support using the Intersection Observer, as we did for the `Image` component:

- Image component: https://github.com/frontity/frontity/blob/dev/packages/components/image.tsx
- PR to remove IO support and make it AMP-aware: https://github.com/frontity/frontity/pull/826

The new version of the Image component is much simpler as well. I never liked the lazy-load IO fallback, to be honest 🙂

packages/components/iframe.tsx
nicholasio: Tried and it showed a couple of TS errors and didn't want to go deep into this as it's unrelated. But it totally make sense to do this 👍

luisherranz: Okey, no problem Nícholas 🙂 👍

PR #881, opened 2021-10-16
just get the name of currentPage selected, with the mars theme

### LINKED ISSUE #846: (mars theme) aria-current="page" should be retained on menu for eg /blog/page/N/
(sorry if this is not the place for this... should these minor issues just be raised on the forum?)

**_Bug report_**

**Expected behavior**

When visiting eg `/blog/page/2/` the "Blog" item should still be highlighted in the menu. 
this is currently created by
`aria-current={state.router.link === link ? "page" : undefined}`
and so only works on the first page

**Observed behavior**

Blog item is underlined in  menu for the first page, but not for subsequent

* go to https://5hok7.sse.codesandbox.io/blog/
* click "older posts"

rubenlorenzo:
**What**:

Keep current page selected when changing old pages. Fixes #846.

**Why**:

To know what page you are on. 

**How**:

It deletes **page/pageNumber** *to url* of the router, to compare the  url to link. So it can give a positive answer.

**Tasks**:

- [x] Code
- [ ] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests
- [ ] TypeScript tests
- [x] Update starter themes
- [ ] Update other packages

**Unrelated Tasks**

**Additional Comments**

### DIFF
--- a/packages/mars-theme/src/components/nav.js
+++ b/packages/mars-theme/src/components/nav.js
@@ -9,12 +9,9 @@ import Link from "./link";
 const Nav = ({ state }) => (
   <NavContainer>
     {state.theme.menu.map(([name, link]) => {
-      // Gets the page number
-      let pageNumber = state.router.link.replace(/\D/g, "");
-
-      // Check if the link matched the current page url, removing page + number from this
-      const isCurrentPage =
-        state.router.link.replace("page/" + pageNumber + "/", "") === link;
+      // Check if the link matched the current page url
+      const data = state.source.get(state.router.link);
+      const isCurrentPage = data.route === link;
 
       return (
         <NavItem key={name}>
--- a/packages/mars-theme/src/components/nav.js
+++ b/packages/mars-theme/src/components/nav.js
@@ -9,8 +9,13 @@ import Link from "./link";
 const Nav = ({ state }) => (
   <NavContainer>
     {state.theme.menu.map(([name, link]) => {
-      // Check if the link matched the current page url
-      const isCurrentPage = state.router.link === link;
+      // Gets the page number
+      let pageNumber = state.router.link.replace(/\D/g, "");
+
+      // Check if the link matched the current page url, removing page + number from this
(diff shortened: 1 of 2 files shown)


rubenlorenzo:
Hi @luisherranz. I made this PR, but I don't know. If there is something, That I have to fix to correct by failures of the test.

packages/mars-theme/src/components/nav.js
```diff
@@ -9,8 +9,13 @@ import Link from "./link";
 const Nav = ({ state }) => (
   <NavContainer>
     {state.theme.menu.map(([name, link]) => {
-      // Check if the link matched the current page url
-      const isCurrentPage = state.router.link === link;
+      // Gets the page number
+      let pageNumber = state.router.link.replace(/\D/g, "");
```
luisherranz: I don't think this regular expression is safe to use. It will break if the slug contains a number. For example, `pageNumber` for `/my-category-12/page/3` would be `123` and not `3`.

You can get the page number from source data:

```js 
const pageNumber = data.page;
```

Be aware that `data.page` is always 1 for posts (not `null` or `undefined`).

luisherranz (changes requested):

packages/mars-theme/src/components/nav.js
luisherranz: Actually, forget about `data.page`, you can use `data.route` instead, which is the link without the page.

rubenlorenzo: Thank you, I will try to use it.

luisherranz:
Hey @rubenlorenzo, thanks for the changes. The code looks much nicer now 🙂

Have you tested it? Does it work?

rubenlorenzo:
Hi @luisherranz. Yes, I have tried it. It works with mars-theme-example in firefox and chrome.

luisherranz:
Awesome. Merged now. Thanks for your contribution @rubenlorenzo 🙂

PR #890, opened 2022-02-02
Dynamic public path

orballo:
**What**:

Implementing dynamic public path as described in its [Feature Discussion](https://community.frontity.org/t/change-publicpath-dynamically-at-runtime/1588/9).

**Why**:

To be able to set a public path using the `frontity.settings.js` file or change it dynamically with the `?frontity_public_path=` query option.

**How**:

Adding the public path value from env var, CLI arg, `frontity.settings` or the initial request query to `state.frontity.options.publicPath`, and using this value to generate the scripts in SSR and to populate `__webpack_public_path__` on the client.

More information in the [Feature Discussion](https://community.frontity.org/t/change-publicpath-dynamically-at-runtime/1588/9).

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests
- [x] Update community discussions
- [x] Add a changeset (with link to its [Feature Discussion](https://community.frontity.org/c/33) if it exists)

**Unrelated Tasks**

- [ ] Update starter themes
- [ ] Update other packages
- [ ] TypeScript tests

**Additional Comments**

PR #892, opened 2022-02-16
Disable newsletter from the CLI

SantosGuillamot:
The purpose of this PR is to disable the newsletter from the CLI. What I did:
* Stop asking the users to subscribe after running `npx frontity create`.
* Throw an error if anyone runs `npx frontity subscribe`.

I wasn't sure if throwing an error is the best idea or we should just log a message.

Moreover, I didn't know how to test it so I hope I'm not breaking anything.


luisherranz:
Thanks, Mario. 

I'll look into the test failures in https://github.com/frontity/frontity/pull/890 before revirewing/merging this.

luisherranz (approved):
Everything perfect. Thanks, Mario! 🙂

PR #909, opened 2022-03-09
Fix public path of static assets

### LINKED ISSUE #908: The public directory path of imported assets are not being loaded correctly
**_Bug report_**

- I've updated both my project and my packages to the latest Frontity versions
- I've searched for existing issues:
-- https://community.frontity.org/t/frontity-release-dynamic-public-path/5790/2
-- https://community.frontity.org/t/imported-image-src-not-being-prepended-with-public-directory-path/5783

**Expected behavior**

Import an static file like an image and load it directly on the code.

**Observed behavior**

 The problem is on the SSR. On the server side rendering the path of the images is the relative path of the current url, so for example:
- If we have the image **/resources/imgs/logo.png**
- on a page that have this slug **/about-page/**

The link to the image on SSR is

- **/about-page/images/logo.png** instead of being **/static/images/logo.png**

1. Create or update a frontity installation with the latest update
2. Try to import any static file, like an image
```
import Logo from "../images/logo.png";

const Header = ({ state }) => {
  return ( 
    <> 
       <img src={Logo} />
    </>
  );
};
```
3. The image will not load if you refresh the page or access via SSR

Link to the repository: 
https://github.com/husseincak/testfrontity/

luisherranz:
**What**:

Fix the public path of the assets (images, fonts, css...).

**Why**:

After the switch to dynamic public paths (#890), they public path of the static assets was not working correctly anymore.

**How**:

It seems like changing the dynamic public path (`__webpack_public_path__`) at request time is not possible, so I'm just doing a string replace in the SSRed HTML.

**Tasks**:

- [x] Code
- [x] End to end tests
- [x] Update community discussions ([this](https://community.frontity.org/t/frontity-release-dynamic-public-path/5790/2))

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


luisherranz:
I've fixed an unrelated problem with the Google Ad Manager E2E tests.

luisherranz:
<a href="https://www.loom.com/share/6d74106e42e54fdd81e1d66482de5dfb">
    
  </a>

AsadZahed:
@luisherranz i updated the frontity code same as you did it in pull request but some of the images have appeared, most of them are still missing. The path on missing images is written as attached in below picture:
errorimage
I updated these files:

- Public-path.ts
- server-side-rendering.ts
- index.ts

luisherranz:
Thanks, @AsadZahed. My mistake, it should be a global replace. I'll fix it right away 🙂

SantosGuillamot (approved):
LGTM 🙂

luisherranz:
Thanks Mario.

AsadZahed:
HI @luisherranz Thanks for the update, i just moved to new version but the font has still issues on reload of the page. I will be thankful if you look into it.

luisherranz:
Would you mind opening a new issue with information about the problem?

orballo (commented):
Looks good to me :)
EDIT: Too late xD

luisherranz:
Setting the public path to `__webpack_public_path__` in the server may not work in all cases: https://github.com/frontity/frontity/issues/913

Issue #913, opened 2022-03-25
Static assets that are used as a background in css emit a warning


husseincak:
**_Bug report_**

- I've updated both my project and my packages to the latest Frontity versions: https://docs.frontity.org/guides/keep-frontity-updated
- I've searched for existing issues.
- I've checked the documentation: https://docs.frontity.org
- I've checked the community forum: https://community.frontity.org

**Expected behavior**
Correctly styled rendering on the server side without any warnings when you add a static url from the server assets to the css props (like a background image)

**Observed behavior**
If you import an image and you add it as a background-image via css or style props, the console shows a warning, 

**Screenshots**
![image](https://user-images.githubusercontent.com/6590146/160214484-2b4bc0a6-7dde-4f9a-966c-cbcc76e86731.png)

**Steps involved to reproduce the problem**

1. Create a frontity repository: npx frontity create
2. Add any image the theme package
3. Import it on the component index
4. Add it as a background of any html element

I have created a repository with a hyper-simplified version of frontity where there is only a div with a static image as background: https://github.com/husseincak/testing/

**Info about the problem**
If you downgrade the package @frontity/core to the v 1.14.3, the warning no longer appears:
 - @frontity/core: ^1.15.1 => 1.14.3
If i console.log the asset i get two different urls on the client and on the server:
- Background URL on client: /static/images/image.jpg
- Background URL on the server: __webpack_public_path__images/image.jpg

Maybe the problem is because the two urls are diferent so the component is being rendered as a different component on the server and on the client

luisherranz:
Interesting. 

> - Background URL on client: /static/images/image.jpg
> - Background URL on the server: __webpack_public_path__images/image.jpg

It is related to this PR, where we started populating the dynamic public path with `__webpack_public_path__` in the server: https://github.com/frontity/frontity/pull/909

The problem is that setting the correct public path dynamically at request time seemed not possible. I guess we need to find a way to do it. If anyone has experience doing this or wants to give it a try, please do so 🙂

orballo:
Is this issue affecting the `publicPath` set in `frontity.settings`? or is it only related to the `frontity_public_path` query?

luisherranz:
We can try two things:

[...]

PR #916, opened 2022-04-06
Feature/server extensibility

orballo:
**What**:

Add Server Extensibility.

**Why**:

So Frontity packages can include server middleware.

**How**:

Following the [Final Implementation proposal](https://community.frontity.org/t/server-extensibility/2810/23).

**Tasks**:

- [x] Code
- [x] TSDocs
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests
- [x] TypeScript tests

**Unrelated Tasks**

- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**


orballo:
@luisherranz Could you have a quick look at this? As far as I can tell, the custom middleware is working with this implementation. I am still missing changing the signature of `beforeSSR` and `afterSSR` actions.

packages/core/src/server/middlewares/settings-and-store.ts
```diff
@@ -34,5 +34,11 @@ export const settingsAndStore = (packages) => async (
     url: ctx.URL,
   });
 
+  // Populate the context for custom middleware.
+  Object.assign(ctx, {
+    ctx: { ...ctx },
```
orballo: @luisherranz This is currently broken here, because if I spread ctx, the reference to `ctx.body` is lost, and it cannot be manipulated in both `fn(ctx)` and `fn({ctx})`.

orballo:
@luisherranz This is ready for review now :)

packages/core/src/server/index.ts
```diff
@@ -47,65 +51,93 @@ const server = ({ packages }: ServerOptions): ReturnType<Koa["callback"]> => {
...
 
-  // Serve static files.
-  app.use(async (ctx, next) => {
-    const publicPath = ctx.state.store.state.frontity.options.publicPath.replace(
-      /^(?:https?:)?\/\/([^/])+/,
-      ""
-    );
+  app.use((ctx, next) => {
+    // Collect custom middleware.
+    const middleware: Middleware[] = ((server: Package["server"]) => {
```
luisherranz: Why do you need to create a function here? It's not using the scope or being reused, is it?

packages/core/src/server/index.ts
```diff
@@ -47,65 +51,93 @@ const server = ({ packages }: ServerOptions): ReturnType<Koa["callback"]> => {
...
+      }, []);
+    })(ctx.server);
+
+    /**
+     * Middleware wrapper to add next to ctx.
+     *
+     * @param fn - Middleware.
+     * @returns - Middleware with next added to ctx.
+     */
+    const wrapper = (fn: Middleware) => (ctx: Context, next: Next) => {
```
luisherranz: This can be moved outside (it's not using any scoped variable).

packages/core/src/server/middlewares/scripts-stats.ts
```diff
@@ -20,8 +20,10 @@ export const scriptsStats = async (
     getStats({ target: "es5" }),
   ]);
 
+  if (!ctx.frontity) ctx.frontity = {};
```
luisherranz: We can move this declaration outside of this middleware because the next middlewares are also using `ctx.frontity` and they can crash if this is reordered in the future. We can create an `init` middleware, for example.

packages/core/src/server/middlewares/settings-and-store.ts
```diff
@@ -22,17 +22,23 @@ export const settingsAndStore = (packages) => async (
...
+    settings: ctx.frontity.settings,
     packages,
     url: ctx.URL,
   });
 
+  // Populate the context for custom middleware.
+  Object.assign(ctx, {
+    ctx: ctx,
+    ...ctx.frontity.store,
+  });
```
luisherranz: There's no need for destructuring here 🙂

```suggestion
  Object.assign(ctx, { ctx }, ctx.frontity.store);
```

packages/core/src/server/middlewares/capabilities-and-actions.tsx
```diff
@@ -17,25 +17,25 @@ export const capabilitiesAndActions = async (
...
+      if (init) return init(ctx);
     })
   );
 
   // Run beforeSSR actions.
   await Promise.all(
-    Object.values(ctx.state.store.actions).map(({ beforeSSR }) => {
-      if (beforeSSR) return beforeSSR({ ctx });
+    Object.values(ctx.frontity.store.actions).map(({ beforeSSR }) => {
+      if (beforeSSR) return beforeSSR(ctx);
```
luisherranz, line 38: Not that I think this is wrong, but why did you have to do this change?

e2e/packages/server-extensibility/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "e2e-server-extensibility",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Package to do e2e testing of Frontity's loadable utility",
```
luisherranz: Wrong description here.

e2e/packages/server-extensibility/types.ts
```diff
@@ -0,0 +1,23 @@
+import { Package } from "frontity/types";
+
+/**
+ * Package to do e2e testing of Frontity's Lodable component.
+ */
+interface ServerExtensibility extends Package {
```
luisherranz, line 6: This type is missing the `server` definition. A new `Server` type that correctly populates the `ctx` with the store is also missing.

luisherranz (changes requested):
As always, fantastic work @orballo 🙂

I've added a few minor comments, but the only thing I miss is a `Server` type that can be used in the package definition and populated `ctx` with the correct `state`, `actions`, `library`, etc.

Would you mind please adding that? Also, it would need some typescript tests as well.

Thanks!!

EDIT: Oh, this is missing a changeset as well.

packages/core/src/server/index.ts
orballo: Yeah, you are right. I usually use these functions when there are complex conditions to avoid using `let`, maybe I started with a different approach and then I changed my mind and still left the function here. I'll fix it.

packages/core/src/server/index.ts
orballo: Sure thing

packages/core/src/server/middlewares/scripts-stats.ts
orballo: I'll do that!

packages/core/src/server/middlewares/settings-and-store.ts
orballo: Hum, I didn't know it took more than one value

packages/core/src/server/middlewares/capabilities-and-actions.tsx
orballo: @luisherranz It seemed natural to do it this way, because if the `server` middleware is going to receive the `ctx` that contains all the parts of the store, the same `ctx` should be used for `beforeSSR` and `afterSSR`.

orballo: Btw, I also passed it to `init`.

luisherranz: Good. Let's keep it like that then 🙂

luisherranz: > Btw, I also passed it to init.

Yes, that's perfect. Thank you.

e2e/packages/server-extensibility/src/server.ts
```diff
@@ -0,0 +1,40 @@
...
+import type ServerExtensibility from "../types";
+
+const serverExtensibility: ServerExtensibility = {
+  name: "e2e-server-extensibility",
+  roots: {
+    serverExtensibility: Root,
+  },
+  server: {
+    serverExtensibility: {
+      robotsTxtOne: get("/robots-one.txt", ({ ctx }) => {
```
orballo: @luisherranz I run into this case when using Koa router middleware, the handler passed as a parameter doesn't have the types defined for `Server` or `AsyncServer`. I think we would need a different type for this case, but not sure how to proceed here.

I had a quick look at the `koa-route` types, but there isn't a type that allows passing a handler definition. Any ideas?

This might happen with other middleware, so I guess we need to create a generic one, instead of a solution for `koa-route`.

orballo: @luisherranz I fixed this casting the types on the `koa-route` middleware handlers.

there was something else you mentioned that needed my attention, but cannot remember. Otherwise this implementation should be ready.

luisherranz:
The e2e browserstack changes seem to be failing for an unrelated issue, but the unit tests are failing. Could you please take a look into those?

e2e/packages/server-extensibility/types.ts
```diff
@@ -18,6 +18,20 @@ interface ServerExtensibility extends Package {
...
+
+  /**
+   * Server middleware.
+   */
+  server?: {
+    /**
+     * ServerExtensibility namespace.
+     */
+    serverExtensibility: {
+      [key: string]:
```
luisherranz: Would you mind using the specific names (`robotsTxtTwo`, `accessFrontityState`...) of the functions here? This is going to be the best example of how to type these functions in our codebase.

e2e/packages/server-extensibility/src/server.ts
```diff
@@ -9,18 +10,18 @@ const serverExtensibility: ServerExtensibility = {
...
     serverExtensibility: {
-      robotsTxtOne: get("/robots-one.txt", ({ ctx }) => {
+      robotsTxtOne: get("/robots-one.txt", (({ ctx }) => {
         const robotsTxt =
           "User-agent: *\nAllow: /\nSitemap: http://www.example.com/sitemap.xml";
         ctx.type = "text/plain";
         ctx.body = robotsTxt;
-      }),
-      robotsTxtTwo: get("/robots-two.txt", (ctx) => {
+      }) as Server<ServerExtensibility>),
```
luisherranz, line 18: That's interesting. Ok, we'll leave it like this for now. Good solution 🙂

packages/types/__tests__/package.tests.ts
```diff
@@ -2,6 +2,9 @@ import Settings from "../settings";
 import Package from "../package";
 import { Action, AsyncAction } from "../action";
 import Derived from "../derived";
+import { Server, AsyncServer } from "../server";
+
+// import { Package, Action, Derived, Server } from "frontity/types";
```
luisherranz:
```suggestion
```

packages/types/index.ts
```diff
@@ -3,6 +3,7 @@ export { default as Package } from "./package";
 export { default as Fill } from "./fill";
 export { default as Frontity } from "./frontity";
 export { Action, AsyncAction, ServerAction, Context } from "./action";
+export { Server, AsyncServer, FrontityContext } from "./server";
```
luisherranz: What's the difference between `FrontityContext`, and action's `Context`? Can't we get rid of `Context` and rename `FrontityContext` to `Context` (and use it in `ServerActions` as well)?

luisherranz (changes requested):

packages/types/index.ts
orballo: Yes, I think it is safe to get rid of `Context`.

e2e/packages/server-extensibility/types.ts
orballo: Oki doki!

luisherranz (approved):
Everything looks perfect to me now, Eduardo. 

Many thanks for implementing such a useful feature! 🎉

---

Browserstack Firefox tests are failing, but it is unrelated to this PR, so I'll merge this without them.

Do you want to wait until you take a deeper look at https://github.com/frontity/frontity/issues/913 before doing a release?

orballo:
@luisherranz I'd rather release this feature before fixing the other one, as I have a bunch of tasks waiting for server extensibility to be available :)

luisherranz:
Ok, let's do a release tomorrow then 🙂

PR #925, opened 2022-06-13
Upgrade Webpack to Webpack 5

orballo:
**What**:

Upgrade Frontity to Webpack 5.

**Why**:

To stay up to date with the latest version and also to switch to the new `publicPath: auto` option.

**How**:

By replacing Webpack 4 with Webpack 5 and making sure everything keeps working.

**Tasks**:

- [x] Code
- [x] Update other packages
- [x] TypeScript
- [x] Unit tests
- [x] End to end tests

**Unrelated Tasks**

- [ ] Update starter themes
- [ ] TypeScript tests
- [ ] TSDocs

**Additional Comments**


packages/core/src/config/babel/index.ts
```diff
@@ -122,7 +122,11 @@ const getConfig = (target: Target): BabelConfig => {
     [
       "babel-plugin-transform-inline-environment-variables",
       {
-        include: ["CWD", "FRONTITY_INTERNAL_PUBLIC_PATH"],
+        include: [
+          "CWD",
+          "FRONTITY_INTERNAL_PUBLIC_PATH",
+          "FRONTITY_SOURCE_AUTH",
```
orballo: @luisherranz I have a question here. After updating to Webpack 5, the following line was throwing an error on the client because process doesn't exist, and it was breaking the JS.
https://github.com/frontity/frontity/blob/e67103a588ee6a6d64219e8a9aca26f7721e6bb2/packages/wp-source/src/actions.ts#L216

Is adding the env variable here the right solution?

luisherranz: If an env variable exists, it should read from it, yes. Docs: https://api.frontity.org/frontity-packages/features-packages/wp-source#state.source.auth

That used to work on the client and server, so maybe Webpack 4 was removing it on the client somehow.

Maybe try doing the `typeof window` trick? Something like:

```js
const authFromEnv = typeof process !== "undefined" && process.env.FRONTITY_SOURCE_AUTH;
```

luisherranz: Oh, wait, sorry. I just realized I didn't answer your question.

> Is adding the env variable here the right solution?

No, it is not. Those variables are replaced _at build time_ (when running `npm run build`), and the `FRONTITY_SOURCE_AUTH` variable should be read _at rendering time_ (in the server doing the SSR).

Try the `typeof` trick instead.

orballo: Oki doki! Thanks! I was trying to only do changes on core, but I'll do that then on wp-source.

luisherranz: For wp-source it should be backward compatible so no problem.

orballo:
@luisherranz I'm running into a hydration warning like the following:

Screenshot 2022-06-17 at 11 22 57

I've been investigating for a few hours and I'd say it is an issue with `loadable` because it happens only when a dynamic imported component is being rendered on server side and then hydrated.

But at the same time I noticed something strange, and it is that the `<style>` tags for `emotion` are rendered before each element in SSR, but when I inspect them in the browser, they are moved to the `<head>`. It doesn't look like that's the cause of the hydration warning though, but I thought of mentioning it just in case. I tried to change the way styles are extracted and added to the body, but I get empty strings from the extractors, and when the component is not a loadable component it doesn't cause any issues. So I'm mentioning it just in case, but it doesn't look like the root of the issue.

I was wondering if you ever run into something similar when setting up webpack?

EDIT: This is happening when running the `mars-theme` project in the repo.

orballo:
@luisherranz I run into this issue https://github.com/gregberge/loadable-components/issues/587

It says that `@loadable/babel-plugin: 5.13.1` fixes it, but apparently it was never released. Instead there is version `5.13.2` which still shows the hydration error.

I'm downgrading `@loadable/babel-plugin` to `5.12.0` for now.

orballo:
@luisherranz I run into a new error when trying to run the e2e tests. Apparently the build process throws an error with the AMP package.

Screenshot 2022-06-21 at 08 27 30

Very descriptive as you can see. While I investigate more, I was wondering if you run into something similar before with AMP?

luisherranz:
> I'm downgrading @loadable/babel-plugin to 5.12.0 for now.

Sounds good. Does it solve all the hydration issues?

> I was wondering if you run into something similar before with AMP?

Not that I'm aware of. Did you manage to extract more information?

orballo:
@luisherranz The error building sites with the `amp` package turned out to be Webpack 5 not polyfilling the `stream` package anymore, so I added the polyfill manually to core and added the fallback rule in webpack.

orballo:
@luisherranz Some unit tests are failing because of `require.resolve()` returning a different path depending on the environment. Not sure how to approach that.

Also, please have a look at this part of Webpack 5 docs: https://webpack.js.org/configuration/resolve/#resolvefallback

Should we add all the polyfills that were used in Webpack 4 just to be safe? Apparently the issue with `process` in the browser not existing also comes from a missing polyfill. I'm a bit afraid that users might run into more missing polyfills that we are not catching.

What do you think?

EDIT: Quick note, I tried to mock `require.resolve()` with Jest but it seems that it is not possible to mock.
EDIT 2: I see some e2e tests failing that are passing in my local env.

luisherranz:
> Some unit tests are failing because of require.resolve() returning a different path depending on the environment

I'm not familiar with `require.resolve()`. Could you please explain what has changed from Webpack 4 to Webpack 5? Why are paths different? How are the paths in W4 and how are they in W5?

> Should we add all the polyfills that were used in Webpack 4 just to be safe?

That'd be nice to maximize compatibility, although I think this still requires a major bump. Is there a list of missing polyfills?

orballo:
> I'm not familiar with `require.resolve()`. Could you please explain what has changed from Webpack 4 to Webpack 5? Why are paths different? How are the paths in W4 and how are they in W5?

I didn't mean that there is a difference between using it in W4 and W5. `require.resolve()` returns an absolute path of the module in the system, and therefore returns different values when you run tests in local, than when you run then on Github. Here it is:

```
@frontity/core:     -         "stream": "/Users/orballo/Code/frontity/node_modules/stream-browserify/index.js",
@frontity/core:     +         "stream": "/home/runner/work/frontity/frontity/node_modules/stream-browserify/index.js",
```
The only workaround I can think of regarding this test, is to remove that part of the object returned from webpack config before matching the snapshot.



> I think this still requires a major bump. Is there a list of missing polyfills?

Here is the list of polyfills in Webpack 4 that are now missing unless we add them manually in Webpack 5:

```js
module.exports = {
  //...
  resolve: {
    fallback: {
      assert: require.resolve('assert'),
      buffer: require.resolve('buffer'),
      console: require.resolve('console-browserify'),
      constants: require.resolve('constants-browserify'),
      crypto: require.resolve('crypto-browserify'),
      domain: require.resolve('domain-browser'),
      events: require.resolve('events'),
      http: require.resolve('stream-http'),
      https: require.resolve('https-browserify'),
      os: require.resolve('os-browserify/browser'),
      path: require.resolve('path-browserify'),
      punycode: require.resolve('punycode'),
      process: require.resolve('process/browser'),
      querystring: require.resolve('querystring-es3'),
      stream: require.resolve('stream-browserify'),
      string_decoder: require.resolve('string_decoder'),
      sys: require.resolve('util'),
      timers: require.resolve('timers-browserify'),
      tty: require.resolve('tty-browserify'),
      url: require.resolve('url'),
      util: require.resolve('util'),
      vm: require.resolve('vm-browserify'),
      zlib: require.resolve('browserify-zlib'),
    },
  },
};
```

At first I thought that a minor bump would be enough, but it is true that people that extended Webpack in `frontity.config.js` might run into breaking changes, so I guess we need to bump the core to `v2.0`.

luisherranz:
Ohhh, ok, got it, thanks 🙂

> The only workaround I can think of regarding this test, is to remove that part of the object returned from webpack config before matching the snapshot

Sounds good to me. Something like `.replace(/^.*\/node_modules/, "")` right?

> Here is the list of polyfills in Webpack 4 that are now missing unless we add them manually in Webpack 5

Sounds goods. Let's add those 🙂

> it is true that people that extended Webpack in frontity.config.js might run into breaking changes

Yes, Webpack 4 -> 5 is in itself a breaking change, so we should do the same.

orballo:
@luisherranz I added the polyfills to Webpack and also changed the node versions in the Github actions to remove those no longer maintained.

I can see locally that `custom-render`, `dynamic-public-path` and `loadable` e2e tests fail in production. I'll investigate a bit more and hopefully do a final push that works.

luisherranz:
> also changed the node versions

We should take this opportunity (Frontity v2) to update the minimum required version to Node 16.

orballo:
@luisherranz I think this is now ready for CR :)

orballo:
@luisherranz I think this is ready for code review.  It is finally passing _almost_ all tests. It fails on browserstack Firefox tests, not sure if that was happening before, but I don't know how to test that locally.

.changeset/beige-days-breathe.md
```diff
@@ -0,0 +1,7 @@
+---
+"@frontity/core": major
```
luisherranz, line 2: We need a major bump of `frontity` as well.

```suggestion
"@frontity/core": major
"frontity": major
```

.github/workflows/release.yml
```diff
@@ -18,10 +18,10 @@ jobs:
           # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
           fetch-depth: 0
 
-      - name: Setup Node.js 12.x
+      - name: Setup Node.js 14.x
```
luisherranz:
```suggestion
      - name: Setup Node.js 16.x
```

luisherranz (approved):
It looks good to me @orballo!! 🙂

We should release this as a `beta` first. I'll create a new branch and I'll change the base.

luisherranz:
>It fails on browserstack Firefox tests, not sure if that was happening before

Yes, it was. Let's ignore those for now.

luisherranz:
Released as v2 beta:

```
frontity@2.0.0-beta.0
@frontity/core@2.0.0-beta.0
```

The rest of the packages have also been updated because their dependencies changed:

```
@frontity/amp@0.1.2-beta.0
@frontity/analytics@1.4.2-beta.0
@frontity/components@1.8.3-beta.0
@frontity/comscore-analytics@1.3.1-beta.0
@frontity/connect@1.3.1-beta.0
@frontity/google-ad-manager@1.0.1-beta.0
@frontity/google-analytics@2.1.1-beta.0
@frontity/google-tag-manager-analytics@1.3.1-beta.0
@frontity/head-tags@1.1.2-beta.0
@frontity/hooks@2.2.3-beta.0
@frontity/html2react@1.7.1-beta.0
@frontity/router@1.3.1-beta.0
@frontity/smart-adserver@1.1.2-beta.0
@frontity/source@1.5.3-beta.0
@frontity/tiny-router@1.4.5-beta.0
@frontity/types@1.10.1-beta.0
@frontity/wp-comments@1.0.1-beta.0
@frontity/wp-source@1.11.8-beta.0
@frontity/yoast@2.1.2-beta.0
```

PR #931, opened 2022-07-18
Fix #930 (Author handler) get all cpt entities

### LINKED ISSUE #930: Populate CPT entities when fetching author page ("/author/:slug")
***CPT***
***Bug Report***

#### Expected behavior
Retrieve all cpt items of an author when fetching author's page, eg. ("/author/admin").
Then find all these items populated in state (frontity.state.data["/author/admin"].items) as {id,link,type}[]

#### Observed behavior
I'm only getting entities of type (post).
I've already defined my cpts in frontity.settings.ts file and can access them correctly by built-in handlers. 
While I have other cpt entites releated to the same author.  

image

https://github.com/frontity/frontity/blob/0e7ed6e6774d49ba2358d3b207299834bc8d935e/packages/wp-source/src/libraries/handlers/author.ts#L58-L77

**Thanks in advance for checking this**

hussam-aldarwish:
**What**:
Trying to fix #930  by fetch (api.get) all cpt endpoints.

**Why**:
To retrive all author cpt entities.

**How**:
Loop over (state.sources.postTypes) array and make api call to each cpt endpoint.

**Tasks**:

- [x] Code
- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [x] Update other packages
- [ ] TypeScript tests
- [ ] Update starter themes

**Unrelated Tasks**

**Additional Comments**

### DIFF
--- a/.changeset/purple-otters-clean.md
+++ b/.changeset/purple-otters-clean.md
@@ -1,0 +1,5 @@
+---
+"@frontity/wp-source": patch
+---
+
+Added get all cpt entities of author when fetching author page ("/author/:slug")
--- a/packages/wp-source/src/libraries/handlers/author.ts
+++ b/packages/wp-source/src/libraries/handlers/author.ts
@@ -68,13 +68,46 @@ const authorHandler: Handler = async ({
   });
 
   // 3. Populate response.
-  const items = await populate({ response, state, force });
-  if (page > 1 && items.length === 0)
-    throw new ServerError(`author "${slug}" doesn't have page ${page}`, 404);
+  let items = await populate({ response, state, force });
 
   // 4. Get posts and pages count.
-  const total = getTotal(response, items.length);
-  const totalPages = getTotalPages(response, 0);
+  let total = getTotal(response, items.length);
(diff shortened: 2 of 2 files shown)


luisherranz:
Thanks for the PR, Hussam 🙂

However, WordPress author archives don't display custom post types by default, so I'd rather keep Frontity as close as possible to WordPress.

But why don't you make this a Frontity package instead? The code is great. You just need to add it as [a new handler with a higher priority](https://api.frontity.org/frontity-packages/features-packages/wp-source#libraries.source.handlers) than the default one. Then, any person that wants this can install your package.

hussam-aldarwish:
Thanks for replying, Luis

I thought it might be better to include custom post types in author archives, but later I've discovered that WordPress author archives don't display custom post types by default, as you stated.

Thanks for reviewing my code.
And too many thanks for Frontity

luisherranz:
Awesome. Thank you for your understanding Hussam 🙂

I'm going to close this now but feel free to reopen it if the situation changes.

PR #935, opened 2022-10-27
Fix issue where feed is matched anywhere in the permalink

tobeycodes:
**What**:

Update the regex for the Yoast plugin to determine when to transform links. Currently we have canonical urls pointing to the WP site that contain the word feed.

**Why**:

The regex pattern only matches wp-* links at the start of the string where as feed, comments and xmlrc anywhere in the string. 

See the regex example of how it currently works https://regex101.com/r/zcxmVX/1

**How**:

I updated the regex to match everything from the start of the string so that `feed-asd` will match but `asd-feed` won't

**Tasks**:

- [x] Code

**Unrelated Tasks**

- [ ] TSDocs
- [ ] TypeScript
- [ ] Unit tests
- [ ] End to end tests
- [ ] TypeScript tests
- [ ] Update starter themes
- [ ] Update other packages

**Additional Comments**

### DIFF
--- a/.changeset/violet-buckets-unite.md
+++ b/.changeset/violet-buckets-unite.md
@@ -1,0 +1,5 @@
+---
+"@frontity/yoast": patch
+---
+
+Only match feed, comments or xmlrpc if they are at the beginning of the string.
--- a/packages/yoast/src/index.ts
+++ b/packages/yoast/src/index.ts
@@ -8,7 +8,7 @@ const yoastPackage: YoastPackage = {
     yoast: {
       renderTags: "both",
       transformLinks: {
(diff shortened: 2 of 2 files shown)


luisherranz:
@tobeycodes, could you please add a full description of the PR? It is difficult to infer what the problem is and how you solved it by just looking at the PR title and the code change.

tobeycodes:
@luisherranz Sorry about that. I updated it. To be clear we are able to update this setting ourselves when importing packages so it is not a priority to merge but I wanted to create the pull request so if anyone else comes across the same issue they see the fix.

luisherranz:
Ok, got it now. Thanks @tobeycodes 🙂

tobeycodes:
Closing this since it has been open for a while without movement. Thanks.

